PR #8, opened 2014-05-04
Trie Update

Arkshine:
General update, based what on [bug 5892](https://bugs.alliedmods.net/show_bug.cgi?id=5892) (SourceMod) by David.
- Removed KTrie dependency to Hash map
- Added new natives, including iterators:
  - `TrieGetSize`
  - `TrieSnapshotCreate`
  - `TrieSnapshotLength`
  - `TrieSnapshotKeyBufferSize`
  - `TrieSnapshotGetKey`
  - `TrieSnapshotDestroy`
- Added new `replace`parameters for `TrieSet*` natives to make operation fail if key already set.
- Added new `size` parameter for `TrieGetString` and **TrieGetArray``` to return the number of bytes written in the buffer.
- Updated documentation.
- Updated test plugin.

### DIFF
--- a/plugins/include/celltrie.inc
+++ b/plugins/include/celltrie.inc
@@ -25,140 +25,143 @@ enum Snapshot
  * been internally replaced with hash tables, which have O(1) insertion time
  * instead of O(n).
  *
- * @return 			New Map Handle, which must be freed via TrieDestroy().
+ * @return 			New Map handle, which must be freed via TrieDestroy().
  */
 native Trie:TrieCreate();
 
 /**
  * Clears all entries from a Map.
  *
- * @param handle	Map Handle.
+ * @param handle	Map handle.
  *
- * @error			Invalid Handle.
+ * @error			Invalid handle.
  */
 native TrieClear(Trie:handle);
 
 /**
  * Sets a value in a hash map, either inserting a new entry or replacing an old one.
  *
- * @param handle	Map Handle.
+ * @param handle	Map handle.
  * @param key		Key string.
  * @param value		Value to store at this key.
  * @param replace	If false, operation will fail if the key is already set.
  *
  * @return			True on success, false on failure.
- * @error			Invalid Handle.
+ * @error			Invalid handle.
  */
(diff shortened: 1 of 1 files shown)


Adidasman1:
not working properly TrieGetString
When you save a number in TrieSetString

Adidasman1:
new Trie:check;
new g_sys[33];
new g_sysnew[33];

public plugin_init()
{
register_plugin("Check GetString","1.0","-");
register_clcmd("SetString", "SetString");
register_clcmd("GetString", "GetString");
check = TrieCreate();
return 0;
}

public SetString(id)
{
g_sys[id] = get_systime(0);
g_sysnew[id] = get_systime(0);
TrieSetString(check, "test", g_sys[id]);
return 1;
}

public GetString(id)
{
if(TrieKeyExists(check, "test"))
{
TrieGetString(check, "test", g_sysnew[id], 11);
server_print("GetString - Old[%d] New[%d]", g_sys[id], g_sysnew[id]);
}
return 1;
}

GetString - Old[1445157091] New[227]

Arkshine:
What you're doing doesn't make sense. If you're using `TrieSetString` this means you associate a key to a string. Here you are associate to a simple number. For sure it's not going to work properly. If you work with number, use `TrieSetCell` and `TrieGetCell`.

Adidasman1:
previously worked in the same way as I wrote above, but stopped after the upgrade.

Nextra:
This is plain API misuse and not a sign of AMXX or this PR being broken. The behavior of `TrieSetString` in that scenario is dependent on the array elements after `g_sys[id]` (as the array is interpreted as a string) anyway, so you'll not get any consistency out of that code. I'd guess that the difference in behavior relates to AMXX UTF-8 support as opposed to this PR, but it doesn't really matter either way. If you want to store a number in a trie you use `TrieSetCell`.

Adidasman1:
I did so on the TrieSetCell.
but I was wondering why previously worked TrieGetCell.

PR #88, opened 2014-07-29
Remove some client_cmd

xPaw:


plugins/admin.sma
```diff
@@ -803,12 +803,12 @@ accessUser(id, name[] = "")
...
 	if (result & 4)
 	{
-		client_cmd(id, "echo ^"* %L^"", id, "PAS_ACC")
+		console_print(id, "* %L, id, "PAS_ACC")
 	}
 	
 	if (result & 8)
 	{
-		client_cmd(id, "echo ^"* %L^"", id, "PRIV_SET")
+		console_print(id, "* %L, id, "PRIV_SET")
```
WPMGPRoSToTeMa: You should use engclient_print instead of console_print, because console_print doesn't work for connecting players.

plugins/admin.sma
```diff
@@ -803,12 +803,12 @@ accessUser(id, name[] = "")
 	
 	if (result & 4)
 	{
-		client_cmd(id, "echo ^"* %L^"", id, "PAS_ACC")
+		console_print(id, "* %L, id, "PAS_ACC")
```
WPMGPRoSToTeMa: You should use engclient_print instead of console_print, because console_print doesn't work for connecting players.

plugins/admin.sma
```diff
@@ -792,7 +792,7 @@ accessUser(id, name[] = "")
 	
 	if (result & 1)
 	{
-		client_cmd(id, "echo ^"* %L^"", id, "INV_PAS")
+		console_print(id, "* %L, id, "INV_PAS")
```
WPMGPRoSToTeMa: You should use engclient_print instead of console_print, because console_print doesn't work for connecting players.

PR #114, opened 2014-08-13
Add filter_chat_input stock

xPaw:


xPaw:
So what do you think, @Arkshine?

Arkshine:
I still don't know what to do. Current stock is pointless as it's fixed in client beta.
The exploit is also not only about `%s`, but with localized string which can be used in `TextMsg`, `SayText` and `ShowMenu`. To properly fix the exploit, you need a global fix, ideally fixing on the client. But  doesn't seem motivated to do something or maybe no one has properly reported the issue?

xPaw:
Which client beta? As far as I'm aware there is currently no beta going for CS1.6?

Arkshine:
There is. Alfted has removed ability to use %s with TextMsg. So exploit with %s%s%s etc, doesn't work with beta client.

WPMGPRoSToTeMa:
@Arkshine client need only crash fixes. Why client_print sends message in TextMsg format argument?

Arkshine:
Is not `%s` filtered in client beta? With or without it doesn't change fact buffer issue need to be fixed on client.

ClaudiuHKS:
Well, it's kinda' hard-coded.
In my opinion, you should only use strlen() inside the function.
Otherwise, you might have to declare it so many times in a large script.

It would be better to improve the code, making sure that only normal, readable characters are passing the filter.

http://www.cplusplus.com/reference/cctype/

This might be merged, if meeting the criteria and those with write access will want to.

PR #148, opened 2014-10-24
Update sqllite

Arkshine:
Updated SQLite to 3.8.7 (amalgamation version). Current version is like 8 year old, so I guess an update is more than welcome.
This should be safe, no breaking changes have to be noted, at least which concerns our use (we don't use custom build, nor obscure APIs).

@Nextra, please.


xPaw:
Looks like sqlite likes to update quite a lot, current version is already 3.8.8.2 which includes bug fixes and performance improvements.

If this pull request doesn't break any backwards compatability, then I see no problem in merging it.

PR #162, opened 2014-12-06
Add C++11 support

Arkshine:
I wanted to sync AMTL at first and seen we're using deprecated c++98.
Later I've found Seta00 has created #120 to build in C++11 mode but has few missing changes.
So there we go.

Changes in `AMBuildScript` are stolen from SourceMod.


dvander:
nice! :ship:

PR #165, opened 2014-12-10
Update versioning system and add missing module RC file (bug 6222)

Arkshine:
Versioning mechanism stolen from SourceMod, so you can compile with VS/MakeFile without having to rely on a missing header file which was generated only with AMBuild.

All those `version.rc` added for each module are actually the same file since it relies on defines from  `moduleconfig.h`. I'm not sure if here is a smart way to somehow generate such file as it's kind of redundant -  not a big deal though.


dvander:
nice. lgtm

PR #169, opened 2015-01-04
Added note for SQL_SetCharset native

voed:
I just used SQL_SetCharset in my plugin and its doesnt work on AMXX < 1.8.3. Please, add this note to native.


asherkin:
The include files are already versioned, and only contain natives currently available.
If you're using the include files from 1.8.3, your plugin is targeting 1.8.3.

voed:
Sure, but https://www.amxmodx.org/api/ generated by this files as i know, and i have no any notes there.

PR #170, opened 2015-01-05
meta_api.cpp: modify csstats_rank value to match amxx.cfg default value

Freeman-AM:
modify csstats_rank value to match amxx.cfg default value aka save by steamID and not by nick.

"if someone forgot to overwrite all their files when they installed the CStrike Addon, it will default to saving by Nick"


Arkshine:
:ship:

PR #173, opened 2015-01-06
Add string literal concatenation and stringizing operator

Arkshine:
This the same kind of patch as SM: https://github.com/alliedmodders/sourcemod/pull/5
I'm not going to repeat so feel free to see link for more details ^.

This basically allows you to concatenate literal strings and stringizing  a parameter in macro substitution *only* (so possible breakage is very limited ).

As per your suggestion, this uses  `+` instead of `...` to concatenate.
Random examples:
```PAWN
#define PROJECT_AUTHOR    "Greyscale"
#define PROJECT_COPYRIGHT "Copyright (C) 2010  " + PROJECT_AUTHOR

#define VERSION_MAJOR      "1"
#define VERSION_MINOR      "8"
#define VERSION_RELEASE    "3"
#define VERSION            VERSION_MAJOR + "." + VERSION_MINOR + "." + VERSION_RELEASE

#define log(%1)  "logging: " + #%1 + "\n"
```

Also I did not bother to remove checks related to packed string because in AMXX you can still use them in a limited way though, but I know people/plugin using them, this would be fine to let this as it is.

Please don't die again with this ~~horrible~~ code!

### DIFF
--- a/compiler/libpc300/sc2.c
+++ b/compiler/libpc300/sc2.c
@@ -36,8 +36,9 @@
 #endif
 
 /* flags for litchar() */
-#define RAWMODE         1
-#define UTF8MODE        2
+#define RAWMODE         0x1
+#define UTF8MODE        0x2
+#define ISPACKED		0x4
 static cell litchar(const unsigned char **lptr,int flags);
 
 static void substallpatterns(unsigned char *line,int buffersize);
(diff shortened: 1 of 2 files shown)


dvander:
Sure. :ship:

PR #175, opened 2015-01-07
plmenu: Fix another alive spectator bug

Freeman-AM:
Switch a newly connected user to spectator who selected his team but not his class can create an alive spectator if done during the first seconds of the round.

Solution : block the team transfer for any people stuck in this menu.
Well it's the worst case and probably the only possible one to completely fix this issue.

### DIFF
--- a/plugins/plmenu.sma
+++ b/plugins/plmenu.sma
@@ -712,9 +712,9 @@ public actionTeamMenu(id, key)
 				return PLUGIN_HANDLED
 			}
 
-			if ( g_fakemeta )
+			if( g_fakemeta )
 			{
-				if ( get_pdata_int(player, m_iMenu) == Menu_ChooseAppearance )
+				if( get_pdata_int(player, m_iMenu) == Menu_ChooseAppearance )
 				{
 					new name[MAX_NAME_LENGTH]
 					get_user_name(player, name, charsmax(name))
@@ -739,6 +739,11 @@ public actionTeamMenu(id, key)
 			log_amx("Cmd: ^"%s<%d><%s><>^" transfer ^"%s<%d><%s><>^" (team ^"%s^")", name, get_user_userid(id), authid, name2, get_user_userid(player), authid2, g_CSTeamNames[destTeamSlot])
 
 			show_activity_key("ADMIN_TRANSF_1", "ADMIN_TRANSF_2", name, name2, g_CSTeamNames[destTeamSlot]);
+			
+			if( !g_fakemeta && destTeamSlot == 2 )
+			{
+				engclient_cmd(player, "joinclass", "6")
+			}
 
 			if ( g_CSPlayerCanSwitchFromSpec[player] && g_cstrike && (CS_TEAM_T <= cs_get_user_team(player) <= CS_TEAM_CT))
 			{
--- a/plugins/plmenu.sma
+++ b/plugins/plmenu.sma
@@ -712,6 +712,18 @@ public actionTeamMenu(id, key)
 				return PLUGIN_HANDLED
 			}
 
+			if ( g_fakemeta )
+			{
+				if ( get_pdata_int(player, m_iMenu) == Menu_ChooseAppearance )
+				{
+					new name[MAX_NAME_LENGTH]
+					get_user_name(player, name, charsmax(name))
+					client_print(id, print_chat, "%L", id, "CANT_PERF_PLAYER", name)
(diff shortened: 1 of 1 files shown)


Nextra:
Executing a "joinclass" engclient_cmd might be enough to properly resolve this situation, allowing a team switch without bugs.

Freeman-AM:
Well, it was the case before. But i forgot something.

Nextra:
So what happens exactly? Which mods does this problem occur on? For CS the joinclass method really should be enough, but it may have to move to a different spot to work properly.

Freeman-AM:
at round start, normal players who join the party can spawn after the start of the round.

This method doesn't work at all in fact, because if it's at round start, the spectator can still spawn that way.

Arkshine:
If I understand, with original code: a player chooses team, but not class. Round starts. An admin switches player to spectator. Player respawns because of joinclass 6 and that player is allowed to spawn at this point, then plugin tries to kill player, and does jointeam 6 to transfer to spectator. At this point you're saying that plugin failed to kill the spawned player, but jointeam worked, which result a spawned spectator, right?

Nextra:
Player joins the server and selects a team (but not a model). Executing the *joinclass* command will set the model correctly for the team he is on. Crucially it also sets `m_iJoiningState = GETINTOGAME` which will then make the player spawn on the next server frame as `CBasePlayer::JoiningThink` [does not check the team the player is on](https://github.com/Arkshine/CSSDK/blob/266a9089818b214caf0d8c87d5a6577d1d7d002f/dlls/player.cpp#L1801). As this happens on the next frame any `is_user_alive()` safeguards immediately within this function will not work properly and we have an alive spectator shortly after.

Here are some ideas for fixing the issue:

1) Don't execute the *joinclass* command when moving players to spectator. I don't know if this has any side effects so this would require further testing. Probably have to cancel the open choose appearance menu.
2) Set `m_iJoiningState = JOINED` as we already work with Fakemeta now, thus preventing the player spawn on the next frame.
3) Work with the *spectate* command. Although that really only does *jointeam 6* anyway, which we already have in this function for the non-cs codepath. I don't know why this was split up originally, but for switching people to spectator using the alternate codepath might actually be the better solution (may also allow us to remove the awkward Fakemeta code from this plugin).

Arkshine:
Solution with `m_iJoiningState` is probably the cleanest as when you choose spectator from team menu, this offset is set to `JOINED`. Setting with `m_bTeamChanged` is probably a good place for that. 

Disallowing could have been an okay solution if it was available for this short period of seconds where a player can respawn anytime. But all the time doesn't seem like a good idea as this can be forced without issues.

Removing code to enforce `joinclass 6` is probably okay. So ti fix patch, I would remove the block of code above, and adding the new offset. This should be okay this way.

Freeman-AM:
http://pastebin.com/j5vNXCDH
Don't work as expected.
We can't control the camera (act like an unassigned team camera) and there is a skin bug as you can see.
I had the same problem when i tried to block spawn.
test
de_dust20002
de_dust20003

Arkshine:
I did not say to remove `joinclass 6`, it's still  needed to unstuck player. The offset is there to avoid player be respawned the next frame.

Freeman-AM:
http://pastebin.com/nRc31SqG
I obviously know, but even with joinclass, as in this code, i have the same problem, that why i removed it, for testing.

Arkshine:
It works for me. I don't know if I test properly, but I do the following:

* I choose team but not appearance
* I'm waiting for round start
* I send a command to execute: joinclass 6 + m_iJoiningState to 0 + jointeam 6.

And I don't respawn and well in spectator team.

Freeman-AM:
Okay, so  the problem is the plugin and a wrong if statement calling a block rather the correct one.

Arkshine:
Maybe you don't have fakemeta enabled ? :P

Freeman-AM:
I have fakemeta enabled, it's just that at the plugin current state, no jointeam 6 is called after or before  m_iJoiningState  JOINED and joinclass 6.

Arkshine:
I would say that because of `g_CSPlayerCanSwitchFromSpec[player]` being basically always true whatever you select a valid team. It's probably okay to pass in this check, but at this point, it needs to add a check that if we select spectator then we set `m_iJoiningState` to 0.

Code is quite messy :/ 
This would need really to be rewritten properly.

Freeman-AM:
I close this one for the moment.

PR #177, opened 2015-01-11
Compiler: Fix tautological warning with clang-3.5

Arkshine:


dvander:
It could be the indent was `sym && sym->name[0]` but who knows. LGTM!

Arkshine:
I'm don't know. Whatever SM or latest Pawn version, it uses either with or without with `!= NULL`.
This is probably fine as it is, I was just happened to compile against clang-3.5.

PR #178, opened 2015-01-11
Geoip: Update libmaxminddb to 1.0.4

Arkshine:
While I was around Geoip, here an update of Geoip2 library: https://github.com/maxmind/libmaxminddb/releases.
Mainly windows improvements. 

The only things I have adjusted are:
* Moved Geoip country codes arrays in `geop_utils.cpp` because they are no more included in Geoip2 and this file is better place for that.
* Moved C99 header files in its own directory to avoid any collision issue and to make more obvious why these files are there.
* Rearranged a bit declaration of stdbool.h and inttypes.h to target specifically VS++ version < 2013.


Arkshine:
Going to update this later, as AMXX compiles now with msvc12 and C99 is now supported, which means I don't need to use tricks anymore and library will be more close of the original one. It will be less a pain to update next time.

dvander:
Looks fine btw. For what its worth SM requires msvc12 + service pack 3 since C99 support has bugs in the older versions.

Arkshine:
#SuperDuck confirmed me Update 4 is installed.

PR #180, opened 2015-01-14
Fix missing DLLFunc_ClientCommand handling

Arkshine:
Even though it's documented,  `DLLFunc_ClientCommand` is missing from switch handled in `dllfunc()` native.


xPaw:
:shipit:

Arkshine:
We discussed about it with Nextra, actually this would makes sense to add it if we had a way to alter passed arguments which is retrieved by pfnCmd_Argv. Because as it is, `DLLFunc_ClientCommand ` will simply call `ClientCommand` from the game, with current set arguments, that's not really useful.

Nextra:
Yeah, right now we have no way of effectively using this, if we can't make it work and it has never worked in the first place we might aswell remove it (or at least clearly mark it as broken/deprecated/nuked).

PR #182, opened 2015-01-18
Fix a mysql crash after CString -> ke::AString conversion

Arkshine:
Related to #112.

`FieldNumToName()` and `GetString()` can return `NULL`.
`AString(const char *str)` doesn't check if pointer is valid, resulting a crash in `strlen`.


Nextra:
:shipit:

PR #184, opened 2015-01-23
Rename change_level to engine_changelevel

Arkshine:
Because name was kind of too generic and some existing plugins would seem to use that name for private functions, making compiler whining.

While we can, and since it has been implemented recently, let's rename it with less generic name.


Nextra:
:shipit:

PR #186, opened 2015-01-29
Wrong product version for amxmodx-installer.nsi

Freeman-AM:


Arkshine:
Thanks. Totally forgotten about this installer. Well ideally, that's something which should be generated, but well, for now, let's do that.

PR #187, opened 2015-01-29
Fix potential formatting issue with nvault_[p]set

Arkshine:
Reported by Bugsy.

```PAWN
native nvault_set(vault, const key[], const value[]);
native nvault_pset(vault, const key[], const value[]);
```
`value` parameter can't be formatted.

If passed data was coincidentally containing adjacent bytes which would be similar to "%s", "%d", etc., native would think data needs to be formatted and will result a *String formatted incorrectly* error.

It would seem that originally `value` was formattable but it has been changed right after and code has not been updated to reflect the change.


Nextra:
:ship:

PR #188, opened 2015-01-29
Establish a default timeout for MySQL connectivity

Arkshine:
Same as https://github.com/alliedmodders/sourcemod/pull/248.

* This adds new read and write timeout.
* This unifies all timeout to use the same value.
* This establishes a default timeout of 60 seconds and considering we don't have proper configuration system like in SM, for existing plugin it may be painful to change timeout, that's why default value is configurable in `core.ini`.


Nextra:
I hope it is okay to put this kind of config into core.ini, it is the best place for it imo. :ship:

xPaw:
Why duplicate `amx_sql_timeout` cvar? Can't amxx plugins read from core.inc as well?

Nextra:
@xPaw: This is not a duplication of the `amx_sql_timeout` variable. This is the default timeout the core MySQL module uses if no timeout is specified (i.e. `amx_sql_timeout` or any other variable passed into `SQL_MakeDbTuple` is 0). You can still have your own timeouts when creating a database connection, but the core will mandate a timeout of 60 seconds if no timeout is specified. This is similar to how SourceMod now does things, and I think it's reasonable. Since AMXX has no unified configuration method for databases and many legacy plugins do not specify a timeout in `SQL_MakeDbTuple` we figured a way to change this mandated timeout is necessary, hence the new entry in core.cfg

@AngeIII: What do you mean? We added the new entry in core.cfg precisely to allow changing the constant.

By the way: It is intended that an undefined timeout and we now have a read/write timeout established.

AngeIII:
@Nextra I mean maybe better to write constant "60" into some #define (if once it will be needed to change it will be very easy to change in one place only).

Arkshine:
@AngeIII I think you don't quite understand. Timeout is per database connection. That's something you provide in `SQL_MakeDbTuple`. _If no timeout is provided_, MySQL module will set a default timeout of 60 seconds. For convenience, this default value is user-configurable in `core.ini`.

Nextra:
@AngeIII: This is an AMXX module config option which is why it is located in core.ini. I don't quite understand how a compile-time define would be in any way easier than a runtime config option. Nobody wants to recompile a module to change such a simple thing.

AngeIII:
@Nextra , @Arkshine thanks, I got it 
```
Nobody wants to recompile a module to change such a simple thing.
```

PR #189, opened 2015-01-29
Add cs_create_entity and cs_find_ent_by_class natives

Arkshine:
Requested on the forum.

To know more, read the pawn include. :P

### DIFF
--- a/dlls/cstrike/cstrike/CstrikeNatives.cpp
+++ b/dlls/cstrike/cstrike/CstrikeNatives.cpp
@@ -1707,6 +1707,28 @@ static cell AMX_NATIVE_CALL cs_set_c4_defusing(AMX* amx, cell* params)
 	return 1;
 }
 
+extern CreateNamedEntityFunc CS_CreateNamedEntity;
+
+// cs_create_entity(const classname[])
+static cell AMX_NATIVE_CALL cs_create_entity(AMX* amx, cell* params)
+{
+	if (CS_CreateNamedEntity > 0)
+	{
+		int len;
+		int iszClass = ALLOC_STRING(MF_GetAmxString(amx, params[1], 0, &len));
+
+		edict_t *pEnt = CS_CreateNamedEntity(iszClass);
+
+		if (!FNullEnt(pEnt))
+		{
+			return ENTINDEX(pEnt);
+		}
+	}
+
+	return 0;
+}
+
+
 #else
 
 static cell AMX_NATIVE_CALL not_on_64(AMX* amx, cell* params)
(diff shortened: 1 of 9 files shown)


dlls/cstrike/cstrike/CstrikeNatives.cpp
```diff
@@ -1707,6 +1708,49 @@ static cell AMX_NATIVE_CALL cs_set_c4_defusing(AMX* amx, cell* params)
 	return 1;
 }
 
+extern CreateNamedEntityFunc CS_CreateNamedEntity;
+extern UTIL_FindEntityByStringFunc CS_UTIL_FindEntityByString;
+
+// cs_create_entity(const classname[])
+static cell AMX_NATIVE_CALL cs_create_entity(AMX* amx, cell* params)
+{
+	if (CS_CreateNamedEntity > 0)
```
Nextra: I think this has to throw an error instead of silently failing, or at the very least leave some trace of this happening in the error/amxx logs. We should indicate that AMXX is at fault and not the plugin. Same goes for cs_find_ent_by_class().

dlls/cstrike/cstrike/CstrikeNatives.cpp
```diff
@@ -1791,6 +1835,8 @@ AMX_NATIVE_INFO CstrikeNatives[] = {
 	{"cs_set_c4_explode_time",		cs_set_c4_explode_time},
 	{"cs_get_c4_defusing",			cs_get_c4_defusing},
 	{"cs_set_c4_defusing",			cs_set_c4_defusing},
+	{"cs_create_entity",			cs_create_entity },
```
xPaw: dem spaces before `}`

plugins/include/cstrike.inc
```diff
@@ -351,6 +351,47 @@ native bool:cs_get_c4_defusing(c4index);
 native cs_set_c4_defusing(c4index, bool:defusing);
 
 /**
+ * Creates an entity using Counter-Strike's custom CreateNamedEntity wrapper.
+ *
+ * @note Unlike other mods CS keeps track of entities using a custom hashtable.
+ *       This function will add entities to this hashtable, providing benefits
+ *       over the default engine functions (used by create_entity() for example):
+ *       - Storing entities in a hashtable allows CS to improve classname lookup
+ *         performance compared to functions like FindEntityByString that usually
```
xPaw: Instead of `FindEntityByString`, it probably would be better to say which amxx function it is (`find_ent_by_class`?)

Nextra: Since we can use the engine functions directly with FakeMeta we always have at least two ways to reach the same function. So instead of using the engine name I resorted to the "<real function name> (used by <engine module native>)" style. I could do that here too if you like.

xPaw: Two lines above it says "used by", so I think it would be better to keep it consistent.

Sylvert0:
In the file dlls/cstrike/cstrike/CstrikeHacks.cpp line 342 should not be if (!GiveShieldDetour)

Arkshine:
Oops.

xPaw:
:shipit:

Nextra:
:ship:

PR #190, opened 2015-01-29
Plmenu: small update: formatex, semicolon, missing things, remove useless TABs

Freeman-AM:


Freeman-AM:
True. I will see when i have time.

Freeman-AM:
You were talking about num_to_str, right ?

Well no, for num_to_str we can't it because on the other hand we will need to convert a string into a num in another base plugin. Furthermore "FOR_MIN" is used is used in approved plugins and modify it could break them a bit.

I think it don't worth it, and it should stayt as it is.

For me this PR is ok.

num_to_str should be faster than format or formatex for the asked task, (otherwise, why num_to_str exist ?)

Arkshine view about the question ?

Nextra:
Since we have official translation files up for usage (unlike SM) we can't change legacy translation strings like that. Either add a completely new string or keep the workaround.

xPaw:
Can we go back in time and slap whoever did this the wrong way? :)

Nextra:
It happened around 1.77 when `BAN_FOR_MIN` was split into `BAN` and `FOR_MIN`. It used to be a `%d` format. Oh well.

Freeman-AM:
Huh ? Talking about that, and thinking to it, i was searching a PR, this one, https://github.com/alliedmodders/amxmodx/pull/19 Well it wasn't released but in fact it could have been a bad idea.

Nextra:
Yes, a new language string should have been added there. We can not edit official translations like that.

Arkshine:
Looks ok.

PR #191, opened 2015-01-30
Fix tagof operator used as a default value for function arguments

Arkshine:
Someone on the forum reported some unexpected behavior about `tagof`, but it's well known that it's kind of broken from the start.

Looking at the Pawn 3.1.3522 changelog, It would seem I've missed a fix about it.

> BUG fix: the tagof operator used as a default value for function arguments
  took the tag of a variable, but not of an expression. Reported and fixed by
  Jonathan on the Pawn forum.

Old pawn forum seems dead, but I found it on a [webarchive](http://web.archive.org/web/20061025160224/http://www.compuphase.com/bitboard/index.php?DATEIN=tpc_vneqalxya_1133208472).

The fix fixes situations like:
```PAWN
func(42.1);
func(true);
func(MyTag:5);
...
func(any:variable, type = tagof variable)  // type would return always 0
{
}
```
Or `new Float:x` ; `tagof(x)` not having the same value `tagof(Float:)`

Since it was not working properly before, and someone asked about it, let's fix this.

### DIFF
--- a/compiler/libpc300/sc3.c
+++ b/compiler/libpc300/sc3.c
@@ -1243,6 +1243,7 @@ static int hier2(value *lval)
     clear_value(lval);
     lval->ident=iCONSTEXPR;
     lval->constval= val;
+    lval->tag=pc_addtag("bool");
     ldconst(lval->constval,sPRI);
     while (paranthese--)
       needtoken(')');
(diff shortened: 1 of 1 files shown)


xPaw:
@dvander wasn't really sure about `tagof` as a whole; but I think if the feature is already in there, might as well have it work correctly. :+1:

PR #192, opened 2015-02-01
Change MAX_PLAYERS value to 32

Arkshine:
It would seem that value set to `33` appears to be uncomfortable for people. 

Usually this define is always declared with `32` in plugins, because it's the technical limit, and a new player's array with `MAX_PLAYERS + 1`. There is nothing wrong doing `+ 1` and it explicits why we do that. Also, `32` can be used on others places, making the change a bit unwelcomed..

Let's revert to `32` which makes more sense in HL1/AMXX context.

### DIFF
--- a/plugins/admin.sma
+++ b/plugins/admin.sma
@@ -32,7 +32,7 @@ new PLUGINNAME[] = "AMX Mod X"
 #define ADMIN_IPADDR	(1<<3)
 #define ADMIN_NAME		(1<<4)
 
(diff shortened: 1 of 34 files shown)


Arkshine:
@xPaw Is there anything else I could have forgotten? I'm pretty sure we have all now.

xPaw:
Should be fine.

Arkshine:
@Nextra , @xPaw said it should be fine. :bee:

Nextra:
@xPaw has spoken. :sheep:

PR #193, opened 2015-02-01
Cstrike: Fix crash at server start

Arkshine:
Related to #189.

@stambeto2006 reported a crash at server start.
Searching `pev` offset (which calls `INDEXENT(0)`) on AMXX attach seemed to be too soon depending server.

Patch moves code on plugins loaded.


Nextra:
:shipit:

PR #196, opened 2015-02-06
Fix missing libm linkage under linux (math functions like floor, etc.)

Arkshine:
Some people seem to have some trouble under linux to load AMXX and get such error `undefined symbol: floorf`. The use of some math functions like `floor` adds a dependency of `libm` which seems to be absent in AMBuild script. 

On side not, `-lm` linkage is present in `MakeFile`!


dvander:
:ship:

PR #198, opened 2015-02-06
Fix various compatibility issue with geoip module

Arkshine:
* After the big update (#99) It would seem I forgot to keep the expected return on failed lookup for `geoip_country`, which is supposed to be "error" - as documentation says it. 
* Considering the retardness of this above native, a new `geoip_country_ex` native is added (`geoip_country` is reverted to its original state)
* Changed `printf` to `MF_PrintSrvConsole` to make sure messages appear in the console of a listen server.
* Cleaned up a bit files.

### DIFF
--- a/dlls/geoip/geoip_natives.cpp
+++ b/dlls/geoip/geoip_natives.cpp
@@ -117,6 +117,7 @@ static cell AMX_NATIVE_CALL amx_geoip_country(AMX *amx, cell *params)
 
 	if (!country)
 	{
+		MF_SetAmxString(amx, params[2], "error", params[3]); // to keep compatibility with previous version
 		return 0;
 	}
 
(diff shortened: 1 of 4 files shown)


dlls/geoip/geoip_natives.cpp
```diff
@@ -117,6 +117,7 @@ static cell AMX_NATIVE_CALL amx_geoip_country(AMX *amx, cell *params)
 
 	if (!country)
 	{
+		MF_SetAmxString(amx, params[2], "error", params[3]); // to keep compatibility with previous version
 		return 0;
```
Nextra: Incorrect return value for bcompat, the old version actually uses the `MF_SetAmxString` return value (and thus `strlen("error")`). I wouldn't consider this sensible behavior so maybe an `_ex` variant with a boolean return value is in order? 

Also it looks like multiple function docs need to be fixed as many say that `0` will be returned on unsuccessful lookup which is incorrect.

Nextra:
Otherwise lgtm.

PR #199, opened 2015-02-08
>> FIXED lookupString() LENGTH ISSUE

ClaudiuHKS:
In (const char *) lookupString() :
For (const char *) NULL result :
- Length must be ZERO, if defined.


ClaudiuHKS:
No longer needed.

Nextra:
... what?

PR #200, opened 2015-02-08
Set length to 0 on failed lookup in lookupString()

ClaudiuHKS:
In (const char *) lookupString() :
For (const char *) NULL result :

Length must be ZERO, if involved.

Also, renamed "maxminddb.cpp" to "maxminddb.c" inside Project's CFGs.

### DIFF
--- a/dlls/geoip/Makefile
+++ b/dlls/geoip/Makefile
@@ -15,7 +15,7 @@ PUBLIC_ROOT = ../../public
 
 PROJECT = geoip
 
-OBJECTS = amxxmodule.cpp GeoIP2/maxminddb.cpp geoip_main.cpp geoip_natives.cpp geoip_util.cpp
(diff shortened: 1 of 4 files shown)


Nextra:
In the future please force-push your branch and keep the PR up to date that way. There was no need to close #199 for this.

ClaudiuHKS:
I still have no idea how to properly add requests to AlliedModders' AMX Mod X, like others do.
I'm either stupid or GitHub is not well explaining.

Arkshine:
Thanks.

Nextra:
@ClaudiuHKS: Creating a pull request from your own GitHub fork doesn't mean that it stays as it is forever. You can add new commits to your branch or force-push a new history, the PR will automatically reflect the new state of the branch. This is more of a git thing than specific to GitHub. If you need help feel free to ask us. Thanks either way for spotting this :)

PR #204, opened 2015-02-13
Add two new natives to AMX Mod X

ClaudiuHKS:
I am willing to add two new natives, one as a Doctor Trophy dedicated for Arkshine and one for the people not willing to use FakeMeta just for reaching a Fake Client on their server.

In my opinion, CreateFakeClient function would look great into AMX Mod X's.

You also tell your opinion, and those with write access will take the decision.
Thanks.

(+) create_fake_client(const name[])
(+) bool:is_arkshine_a_doctor()

### DIFF
--- a/amxmodx/amxmodx.cpp
+++ b/amxmodx/amxmodx.cpp
@@ -4422,6 +4422,66 @@ static cell AMX_NATIVE_CALL is_rukia_a_hag(AMX *amx, cell *params)
 	return 1;
 };
 
+/** native bool:is_arkshine_a_doctor() */
+static cell AMX_NATIVE_CALL is_arkshine_a_doctor(AMX *amx, cell *params)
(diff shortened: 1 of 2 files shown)


Arkshine:
Thanks for the native, but we are trying to clean up AMXX not introducing more unnecessary stuffs! I appreciate the thought.

About `create_fake_client`, the invoked reason is really not a good one. There is no valid reason to not use a module when you need it. Encouraging this is quite the bad idea. I'm not sure adding this as native is useful as there are not much demands to create fake clients, and you would need such tool like Fakemeta anyway. At least adding it in core doesn't seem like worth since pfnCreateFakeClient is already exposed by Fakemeta. Maybe a stock for fakemeta could be okay.  Anyway, closing this, but thanks for proposing.

PR #206, opened 2015-02-15
Add new hashers and new natives

ClaudiuHKS:
Replace the only hasher called MD5 with the ones listed below.

* CRC32, MD5, SHA1, SHA256, SHA3 224 BIT, SHA3 256 BIT, SHA3 384 BIT,
SHA3 512 BIT, Keccak 224 BIT, Keccak 256 BIT, Keccak 384 BIT and Keccak
512 BIT.

Add the functions listed below.

* hash_string(const string[], HashType:type, output[], const
outputSize)
* hash_file(const fileName[], HashType:type, output[], const outputSize)

Add the following enumeration tag.

* HashType {}

Deprecate the following functions.

* md5()
* md5_file()

It has been tested on Windows and Linux. The sanity checks seems to be
properly working, so no worries about them.

These are useful if people are using Sockets, cURLs or MySQLs in order
to compare hashes of different files On-line for further investigation.
You are not able to check if the files are older or newer, but you can
see if the content is different (checksum mismatch).

The speed tests resulted that this MD5 currently hashes 5x faster than the old one.
Hashing de_dust2.bsp (approximately 2 MB) 128 times in 0.5 seconds rather than before, 1.0 seconds.

I'm glad I did this. Thanks to Stephan Brumme for their portable hashing library!


xPaw:
Looks interesting, however as it was discussed on IRC, adding new hidden natives is not something we want for a 1.8.3 release.

Also, is there any use for checking endian type in a amxx plugin?

Enums should follow CamelCase style, to match existing ones.

ClaudiuHKS:
We discuss, and when everything it's done and people have finished discussing, I'll do a cleanup and any other modifications to get it as you expect it.

Please rectify why not to add the Hidden native and the Endian type check (if there is an important reason so I can change my mind). I was thinking to add these two since they won't take much space (bytes) and won't use much RAM (bytes) and there are already such natives (Rukia, Hawk, ...). Arkshine really deserves that. Furthermore, SourceMod also use RandomSeed(), which is really not needed since RandomNum() and RandomFloat() already exist. I thought some people might find getSysEndianness() useful.

Thanks for considering it interesting. Rather than the old MD5 worker, it's a few more times easier to use and much more readable/ reachable. MD5 hash seems to execute 2x faster than before.

xPaw:
With 1.9 release, we (@arkshine and others) are aiming to remove unnecessary stuff from AMXX, and adding new ones that serve no purpose goes against that.

As for endian check, in my opinion it serves absolutely no purpose.

ClaudiuHKS:
Thank you. I am now learning, as you know, I was just a plug-in creator before. I hope I won't be bothering any longer. Willing to learn more about GitHub and stuff.

Thanks for answering shortly.
I did what you said, and yes, you are right.

xPaw:
It would be a good idea to move hashing files to `public/` folder as it is third-party code.

Also you should squash your commits: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request#squash-your-changes-optional

amxmodx/CLang.cpp
```diff
@@ -318,8 +318,9 @@ void reparse_color(String* def)
 // -- BAILOPAN
 int CLangMngr::MergeDefinitionFile(const char *file)
 {
-	FILE *fp = fopen(file, "rt");
-	if (!fp)
+	const char* md5buffer = hashFile((const char *)&file[0], Hash_Md5);
```
Arkshine: Is there a reason you're using `&file[0]` instead of passing directly `file`?

amxmodx/CLang.cpp
```diff
@@ -331,15 +332,10 @@ int CLangMngr::MergeDefinitionFile(const char *file)
 				break;
 			}	
 		}
-		AMXXLOG_Log("[AMXX] Failed to open dictionary file: %s", file);
+		AMXXLOG_Log("[AMXX] There was a problem opening or hashing dictionary file: %s", file);
```
Arkshine: Keep the original message error. At this point, it can be only that since file is unlikely to be empty and you pass manually the constant.

amxmodx/amxmodx.cpp
```diff
@@ -1438,44 +1438,75 @@ static cell AMX_NATIVE_CALL get_plugin(AMX *amx, cell *params) /* 11 param */
...
-	char *str = get_amxstring(amx, params[1], 0, len);
-	char buffer[33];
+	int strLen;
+	char* pStr = get_amxstring(amx, params[1], 0, strLen);
 
-	MD5 md5;
-	md5.update((unsigned char *)str, len);
-	md5.finalize();
-	md5.hex_digest(buffer);
+	const char* pHash = hashString((const char *)&pStr[0], strLen, Hash_Md5);
```
Arkshine: Same, why `&pStr`?

amxmodx/amxmodx.cpp
```diff
@@ -1438,44 +1438,75 @@ static cell AMX_NATIVE_CALL get_plugin(AMX *amx, cell *params) /* 11 param */
...
+	build_pathname_r(fileLocation, sizeof(fileLocation) - 1, "%s", pFileName);
 
-	FILE *fp = fopen(file, "rb");
-	
-	if (!fp)
+	const char* pHash = hashFile((const char *)&fileLocation[0], Hash_Md5);
+	if (!pHash)
 	{
-		LogError(amx, AMX_ERR_NATIVE, "Cant open file \"%s\"", file);
+		LogError(amx, AMX_ERR_NATIVE, "There was a problem hashing file (\"%s\"). " \
```
Arkshine: Same here, keep the original message.

amxmodx/amxmodx.cpp
```diff
@@ -1438,44 +1438,75 @@ static cell AMX_NATIVE_CALL get_plugin(AMX *amx, cell *params) /* 11 param */
...
+{
+	int strLen;
+	char* pStr = get_amxstring(amx, params[1], 0, strLen);
+	hashType Type = (hashType)params[2];
+
+	const char* pHash = hashString((const char *)&pStr[0], strLen, Type);
+	if (!pHash) /** Scripter's fault. Let the server still run. */
+	{
+		LogError(amx, AMX_ERR_NATIVE, "Unable to hash string (\"%s\"). " \
+			"Perhaps \"Type\" parameter isn't valid.", pStr);
```
Arkshine: I think you can remove "Perhaps.." sentence.

amxmodx/amxmodx.cpp
```diff
@@ -1438,44 +1438,75 @@ static cell AMX_NATIVE_CALL get_plugin(AMX *amx, cell *params) /* 11 param */
...
+
+	char fileLocation[256];
+	build_pathname_r(fileLocation, sizeof(fileLocation) - 1, "%s", pFileName);
+
+	hashType Type = (hashType)params[2];
+
+	const char* pHash = hashFile((const char *)&fileLocation[0], Type);
+	if (!pHash)
+	{
+		LogError(amx, AMX_ERR_NATIVE, "There was a problem hashing file (\"%s\"). \
```
Arkshine: Be consistent, use the same message as the previous native, and yes remove the "perhaps" sentence.

amxmodx/endianness.cpp
```diff
@@ -0,0 +1,18 @@
+#include <endianness.h>
+
+/**
+* Retrieves system endianness.
+*/
+sysEndianness getSysEndianness(void)
```
Arkshine: Why do you need actually this? Was there an issue with the defines?

amxmodx/hashing.cpp
```diff
@@ -0,0 +1,1963 @@
+// Copyright (c) 2014 Stephan Brumme. All rights reserved.
```
Arkshine: I've looked a bit more what changes you did.
I think if we are going to provide a third-party library, this needs to be used as it is, unless there are necessary changes. Here removing std::string dependency is fine, but you should change only that.

So, this would be likely more appropriate to:
* Keep the original files organizations
* Move them in `public/hashlib` or similar name

About the `std::string` changes, did you test the safety?
You might as well use `ke::AString`.

plugins/include/amxconst.inc
```diff
@@ -450,3 +450,27 @@ enum AdminProp
 	AdminProp_Access,
 	AdminProp_Flags
 };
+
+/**
+ * HashType
```
Arkshine: Be consistent.

```
/**
 * HashType constants
 * To be used on hash_file() and hash_string()
 */

File: plugins/include/amxmodx.inc [hunk 10]
```diff
@@ -2321,9 +2322,35 @@ native md5(const szString[], md5buffer[34]);
  * @return              Number of cells written to the buffer (always 32)
  * @error               If the file can not be opened, and error is thrown.
  */
+#pragma deprecated Use now hash_file() function. See INC files for more info.
 native md5_file(const file[], md5buffer[34]);
 
 /**
```
Arkshine: Keep it consistent. 

* Use spaces for alignment.
* About `type` description, I would say: `Hash type, see Hash_* constants in amxconst`

plugins/testsuite/hashing_test.sma
```diff
@@ -0,0 +1,53 @@
+// vim: set ts=4 sw=4 tw=99 noet:
+//
+// AMX Mod X, based on AMX Mod by Aleksander Naszko ("OLO").
+// Copyright (C) The AMX Mod X Development Team.
+//
+// This software is licensed under the GNU General Public License, version 3 or higher.
+// Additional exceptions apply. For full license details, see LICENSE.txt or visit:
+//     https://alliedmods.net/amxmodx-license
+
```
Arkshine: This file should be also inserted in the list in `support/AMBuildScript`

I would also do some changes:
* Use a server commands
* Don't use `static`
* No need to initialize
* Output wtih `server_print` or `log_amx`

ClaudiuHKS:
OK Arkshine, I will do the changes shortly.
I think I already explained in the first post I've tested it under Windows and Linux more than I should and everything's working like a charm. I compared AMX Mod X hashing checksums with other On-line hashing websites, and the they were the same.

What will I do in the next commit?

* Move Hashers to 'Public/'.
* Change &pointer[0] to pointer for readability.
* Leave error messages as they were (no change).
* I will be more consistent in .INC files.
* "Static" is required in a few places. I will try to get rid of it as much as possible.
* I will get rid of endianness.cpp file, but getSysEndiannes() will be inside hashing.cpp (locally). I consider it the safest way to get the byte order properly.
* Definitely change code inside 'hashing_test.sma'.

plugins/testsuite/hashing_test.sma
ClaudiuHKS: Fixed, thanks for the info. Didn't know about that BuildScript.

plugins/include/amxmodx.inc
ClaudiuHKS: Fixed.

plugins/include/amxconst.inc
ClaudiuHKS: Yes.

amxmodx/endianness.cpp
ClaudiuHKS: Better to get the system endianness (byte order) while processing data in this case,

amxmodx/hashing.cpp
ClaudiuHKS: They have been transferred to Public/. Impossible to use ke::AString.

amxmodx/CLang.cpp
ClaudiuHKS: Not really as I see. Let's keep things as they were, rolling back action, change &Var[0] to Var.

Arkshine:
> Regarding ke:::AString. Impossible. It has no += operator. C strings seems to work fine.

I told you already there is `AutoString`, which is based on `ke::AString` and does that.
I'm not saying to use it, but if you are not absolutely sure using static variable will be fine, then you should use it.

> Otherwise, it will be a problem if someone will include them separately w/ o including "hashing.h" file, plus you have to define uint8_t, uint32_t and uint64_t types in each one. Too much headache, folders and used space. Also, the respective owner will never update their libraries, and merging them together is OK

I don't see where is the issue.It doesn't matter if it uses several files, spaces or whatever. You miss the point it's important to keep as much as the original for maintainability. If author updates, it's going to be a massive pain to update it in AMXX. About defining, you can just create an extra .h and adding it where you need it.

ClaudiuHKS:
Alright, changing in a moment. Thanks for taking your time.

plugins/include/amxmodx.inc
```diff
@@ -2310,6 +2310,7 @@ native force_unmodified(force_type, const mins[3], const maxs[3], const filename
  *
  * @return              Number of cells written to the buffer (always 32)
  */
+#pragma deprecated Use now hash_string() function. Also, see Hash_* constants.
```
xPaw: Remove word `now`.

plugins/include/amxmodx.inc
```diff
@@ -2321,9 +2322,35 @@ native md5(const szString[], md5buffer[34]);
  * @return              Number of cells written to the buffer (always 32)
  * @error               If the file can not be opened, and error is thrown.
  */
+#pragma deprecated Use now hash_file() function. Also, see Hash_* constants.
```
xPaw: Same here.

plugins/include/amxmodx.inc
```diff
@@ -2321,9 +2322,35 @@ native md5(const szString[], md5buffer[34]);
...
+ * Hashes a string.
+ *
+ * @param string        String to hash.
+ * @param type          Type of hash. See Hash_* constants in amxconst.inc file.
+ * @param output        Output string to store hash in.
+ * @param outputSize    The maximum size of the output string to store hash in.
+ *
+ * @return              Number of written bytes.
+ */
+native hash_string(const string[], const HashType:type, output[], const outputSize);
```
xPaw: I would copy PHP's description for this method:
```
 /**
 * Generate a hash value (message digest)
 *
 * @param string        String to be hashed.
 * @param type          Type of selected hashing algorithm. See Hash_* constants in amxconst.inc file.
 */
```

plugins/include/amxmodx.inc
```diff
@@ -2321,9 +2322,35 @@ native md5(const szString[], md5buffer[34]);
...
+ *
+ * @param fileName      File to hash.
+ * @param type          Type of hash. See Hash_* constants in amxconst.inc file.
+ * @param output        Output string to store hash in.
+ * @param outputSize    The maximum size of the output string to store hash in.
+ *
+ * @return              Number of written bytes.
+ * @error               If the file couldn't be opened, an error is thrown.
+ */
+native hash_file(const fileName[], const HashType:type, output[], const outputSize);
```
xPaw: I would copy PHP's description for this method:
```
 /**
 * Generate a hash value using the contents of a given file
 *
 * @param string        Path to file to be hashed. (might need to word this better!)
 * @param type          Type of selected hashing algorithm. See Hash_* constants in amxconst.inc file.
 */
```

public/hashing/hashing.cpp
```diff
@@ -0,0 +1,170 @@
...
+	SHA256 Sha256;
+	SHA3 Sha3;
+	Keccak Kec;
+
+	/**
+	 * Changes bits if needed
+	 */
+	switch (Type)
+	{
+	case Hash_Sha3_224:		Sha3.changeBits(SHA3::Bits224);		break;
```
xPaw: Needs extra tabs.

public/hashing/hashing.cpp
```diff
@@ -0,0 +1,170 @@
...
+	/**
+	 * Retrieves file's content and fills hashers in.
+	 */
+	char Bytes[8192];
+	size_t Count;
+	while ((Count = fread(Bytes, sizeof(char), sizeof Bytes, pFile)))
+	{
+		switch (Type)
+		{
+		case Hash_Crc32:		Crc32.add(Bytes, Count);	break;
```
xPaw: Needs extra tabs.

public/hashing/hashing.cpp
```diff
@@ -0,0 +1,170 @@
...
+	 * Closes file
+	 */
+	fclose(pFile);
+
+	/**
+	 * Retrieves hash
+	 */
+	switch (Type)
+	{
+	case Hash_Crc32:		return Crc32.getHash();
```
xPaw: Needs extra tabs.

public/hashing/hashing.cpp
```diff
@@ -0,0 +1,170 @@
...
+	SHA256 Sha256;
+	SHA3 Sha3;
+	Keccak Kec;
+
+	/**
+	 * Changes bits if needed
+	 */
+	switch (Type)
+	{
+	case Hash_Sha3_224:		Sha3.changeBits(SHA3::Bits224);		break;
```
xPaw: Needs extra tabs.

public/hashing/hashing.cpp
```diff
@@ -0,0 +1,170 @@
...
+	case Hash_Keccak_384:	Kec.changeBits(Keccak::Keccak384);	break;
+	case Hash_Keccak_512:	Kec.changeBits(Keccak::Keccak512);	break;
+	};
+
+	/**
+	 * Fills hashers in, computes string hash and returns the hash
+	 */
+	switch (Type)
+	{
+	case Hash_Crc32:		return Crc32(String, stringLen);
```
xPaw: Needs extra tabs.

Arkshine:
Looks good now. Thanks.

:ship:

PR #207, opened 2015-02-17
Fix SQL Modules Threading

ClaudiuHKS:
Fix SQL modules threading because of the problem(s) listed below.

* The game server is having problems when too many threaded queries are started and the connection between the game server and the MySQL remote server is lost. Problems consist in 'queueTime' (Float) higher than 5.0 seconds, up to ...

* The server is having problems when map changes (during OnPluginsUnloading() AMX Mod X Core Forward) if the connection is lost or too many threaded queries are started. It just freezes to execute all the remaining threaded queries, and some OS such Linux are killing processes that are frozen for 5-10-15 seconds, unfortunately.

What can I say about my changes?

* Changing the SleepTime (milliseconds) from 250...500 to 25, because SourceMod uses the same (they even use 20 milliseconds).
* Reducing the interval of SQL Threaded Queries check in StartFrame() Meta Mod Forward.
* Making sure so the BaseThreader executes much more threads when OnAmxxDetach and OnPluginsUnloading AMX Mod X Core Forwards are reached so everything is getting flushed amazingly fast when the map ends.
* Also, making sure that everything is going to be executed, even with Internet connection or without.
* Changed WinSock to WinSock2 in MySQL module.

Did I test that?

You can see a SQL Stress Test plug-in I made for myself, I will not include that as a "Test Suite" because it's too abusive.
http://pastebin.com/UZmVhpQA

You can see that even spamming, everything is going to be executed safe.

### DIFF
--- a/dlls/sqlite/threading.cpp
+++ b/dlls/sqlite/threading.cpp
@@ -28,6 +28,8 @@ void ShutdownThreading()
 {
 	if (g_pWorker)
 	{
+		// Flush all the remaining job fast!
+		g_pWorker->SetMaxThreadsPerFrame(8192);
 		g_pWorker->Stop(true);
 		delete g_pWorker;
 		g_pWorker = NULL;
@@ -293,7 +295,7 @@ void OnPluginsLoaded()
 		g_QueueLock = g_Threader.MakeMutex();
 	}
 
-	g_pWorker = new ThreadWorker(&g_Threader, 250);
+	g_pWorker = new ThreadWorker(&g_Threader, DEFAULT_THINK_TIME_MS);
 	if (!g_pWorker->Start())
 	{
 		delete g_pWorker;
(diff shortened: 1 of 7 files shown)


dvander:
In the future these should all be separate PRs as they are separate issues.

dlls/sqlite/threading.cpp
```diff
@@ -310,7 +312,7 @@ void StartFrame()
 {
 	if (g_pWorker && (g_lasttime < gpGlobals->time))
 	{
-        g_lasttime = gpGlobals->time + 0.05f;
+        g_lasttime = gpGlobals->time + 0.025f;
```
dvander: Nit: fix identation here, and in the other extension like it.

Arkshine:
Looks fine. Will ask people on forum to test it to confirm it helps.

:ship:

PR #209, opened 2015-02-17
Change AMX Mod X Lang Files Hash Method

ClaudiuHKS:
Change AMX Mod X Language Files hashing method from MD5 to CRC32 because it is 2x or even 3x faster and more reliable for that type of hash. Not for security, but for comparing files' content difference.

### DIFF
--- a/amxmodx/CLang.cpp
+++ b/amxmodx/CLang.cpp
@@ -318,15 +318,15 @@ void reparse_color(String* def)
 // -- BAILOPAN
 int CLangMngr::MergeDefinitionFile(const char *file)
 {
-	const char* md5buffer = hashFile(file, Hash_Md5);
-	if (!md5buffer)
+	const char* hashBuffer = hashFile(file, Hash_Crc32);
+	if (!hashBuffer)
 	{
-		CVector<md5Pair *>::iterator iter;
+		CVector<CRC32Pair *>::iterator iter;
 		for (iter = FileList.begin(); iter != FileList.end(); ++iter)
 		{
 			if ((*iter)->file.compare(file) == 0)
 			{
-				char buf[33] = {0};
+				char buf[8 /* CRC32 Length */ + 1] = { 0 };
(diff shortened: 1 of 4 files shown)


xPaw:
What about replacing `FileList`/`CRC32Pair` with a `HashMap` from AMTL, I think that would be better.

ClaudiuHKS:
Yes, it will be a good idea, but I need now to get rid of that "MySQLThreadingFix" commit made on this PR by mistake.

xPaw:
If you are going to work more on this, it would be worth looking into using AMTL for other lookup tables as well (for `LookUpVec`, `m_LookUpTable` and some others).

Nextra:
This whole mechanism should be redone as it isn't working properly and is mostly pointless, so this is just a band-aid fix that doesn't tackle the underlying problem. You should expand the scope of your changes (feel free to contact me for details) or close the PR if you don't want to do so. Judging from what I see surrounding this whole stuff I wouldn't feel comfortable merging this short-sighted improvement.

ClaudiuHKS:
Of course I will do some updates in here as Arkshine and xPaw recommended.
But, I don't get Nextra's message.

"You should expand the scope of your changes (feel free to ask around for details) or close the PR if you don't want to do so."

How can I expand the scope of my changes?

Nextra:
What I mean by that is that the PR is currently focused on changing the hash function for language files. I don't think this is a good change, as the whole hashing mechanism is currently broken and arguably pointless. If you wanted to change anything about this particular section of code you should *expand the scope of this PR* to include much more than just changing the hash function.

ClaudiuHKS:
Exactly. Because this Pull Request was not mergeable due to a branch problem, it is going to be now closed. A new Pull Request is now opened, that recommends not using Language Files hashing anymore, but a more efficient method, .st_mtime comparison.

Also, some unused code will be removed in the opened Pull Request.

PR #208, opened 2015-02-17
Update include documentation #2

Nextra:
Second grind at our outdated/non-existent/inconsistent include documentation (see #108). Will start in alphabetical order and see when I get tired of doing it :)

Early work in progress PR, feedback welcomed. ~~Documentation ideas for `constraint_offset` in amxmisc.inc very welcomed.~~

Thanks to @xPaw, @Freeman-AM and of course @Arkshine for helping out :sun_with_face:

PR #210, opened 2015-02-17
Fix ArrayDestroy/DestroyStack erroring on invalid handle

Nextra:
Allow passing 0/Invalid_* to ArrayDestroy and DestroyStack because it makes sense (and in the case of ArrayDestroy is also legacy behavior, so a must). Updated documentation lands with #208


xPaw:
:+1:

PR #185, opened 2015-01-28
Introduce new features for cvars

Arkshine:
New features are:

<table>
    <thead>
        <tr>
            <th> Description </th>
            <th align="center"> New natives </th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td> Hooking a cvar change </td>
            <td>  
                <code>hook_cvar_change</code><br />
                <code>enable_cvar_hook</code><br />
                <code>disable_cvar_hook</code><br />
            </td>
        </tr>
        <tr>
            <td>Binding a cvar to a global variable (caching)</td>
            <td>  
                <code>bind_cvar_num</code><br />
                <code>bind_cvar_float</code><br />
                <code>bind_cvar_string</code><br />
            </td>
        </tr>
        <tr>
            <td>Creating cvar with more information (description and bounds)</td>
            <td>  
                <code>create_cvar</code>
            </td>
        </tr>
        <tr>
            <td>Defining a cvar bounds</td>
            <td>  
                <code>set_pcvar_bounds</code><br />
                <code>get_pcvar_bounds</code><br />
            </td>
        </tr>
        <tr>
            <td>Adding boolean natives for convenience and readability</td>
            <td>  
                <code>set_pcvar_bool</code><br />
                <code>get_pcvar_bool</code><br />
            </td>
        </tr>
    </tbody>
</table>

To reflect new additions:
* `get_plugins_cvar` can retrieve cvar description
* `amxx cvars` command has been extended to show more status (binded/hooked/bounded) and by providing an index from listing, it can show detailed information and from what plugins cvar is handled (useful for troubleshooting).

Worth to be noted:
* Cvars handled by AMXX are now cached and cache is now used by most of natives. It should give some welcomed performance boost.
* The function hook used to catch cvar change is disabled by default and also at each map change. So it should be enabled only when necessary.

PR #211, opened 2015-02-19
Fix issue where at mapchange a cvar current value is overwritten by defined bounds where it should not

Arkshine:
Reported on the [forum](https://forums.alliedmods.net/showpost.php?p=2263610&postcount=8), related to #185.

If a cvar has bounds defined in `create_cvar`, at map change, even though current value is not out of bounds, cvar was set again with the defined bounds, resulting overwriting the current value.

This is happening because:
* Saved bounds values are reset at map change to reflect any changes in plugin and `SetCvarMin/Max` was relying on these values.
* No checks are done to verify if a value is already in the expected bounds.

Patch refactors a bit some code:
* Since bounds are set after cvar creation, no need to pass such datas `CreateCvar`
* The invalid bounds check is done now in native directly before calling `SetCvarMin/Max`
* A check has been added in `SetCvarMin/Max` to not set cvar if value is already in range.

### DIFF
--- a/amxmodx/CvarManager.cpp
+++ b/amxmodx/CvarManager.cpp
@@ -155,7 +155,7 @@ void CvarManager::CreateCvarHook(void)
 }
 
 CvarInfo* CvarManager::CreateCvar(const char* name, const char* value, const char* plugin, int pluginId, int flags,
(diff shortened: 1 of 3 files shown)


Nextra:
lgtm. Thanks for the clean-up. Updated/Detailed docs land with #208

PR #212, opened 2015-02-19
Replace compile.exe with compile.bat

xPaw:
This removes 10 year old Deplhi thing, and replaces with a much smaller batch script.
However this means losing feature where compile.exe would check timestamps and not recompile plugins if they were not modified. Seeing as compile.sh doesn't have this feature either, I'm not sure if it's worth trying to implement it in batch.

This PR also brings compile.sh up to date.

### DIFF
--- a/plugins/compile.sh
+++ b/plugins/compile.sh
@@ -28,7 +28,7 @@ if [[ $# -ne 0 ]]; then
 		fi
 	done
 else
-	for sourcefile in *.sp
+	for sourcefile in *.sma
 	do
 		outfile="$(echo "$sourcefile" | sed -e 's/\.sma$/\.amxx/')"
 		echo -e "// Compiling $sourcefile ..."


ClaudiuHKS:
Looks amazing. I am with you.
If possible, changing the icon of .bat file with favorite.ico would be great.

http://www.makeuseof.com/answers/change-icon-bat-file-windows-7/

xPaw:
It's not possible to directly modify the icon, and I'm not going to ship a shortcut or bat-to-exe converter.

Nextra:
Even though I wouldn't mind and the .sh doesn't provide it either, this is a step backwards in functionality. I'd prefer having the old functionality in the .bat and upgrading the .sh to do the same.

Arkshine:
I would agree with Nextra too. It would make more sense this way.

xPaw:
I've came up with the idea of checking timestamp of source file against compiled file, instead of storing it `compile.dat` file. However it's still rather hard to implement in batch.

Nextra:
Yeah, batch is terrible for any kind of date manipulation. While the Delphi thing might not be pretty - it works. Having it around doesn't hurt anyone as it is maintenance free. Without a proper alternative we should not get rid of it.

Freeman-AM:
However this means losing feature where compile.exe would check timestamps and not recompile plugins if they were not modified.

> -1 for me, just because of that.

xPaw:
See my comment above about timestamps, I thought about adding a new parameter in amxxpc which would accomplish this.

xkpx64:
Add some protection that make harder dissambly of .amxx.
This is major thing to do!

Nextra:
Making .amxx harder to disassemble is the very very very very very very very very last thing on any priority list that anyone has ever thought of regarding AMXX. But thanks for the input.

PR #213, opened 2015-02-19
Replace CLang file hashing with .st_mtime

ClaudiuHKS:
Replace CLang file hashing with .st_mtime for performance.
Also, fix a problem in CLangMngr::MergeDefinitionFile.

Thanks Arkshine for their tutorials of using StringHashMap.
He really contributed very hard on this CLang change.

### DIFF
--- a/amxmodx/CLang.h
+++ b/amxmodx/CLang.h
@@ -11,18 +11,13 @@
 #define _INCLUDE_CLANG_H
 
 #include "sh_tinyhash.h"
+#include "sm_stringhashmap.h"
 
 #define LANG_SERVER 0
 #define LANG_PLAYER -1
 
 #define ERR_BADKEY	1	// Lang key not found
 #define ERR_BADLANG 2	// Invalid lang
-
-struct md5Pair
-{
-	String file;
-	String val;
-};
 
 struct sKeyDef
 {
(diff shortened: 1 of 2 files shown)


Arkshine:
Much better. (I'm one who asked him to redo PR/branch, because previous branch was completely fucked up, and I failed to help him to clean it up. It was just more fast and simple to redo the PR).

amxmodx/CLang.h
```diff
@@ -118,7 +113,7 @@ class CLangMngr
 	
 	LangVec m_Languages;
 
-	CVector<md5Pair *> FileList;
+	StringHashMap<time_t> FileList;
 	CVector<String *> KeyList;
 	THash<String, keytbl_val> KeyTable;
```
xPaw: You should try merging KeyList/KeyTable into a single HashTable from AMTL, which will mean tinyhash class will no longer be needed.

Arkshine: I would do that in another PR. The following patch is just to fix the shit code on this specific chunk of code.
I will make a PR to clean the whole class.

ClaudiuHKS: Exactly. xPaw, I'm not familiarized with amtl. Really, I'd appreciate Arkshine's hands doing these amtl implementations. I will help any other ways, if needed, for now.

Agree, that CLang code really needed a change.

Arkshine:
`stat` should be fine under windows, since both `stat` and `_stat` seems to point to the the same internal function.
 
:palm_tree:

Nextra:
ltgm then

PR #215, opened 2015-02-24
Remove version checks from all include files

xPaw:


Freeman-AM:
may this cause compatibility problems with include update ?
I mean having both 2 PR touching same files.

Arkshine:
Should be okay.

PR #216, opened 2015-03-04
Some missing charsmax and more readability.

luxxxoor:


ClaudiuHKS:
Seems good, getting rid of hardcoded **stringMaxSize** parameter.
By the way, you've missed **get_user_name(player, arg, 31)**  @ **AdminVote.SMA** script.

luxxxoor:
Yea...I looked only at admincmd.sma.

ClaudiuHKS:
Wait... what do you mean *Maybe FormatEx is better?*?
You need to test everything before committing. Now you've already got 3 commits and a few typos for a few lines.

If you know **FormatEx** is better **and yes, it is**, do the test and if everything works properly commit a **FormatEx** change as well.

Arkshine:
Thanks for contributing luxxxoor. 
It would be a good idea you test what you're doing before committing and pushing. In your last commit, you write `charsMax` instead of `charsmax`. Something you would have seen if you would have compiled plugin before going further.
Also, as ClaudiuHKS said, you should know what you're doing, here when to use `formatex` (which is fine here), but may not really related to the current PR. This means also you should avoid to insert such comment in plugin.

ClaudiuHKS:
So, not to leave you uninformed, my opinion is to erase that **Maybe FormatEx is better** comment and replace **charsMax** with **charsmax** so this Pull Request can be merged.

**charsmax** and **formatex** are **separated issues** and **require separated pull requests** as well.

luxxxoor:
Sorry, I'm still a newbie.. next time I will test it.

Freeman-AM:
Helping amxmodx development require at least to know what the tool you use do.
Especially for the case of formatex/format replacement.

luxxxoor:
I read about the difference between format and formatex.. i only proposed to change format with formatex where appropriate

Arkshine:
@luxxxoor Could you try to squash all your commits into one, if you know how? This is to avoid having unnecessary noise in the history while your commits are just kind of typos.

luxxxoor:
I don't know how, but I can submit a new commit with the correct changes

Arkshine:
You can probably revert locally your commits, then create one commit and pushing with --force.

A fast way is to rebase too, like you do: $ git rebase -i HEAD~4
You will see a list of commits starting with "pick". Just keep the first "pick" and replace the 3 others by "squash".  Save. Then, do: $ git push origin --force

Arkshine:
Thanks!

PR #217, opened 2015-03-09
Add support for Valve FS to natives that use file handles

Arkshine:
## Introduction

Purposes to make more operations possible with files via Valve File System through Valve search paths. 
Interface is available from HLSDK: [FileSystem.h](https://github.com/alliedmodders/hlsdk/blob/master/public/FileSystem.h)
This is handy when you want to find files which are not necessary existing directly in gamedir.

This is similar to what did psychonic for SM:  https://github.com/alliedmodders/sourcemod/pull/120.

## Valve paths
In Half-Life 1, paths and their associated IDs are sets up in the engine. This is rather straight but adapted to HL1 server. For reference, engine registers such paths in priority order:
<table>
  <tbody>
    <tr>
      <th align="center">Path ID</th>
      <th align="center">Description</th>
    </tr>
    <tr>
      <td align="right"><code>GAME</code></td>
      <td align="left">All paths related to current mod, including fallback<br />
Depending settings, it includes: <code>$gamedir_lv</code>/<code>_addon</code>/<code>_$language</code>/<code>_hd</code> and <code>$gamedir</code> itself</td>
    </tr>
    <tr>
      <td align="right"><code>GAMECONFIG</code></td>
      <td align="left">The default writable directory (<code>$gamedir</code>)</td>
    </tr>
    <tr>
      <td align="right"><code>GAMEDOWNLOAD</code></td>
      <td align="left">The download directory (<code>$gamedir_download</code>)</td>
    </tr>
    <tr>
      <td align="right"><code>GAME_FALLBACK</code></td>
      <td align="left">All paths related to fallback game, same as <code>GAME</code> (e.g. <code>czero</code> has <code>cstrike</code> as fallback</td>
    </tr>
    <tr>
      <td align="right"><code>DEFAULTGAME</code></td>
      <td align="left">All paths related to the default game which is "valve", same as <code>GAME</code></td>
    </tr>
    <tr>
      <td align="right"><code>BASE</code></td>
      <td align="left">The base path where server is installed</td>
    </tr>
  </tbody>
</table>

Note that Valve FS supports a null path ID, which means "search in all paths". For that purpose, a fake  `ALL` path ID can be used instead of using an empty string, for readability. Native will simply checks if `ALL` is passed and will use a null pointer in valve function.

## Support

Some natives supports partially Valve FS, where a path ID can't be provided and this will search in all paths.

Here a a list of natives supporting Valve FS:
* `delete_file`, `unlink`
* `file_exists` - _partially_
* `dir_exists` - _partially_
* `file_size`
* `fopen`
* `mkdir`
* `open_dir`

Some natives are not adjusted:
* `read_dir`, `read_file`, `write_file` - _I did not bother to add support for those old and and bad designed natives_
* `rename_file` - _This is not exposed by Valve FS_
* `filesize` - _No new params can be added_
* `rmdir` - _Not exposed by Valve FS_

## New natives and minor changes

Some new natives stolen from SM:
* `SetFilePermissions` - _To change a file or directories permissions_
* `FileRead*` and `FileWrite*` - _More sane and better designed natives to read/write a single value (proper type, unsigned byte/short support and returns 1 or 0)_

There are as well some minor additions or changes:
* Added a new define `PLATFORM_MAX_PATH` to represent a maximum path length
* Added  `any:` tag to some natives
* Include file is now properly documented
* `read_dir` - _`outlen` parameter is now optional_
* `read_file` - _`txtlen` parameter is now optional. Adjusted as well buffer to `2048` to match buffer size in `write_file`_
* `file_size` - _Fixed potential issue with old compiled plugin which doesn't have the flag parameter._ _Added `FSOPT_*` _constants as well for use with this flag parameter._
* `fputs` - _a new parameter has been added to choose whether you want to append a null terminator_
* `mkdir` -  _a new parameter has been added to set permissions mode (linux only)_
* `open_dir`, `next_file` - _a new parameter has been added to retrieve an entry type (if file or directory)_

### DIFF
--- a/amxmodx/CFileSystem.h
+++ b/amxmodx/CFileSystem.h
@@ -79,14 +79,14 @@ class ValveFile : public FileObject
 
 		static bool Delete(const char* filename, const char* pathID) 
 		{
-			if (!g_FileSystem->FileExists(filename))
+			if (!Exists(filename))
 			{
 				return false;
 			}
 
 			g_FileSystem->RemoveFile(filename, pathID);
 
-			if (g_FileSystem->FileExists(filename))
+			if (Exists(filename))
 			{
 				return false;
 			}
--- a/amxmodx/CFileSystem.h
+++ b/amxmodx/CFileSystem.h
@@ -1,0 +1,256 @@
+// vim: set ts=4 sw=4 tw=99 noet:
+//
+// AMX Mod X, based on AMX Mod by Aleksander Naszko ("OLO").
+// Copyright (C) The AMX Mod X Development Team.
+//
+// This software is licensed under the GNU General Public License, version 3 or higher.
+// Additional exceptions apply. For full license details, see LICENSE.txt or visit:
+//     https://alliedmods.net/amxmodx-license
+
+#ifndef _INCLUDE_CFILESYSTEM_H_
+#define _INCLUDE_CFILESYSTEM_H_
+
+#include <FileSystem.h> // IFileSystem, FileSystemSeek_t, FileHandle_t (HLSDK)
+#include <stdio.h>      // FILE*
+#include <am-cxx.h>     // KE_OVERRIDE
+
+extern IFileSystem* g_FileSystem;
+
+class ValveFile;
+class SystemFile;
+
+class FileObject
+{
+	public:
+
+		virtual ~FileObject() {};
+
+		virtual size_t Read(void* pOut, size_t size) = 0;
+		virtual char* ReadLine(char* pOut, size_t size) = 0;
+		virtual size_t Write(const void* pData, size_t size) = 0;
+
+		virtual bool Seek(int pos, int seek_type) = 0;
+		virtual int Tell() = 0;
+		virtual int Flush() = 0;
+
+		virtual bool HasError() = 0;
+
+		virtual bool EndOfFile() = 0;
+		virtual void Close() = 0;
+
+		virtual ValveFile *AsValveFile() 
+		{
+			return nullptr;
+		}
+
+		virtual SystemFile *AsSystemFile()
+		{
+			return nullptr;
+		}
+};
+
+class ValveFile : public FileObject
+{
+	public:
+
+		ValveFile(FileHandle_t handle) : handle_(handle) {}
+
+		~ValveFile() 
+		{
+			Close();
+		}
+
+		static bool Exists(const char* file)
+		{
+			return g_FileSystem->FileExists(file);
+		}
+
+		static ValveFile* Open(const char* filename, const char* mode, const char* pathID) 
+		{
+			FileHandle_t handle = g_FileSystem->OpenFromCacheForRead(filename, mode, pathID);
+			
+			if (!handle)
+			{
+				return nullptr;
+			}
+
+			return new ValveFile(handle);
+		}
+
+		static bool Delete(const char* filename, const char* pathID) 
+		{
+			if (!g_FileSystem->FileExists(filename))
+			{
+				return false;
+			}
+
+			g_FileSystem->RemoveFile(filename, pathID);
+
+			if (g_FileSystem->FileExists(filename))
+			{
+				return false;
+			}
+
+			return true;
+		}
+
+		size_t Read(void* pOut, size_t size) KE_OVERRIDE
+		{
+			return static_cast<size_t>(g_FileSystem->Read(pOut, size, handle_));
+		}
+		
+		char* ReadLine(char* pOut, size_t size) KE_OVERRIDE
+		{
+			return g_FileSystem->ReadLine(pOut, size, handle_);
+		}
+
+		size_t Write(const void* pData, size_t size) KE_OVERRIDE
+		{
+			return static_cast<size_t>(g_FileSystem->Write(pData, size, handle_));
+		}
+		
+		bool Seek(int pos, int seek_type) KE_OVERRIDE
+		{
+			g_FileSystem->Seek(handle_, pos, static_cast<FileSystemSeek_t>(seek_type));
+			return !HasError();
+		}
+		
+		int Tell() KE_OVERRIDE
+		{
+			return g_FileSystem->Tell(handle_);
+		}
+		
+		bool HasError() KE_OVERRIDE
+		{
+			return !handle_ || !g_FileSystem->IsOk(handle_);
+		}
+		
+		int Flush() KE_OVERRIDE
+		{
+			g_FileSystem->Flush(handle_);
+			return 0;
+		}
+		
+		bool EndOfFile() KE_OVERRIDE 
+		{
+			return g_FileSystem->EndOfFile(handle_);
+		}
+	
+		void Close() KE_OVERRIDE 
+		{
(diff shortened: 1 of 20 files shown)


Arkshine:
Forgot to say that I've essentially tested almost everything.

amxmodx/CLang.cpp
```diff
@@ -12,6 +12,8 @@
 #include "CLang.h"
 #include "format.h"
 
+CVector<FILE *> FileList;
```
xPaw: Is this change supposed to be part of this PR?

plugins/include/file.inc
```diff
@@ -12,152 +12,655 @@
...
-native fread(file, &data, mode);
-native fread_blocks(file, data[], blocks, mode);
-native fread_raw(file, stream[], blocksize, blocks);
-native fwrite(file, data, mode);
-native fwrite_blocks(file, const data[], blocks, mode);
-native fwrite_raw(file, const stream[], blocksize, mode);
-
-//Returns 1 if the file is ended, 0 otherwise
+/**
+ * Reads a single binary data from a file.
```
xPaw: A single binary data?

xPaw:
Minor nit: In some documentation blocks, sentences are missing period at the end.

amxmodx/CLang.cpp
Arkshine: Yep, this file has been cleaned up, and there was no reason to have such global var defined here when it's unrelated to file natives.

plugins/include/file.inc
Arkshine: Well comment is stolen from SM, feel free to propose any changes.

Arkshine:
I've no idea when to put a dot. Tried to match Nextra's behavior but I probably failed. Feel free to point where a dot is needed...

Nextra:
Don't worry too much about the docs, I will touch it up if necessary.

amxmodx/CFileSystem.h
```diff
@@ -0,0 +1,256 @@
...
+			{
+				return nullptr;
+			}
+
+			return new ValveFile(handle);
+		}
+
+		static bool Delete(const char* filename, const char* pathID) 
+		{
+			if (!g_FileSystem->FileExists(filename))
```
Nextra: This could use Exists(), now that you added it.

amxmodx/CLibrarySys.cpp
```diff
@@ -0,0 +1,559 @@
...
+
+void CDirectory::NextEntry(cell* offset)
+{
+#if defined PLATFORM_WINDOWS
+
+	if (offset)
+	{
+		// Should be declared after loop so entry starts to '.' and not '..'
+		// But old code did that, so keep this for compatibility.
+		++*offset;
```
Nextra: I think keeping this function (and the class in general) free from `read_dir` quirks is the better way to go. `CDirectory` should cleanly provide access to a directory, so making legacy quirks part of it's interface is bad.

The offset stuff in general should probably not be part of `NextEntry` at all. It should only do what it says, so it can be used for a potential SM-style alternative to `read_dir` in the future. If we want to keep the seekdir/telldir optimization for POSIX we should move the offset stuff out into another function like `MoveEntry`, `SkipEntry` or similar.

amxmodx/file.cpp
```diff
@@ -7,484 +7,503 @@
...
-	
-	return (unlink(build_pathname("%s", sFile)) ? 0 : 1);
+	int length;
+	const char* file = get_amxstring(amx, params[1], 0, length);
+
+	if (*params / sizeof(cell) >= 2 && params[2] > 0)
+	{
+		const char* pathID = get_amxstring(amx, params[3], 1, length);
+
+		if (!strcmp(pathID, "ALL"))
```
Nextra: I don't really like adding this quirk. Thoughts?

xPaw: Would passing empty `valve_path_id` be better in this case?

amxmodx/file.cpp
```diff
@@ -7,484 +7,503 @@
...
+
+	if (!fp)
+	{
+		return -1;
+	}
+
+	switch (flag)
 	{
-		if (params[0] < 2 || params[2] == 0)
+		default:
```
Nextra: This shouldn't silently switch over to valid behavior, when it should actually have been an error path all along (I know we may not be able to fix that, but still).

amxmodx/file.cpp
```diff
@@ -7,484 +7,503 @@
...
+				{
+					break;
+				}
+
+				read += sizeof(value);
+			}
+			break;
 		}
-	case 4:
+		default:
```
Nextra: Same here. Invalid parameters were provided, this shouldn't work. If we ever need to add an extra mode here this will get hairy.

amxmodx/file.cpp
```diff
@@ -7,484 +7,503 @@
...
-			*addr = static_cast<cell>(a);
+			short value;
+			size_t res = fp->Read(&value, sizeof(value));
+			*data = static_cast<cell>(value);
 			return res;
 		}
-	case 4:	//int
-	default:
+		case BLOCK_INT:
+		default:
```
Nextra: ... I say as the old code does the same thing. Can we retroactively remove this...?

amxmodx/file.cpp
```diff
@@ -492,333 +511,545 @@ static cell AMX_NATIVE_CALL amx_fread(AMX *amx, cell *params)
...
+	{
 		return 0;
-	struct dirent *ep = readdir(dp);
-	
-	if (!ep)
+	}
+
+	size_t numParams = *params / sizeof(cell);
+
+	if (*params / sizeof(cell) >= 4 && params[5] > 0)
```
Nextra: numParams

amxmodx/file.cpp
```diff
@@ -7,484 +7,503 @@
...
-	cell *length = get_amxaddr(amx, params[5]);						// pointer to the outLen parameter
-	*length = set_amxstring(amx, params[3], fd.name, params[4]);	// set output and outLen parameters
-	_findclose(handle);
 
-	return tmp;
-#endif // __GNUC__
+	const char* entry = dir->GetEntryName();
+
+	size_t entrylen = set_amxstring_utf8(amx, params[3], entry, strlen(entry), params[4] + 1);
+	*get_amxaddr(amx, params[5]) = entrylen;
```
Nextra: Not really fond of combining `get_amxaddr` with the write access like that. Several other instances of this in the file.

amxmodx/file.cpp
Arkshine: Valve support empty string, but the whole purpose of having "ALL", is just to avoid people doing `..., true, ""` instead of `..., true, "ALL"`.

In SM, it uses a `public const String:NULL_STRING[1]` which acts as a C++ NULL ; basically, it uses a custom `LocalToStringNULL`,where it retrieves address of param, and if value is the same the public var index, it sets address to NULL. I wanted to do similar, but compiler doesn't like that;

Feel free to suggest.

Nextra:
:ship:

PR #218, opened 2015-03-09
Use upstream AMTL as a submodule

Arkshine:


dvander:
:ship:! you will probably have to fix up appveyor, see: https://github.com/alliedmodders/sourcemod/pull/299

PR #221, opened 2015-03-10
Remove scpack

xPaw:
@Arkshine wanted me to do it.

See https://github.com/alliedmodders/sourcemod/pull/297


Arkshine:
:fire:

Freeman-AM:
https://forums.alliedmods.net/showthread.php?t=259951
As noted here, this bug seems to be related to this PR.

PR #222, opened 2015-03-11
Make include files a little more spcomp2 compatible

xPaw:
This doesn't fix all the issue, just some simple ones.


Arkshine:
Tag changes should be fine here, it doesn't break compatibility nor doesn't change safety. Structure is the same (so any tags doesn't matter much) and it will allow the use of tagged structure or custom constants more easily.

PR #223, opened 2015-03-12
Add missing SVC Messages

9iky6:
add svc messages id to use hardcode


Arkshine:
Thanks for contributing, but you do some wrong things.

* Your title/commit message doesn't reflect the changes. Please take the time to write a proper message/title.
* While adding missing SVC messages are welcomed, client events are not, because they are mod-based and you can't write all, nor some won't have the same id depending the mod. Also there a native for that. Please remove them.
* You should avoid to insert random comments, especially when you mix english and some other language.
* You still don't name properly your branch name. Right now it's called "pathc-14", can't you name it like "missing-svc-messages" or something? Now it's too late, but next time, please try to name your branch properly.

Arkshine:
Looks ok, thanks.

PR #225, opened 2015-03-13
Move dlls/ to modules/

xPaw:
Shipped archives have modules inside `modules/` folder, and `dlls/` contains amxx core, but in the repository, `dlls/` folder contains modules, which is rather confusing.


Arkshine:
Looks good. Let's hope nothing is missing.

PR #228, opened 2015-03-24
Fix crash under linux in Hamsandwich module (bug 6323)

Arkshine:
Silly bug where a same pointer was saved 2 times in a list, resulting segfault on cleanup.
Bug appears only now because libc has been updated recently and the magic has disappeared.
Thanks to The Genius (a.k.a Nextra) for his help to narrow down the issue.

Related to #153


Nextra:
Ship it.

PR #229, opened 2015-03-26
Add %l format specifier and SetTranslationTarget() native (bug 6331)

Arkshine:
This adds a new `%l` format specifier which allows you to specify a client's index just once.
E.g. : 
`client_print(index, print_chat, "%L", index, "EGG"`
 is now 
`client_print(index, print_chat, "%l", "EGG"`

This new format will use the current index set internally with `SetDefLang`.

To complement this, a new native to define the current index has been added: `SetTranslationTarget`.
This is useful to be used before with natives which are not "players directed", like a bunch of `format[ex]` or custom function with `v[d]format`.

I wanted to fix base plugins, but I figured out that considering plugins need to be rewritten anyway, then changes will happen at this point.

### DIFF
--- a/amxmodx/amxmodx.cpp
+++ b/amxmodx/amxmodx.cpp
@@ -4431,6 +4431,14 @@ static cell AMX_NATIVE_CALL LookupLangKey(AMX *amx, cell *params)
 	return 1;
 };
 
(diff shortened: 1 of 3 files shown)


Nextra:
:shipit:

PR #230, opened 2015-03-26
Fix checking cvar's bounds

Ni3znajomy:
It looks like the current code that checks bounds for a cvar will fail in two cases. 

**1st case**
If we define min value that is higher than max value and max value is set to false,
```
create_cvar("cvar_name", "5", FCVAR_NONE, "desc", true, 1.0, false);
```
it will print -> *The maximum value can not be below the minimum value*

**2nd case**
If we define max value that is lower than min value and min value is set to false,
```
create_cvar("cvar_name", "5", FCVAR_NONE, "desc", false, 0.0, true, -1.0);
```
it will print -> *The minimum value can not be above the maximum value*

**Solution**
I think first we should check them if they are defined (both are set to true), then we can compare them. (Like I did in the this PR)


Arkshine:
Good catch. You're right, it makes sense to compare only if both bounds are set.

Thanks!

PR #231, opened 2015-03-28
Datapack update

Nextra:
- Adds type safety, brought over from SourceMod
- Removes `IsPackReadable` because it is terrible (comment in commit)
- Adds `CanReadX` functions ~~and natives~~ to determine readability, as a replacement to `IsPackReadable`
- Updates test plugin

### DIFF
--- a/amxmodx/CDataPack.cpp
+++ b/amxmodx/CDataPack.cpp
@@ -168,23 +168,32 @@ bool CDataPack::SetPosition(size_t pos) const
 	return true;
 }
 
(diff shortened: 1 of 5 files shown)


Nextra:
After some discussion with @dvander there will only be a `IsEnded` native as a replacement to `IsReadable`. The internal factoring out of `CanReadX` will stay as it makes for cleaner code and consistent error handling in the natives.

xPaw:
:+1: Changes seem fine to me.

Arkshine:
:ship:

PR #232, opened 2015-03-28
Fix crash in TFCX module when Infection weapon is used (bug 6333)

Arkshine:
The `PD_TIMER_OWNER` offset (which is by the way `m_hLastIDTarget`) is a player's offset and thus the extra linux should be `+5`.

Also, while I was on it, I did change the way how a `EHANDLE` offset is retrieved, using the fakemeta native approach, which is more appropriate and safe way.

Looks like TFCX module was using wrong linux extra offset since years.


Nextra:
:sheep:

PR #233, opened 2015-03-29
Update copyright year

Ni3znajomy:
Update copyright year.


xPaw:
I would just remove copyright years from makefiles, since most amxx files do not contain year in copyright notices.

Arkshine:
Thanks.

PR #234, opened 2015-04-01
Remove copyright years from Makefiles.

Ni3znajomy:
I did that what suggest xPaw in my previous PR to remove copyright years from makefiles because as he said most amxx's files don't contain them and these years won't need to be updated in the future anymore.


xPaw:
:+1:

Arkshine:
Makefile are now removed from project, closing this.

PR #235, opened 2015-04-12
Fix fgets native not setting always the plugin buffer

Arkshine:
Just a silly issue where a buffer from a plugin should be always set. It results that if you parse an empty line after a valid one, the plugin variable would contain previous value.

Related #217


Nextra:
Yup. :ship:

PR #236, opened 2015-04-12
Fix missing SQL_SetCharset native in SQLite

Arkshine:
Even though SQLite doesn't need this, It's motivated by the following:
* For consistency, considering SQLite and MySQL use the same API, it should be there
* For usability, when your plugin deals with both driver dynamically, you would want to have the same code for both (meaning not checking driver name to use native)


xPaw:
Documentation for SQL_SetCharset should have a note saying that this native does nothing in sqlite.

Arkshine:
Done.

Nextra:
lgtm

PR #237, opened 2015-04-19
Fix potential crash using ke::AString and a valid given null string (bug 6351)

Arkshine:
Just a misuse of `ke::AString`. 

Related to #112.


Nextra:
:ship:

dvander:
Why not patch upstream AString instead?

PR #239, opened 2015-05-04
Update include documentation #3 (mostly engine.inc)

Nextra:
Third time's a charm (#208 followup).

This one mostly touches engine.inc, which was awkward and took way too much time. A lot of old, weird API, some of it close to being impossible to document (`trace_forward` is a real gem, take a look - I tried my best).

Some of the comments already reference improvements and modifications that I will bring to the engine module soon-ish.

There are probably mistakes in here, and a ton of awkwardness. Feedback welcomed.

### DIFF
--- a/plugins/include/engine.inc
+++ b/plugins/include/engine.inc
@@ -24,243 +24,1273 @@
 	#pragma loadlib engine
 #endif
 
-native traceresult(type,any:...);
+/**
+ * Retrieves a result from the global trace handle.
+ *
+ * @note For a list of trace results available see the TR_* constants in
+ *       engine_const.inc.
+ * @note Usage examples:
+ *       value = traceresult(TR_AllSolid);
+ *       traceresult(TR_Fraction, floatvalue);
+ *       traceresult(TR_EndPos, vector);
+ *
+ * @param type      Result to retrieve
+ * @param ...       Depending on the result type a different number of
+ *                  additional parameters should be provided:
+ *                     int      - Returns the result integer value directly, no
+ *                                additional parameters required
+ *                     float    - Stores the result float value into the
+ *                                variable provided as the second parameter
+ *                     vector   - Copies the result vector to the Float:array[3]
+ *                                provided in the second parameter
+ *
+ * @return          Changes depending on the result type:
+ *                     int      - Returns the result integer value
+ *                     float    - Returns 1
+ *                     vector   - Returns 1
+ */
+native traceresult(type, any:...);
 
-/* Registers a client impulse to a function.  Function is passed the ID of the user. */
+/**
+ * Registers a function to be called on a client impulse.
+ *
+ * @note The function will be called in the following manner:
+ *
+ * public impulse_handler(client, impulse)
+ *
+ *   client     - Client index
+ *   impulse    - Impulse triggered by the client
+ *
+ * @note The callback should return PLUGIN_CONTINUE to ignore the impulse,
+ *       PLUGIN_HANDLED or higher to nullify it (CmdStart() is not blocked).
+ * @note When returning PLUGIN_HANDLED or higher from the callback, Engine will
+ *       still fire other impulse functions. This includes the client_impulse()
+ *       and client_cmdStart() forwards.
+ *
+ * @param impulse   Impulse to hook
+ * @param function  Name of callback function
+ *
+ * @noreturn
+ */
 native register_impulse(impulse, const function[]);
 
-/* Registers a touch action to a function by classnames.  Use * to specify any classname. */
+/**
+ * Registers a function to be called on a touch action between entities of
+ * specified classes.
+ *
+ * @note The function will be called in the following manner:
+ *
+ * public touch_handler(touched, toucher)
+ *
+ *   touched    - Index of entity being touched
+ *   toucher    - Index of entity touching
+ *
+ * @note The callback should return PLUGIN_CONTINUE to ignore the touch,
+ *       PLUGIN_HANDLED or higher to block it.
+ * @note When returning PLUGIN_HANDLED from the callback, Engine will still fire
+ *       other touch functions like the pfn_touch() forward before actually
+ *       blocking the touch. To immediately block return PLUGIN_HANDLED_MAIN
+ *       instead.
+ *
+ * @param Touched   Entity classname being touched, "*" or "" for any class
+ * @param Toucher   Entity classname touching, "*" or "" for any class
+ * @param function  Name of callback function
+ *
+ * @noreturn
+ */
 native register_touch(const Touched[], const Toucher[], const function[]);
 
-/* Registers a think action to a function by classname. */
+/**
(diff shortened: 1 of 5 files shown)


plugins/include/engine.inc
```diff
@@ -24,242 +24,1271 @@
...
-/* Drops an entity to the floor (work?) */
+/**
+ * Uses the DROP_TO_FLOOR() engine function on an entity, which attempts to put
+ * it down on the floor.
+ *
+ * @note This engine function traces 256 units straight downwards from the
+ *       entity origin. If the trace hits the floor, the origin is updated to
+ *       the end position of the trace, FL_ONGROUND is added to the flags and
+ *       EV_ENT_groundentity is updated. When the trace does not hit anything or
+ *       the entity would be stuck inside something, the process fails.
```
xPaw: What does `fails` mean exactly? Does it simply do nothing?

plugins/include/engine.inc
```diff
@@ -24,242 +24,1271 @@
...
 
-/* Use an entity with another entity. "entUsed" could be a hostage, "entUser" a player. */
+/**
+ * Forces an entity (such as a player) to use another entity (such as a button).
+ *
+ * @param entUsed   Index of entity being used
+ * @param entUser   Index of entity using
+ *
+ * @noreturn
+ * @error           If an invalid entity index is provided or, if the index is a
```
xPaw: This is worded poorly imo.

plugins/include/engine.inc
```diff
@@ -24,242 +24,1271 @@
...
-native Float:entity_range(ida,idb);
+/**
+ * Returns the distance between two entities.
+ *
+ * @param ida   Entity index 1
+ * @param idb   Entity index 2
+ *
+ * @return      Distance between the entities
+ * @error       If an invalid entity index is provided or, if the index is a
+ *              client index, the client is not connected, an error will be
```
xPaw: You can't find a distance between players?

plugins/include/engine.inc
```diff
@@ -24,242 +24,1271 @@
...
+ *
+ * @return          1 if entity is valid, 0 otherwise
+ */
 native is_valid_ent(iIndex);
 
-/* Proper origin setting, keeps updated with Half-Life engine. */
+/**
+ * Sets the origin of an entity.
+ *
+ * @note This native uses engine functions to set the origin, keeping it
```
xPaw: Why plural engine functions?

plugins/include/engine.inc
Nextra: Yes. It runs the trace and if the check fails (nothing hit or trace stuck inside something) it returns 0 and does nothing at all. I can reword that to make it more clear.

plugins/include/engine.inc
```diff
@@ -24,242 +24,1271 @@
...
+ * @error               If an invalid entity index is provided, an error will be
+ *                      thrown.
+ */
 native entity_set_origin(iIndex, const Float:fNewOrigin[3]);
 
-/* Sets the model of an Entity. */
+/**
+ * Sets the model of an entity.
+ *
+ * @note This native uses engine functions to set the model, keeping it
```
xPaw: Same here.

plugins/include/engine.inc
```diff
@@ -24,242 +24,1271 @@
...
+ *
+ * @noreturn
+ * @error           If an invalid entity index is provided, an error will be
+ *                  thrown.
+ */
 native entity_set_model(iIndex, const szModel[]);
 
-/* Remove an entity from the world. */
+/**
+ * Removes an entity from the world.
```
xPaw: IIRC calling `remove_entity` on a client index will crash the server, so that might be worthy of a note.

plugins/include/engine.inc
Nextra: Do you mean the error annotation? I'll think of something. Honestly I don't even know if I should be using the full-sentence-style for the errors. Thought about changing it all to a simpler enumeration/list style...

xPaw: I meant it's confusing which index you're talking about. You should specify if it's entUsed or entUser.

plugins/include/engine.inc
```diff
@@ -24,242 +24,1271 @@
...
-//from jghg2
-/* As above, but returns number of ents stored in entlist. Use to find a specific type of entity classname (specify in _lookforclassname) around a
- * certain entity specified in aroundent. All matching ents are stored in entlist. Specify max amount of entities to find in maxents.
- * If aroundent is 0 its origin is not used, but origin in 6th parameter. Ie, do not specify 6th parameter (origin) if you specified an entity
- * in aroundent.
+/**
+ * @endsection
+ */
+
+/**
```
xPaw: Move this function up next to other find_ functions?

plugins/include/engine.inc
Nextra: It is for both, I left the text generic if it applies to all entity parameters. I can probably change that?

plugins/include/engine.inc
Nextra: You can, it is not specific either way. When it says "entity" it means literally any entity. If functions are specific to clients it will explicitly say so.

plugins/include/engine.inc
xPaw: > if the index is a client index

Why would it error if a client is trying to use something?

plugins/include/engine.inc
xPaw: Ooooooooh I re-read it, and it makes sense now.

plugins/include/engine.inc
Nextra: It uses both `SET_SIZE` and `SET_MODEL`. Either way this is just specifics, and the note should only make the user aware that there is a difference between using `EV_VEC_origin` and `entity_set_origin()`.

plugins/include/engine.inc
Nextra: Copy-paste from `entity_set_origin()`. I'll fix it, but this is not the main point of the note.

xPaw: Just nit picking :smile_cat:

plugins/include/engine.inc
Nextra: Will do.

plugins/include/engine.inc
Nextra: All of this is already wordy enough, so I kind of prefer keeping the implicit way and only making the distinction when a function is actually client only. I'll see if I can come up with something.

plugins/include/engine.inc
Nextra: I will keep that in mind for the upcoming engine update and make it a real error path.

Nextra:
This can be merged unless someone has further feedback.

xPaw:
:+1:

Arkshine:
Let me read one time in few minutes.

Arkshine:
Looks good. :ship:

PR #240, opened 2015-05-05
Improve handling of KeyValueData

Nextra:
Despite Hamsandwich documentation pointing to FM natives for handling KVD, FM was actually incompatible to the raw pointers Ham works with. This is due to the `KVD_Wrapper` helper structure in FM, the basic compatibility fix is making `&kvdw == &kvdw.kvd`, so the wrapper can be pushed into Ham without issue.

`create_kvd()` and `free_kvd()` allow creation of new KVD structures that can be used with Ham. These can and should also be used in FM to replace the global `0` KVD that has multiple problems and can unfortunately not be fixed up properly.

Documentation will be done properly when I touch fakemeta.inc in the next doc update.


Arkshine:
:sheep:

PR #241, opened 2015-05-05
Update cmdmenu.sma

luxxxoor:
- Changed format with formatex where needed.
- Removed many hardcoded contents in plugin.
- Added semicolons.

### DIFF
--- a/plugins/cmdmenu.sma
+++ b/plugins/cmdmenu.sma
@@ -26,81 +26,81 @@ new g_cmdMenuName[MAX_CMDS_LAYERS][] =
 	"CMD_MENU",
 	"CONF_MENU",
 	"SPE_MENU"
-}
+};
 
 new g_cmdMenuCmd[MAX_CMDS_LAYERS][] =
 {
 	"amx_cmdmenu",
 	"amx_cfgmenu",
 	"amx_speechmenu"
-}
+};
 
 new g_cmdMenuCfg[MAX_CMDS_LAYERS][] =
 {
 	"cmds.ini", 
 	"configs.ini",
 	"speech.ini"
-}
+};
 
 new g_cmdMenuHelp[MAX_CMDS_LAYERS][] =
 {
 	"- displays commands menu",
 	"- displays configs menu",
 	"- displays speech menu"
-}
+};
 
 /* End of Commands Menu */
 
 #define MAX_CMDS    64
 #define MAX_CVARS   48
 
-new g_cmdName[MAX_CMDS*MAX_CMDS_LAYERS][32]
-new g_cmdCmd[MAX_CMDS*MAX_CMDS_LAYERS][64]
-new g_cmdMisc[MAX_CMDS*MAX_CMDS_LAYERS][2]
-new g_cmdNum[MAX_CMDS_LAYERS]
+new g_cmdName[MAX_CMDS*MAX_CMDS_LAYERS][32];
+new g_cmdCmd[MAX_CMDS*MAX_CMDS_LAYERS][64];
+new g_cmdMisc[MAX_CMDS*MAX_CMDS_LAYERS][2];
+new g_cmdNum[MAX_CMDS_LAYERS];
 
-new g_cvarNames[MAX_CVARS][32]
-new g_cvarMisc[MAX_CVARS][3]
-new g_cvarCmd[MAX_CVARS*5][32]
-new g_cvarCmdNum
-new g_cvarNum
+new g_cvarNames[MAX_CVARS][32];
+new g_cvarMisc[MAX_CVARS][3];
+new g_cvarCmd[MAX_CVARS*5][32];
+new g_cvarCmdNum;
+new g_cvarNum;
 
-new g_menuPosition[MAX_PLAYERS + 1]
-new g_menuSelect[MAX_PLAYERS + 1][MAX_CMDS]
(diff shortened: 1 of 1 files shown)


plugins/cmdmenu.sma
```diff
@@ -140,26 +146,37 @@ public plugin_precache( )
 			copy( szSound, charsmax(szSound), field2 );
 			copy( sndExt, charsmax(sndExt), ".wav" );
 		}
-		else if ( fieldNums == 3 && field1[0] == 'm' && ( field2[0] == 'p' || field2[0] == 'l' ) )	// .mp3 (mp3 play | mp3 loop)
-		{
-			copy( szSound, charsmax(szSound), field3 );
-			copy( sndExt, charsmax(sndExt), ".mp3" );
+		else 
+		{
```
Arkshine: Why did you "expand" this condition? It was fine before. No reason to change it.

plugins/cmdmenu.sma
```diff
@@ -140,26 +146,37 @@ public plugin_precache( )
...
 
 		replace_all( szSound, charsmax(szSound), "\'", "" );								// Strips all ugly (and sometimes useless) \'
 
 		if ( szSound[0] == '/' )
+		{
 				replace( szSound, charsmax(szSound), "/", "" );						// Strip leading slash
+		}
 
-		if ( sndExt[1] == 'm' || ( ! equali( szSound, "vox", 3 ) && ! equali( szSound, "fvox", 4 ) && ! equali( szSound, "barney", 6 ) && ! equali( szSound, "hgrunt", 6 ) ) )
+		if ( sndExt[1] == 'm' || ( ! equali( szSound, "vox", strlen("vox") ) && ! equali( szSound, "fvox", strlen("fvox") ) && ! equali( szSound, "barney", strlen("barney") ) && ! equali( szSound, "hgrunt", strlen("hgrunt") ) ) )
```
Arkshine: I don't see much improvement here: one native call + duplicating string. Using directly the length was fine in this context. If it bothers you, then you could simple declare `new const voxIdent[] = "vox";` then using `charsmax(voxIdent)`.

plugins/cmdmenu.sma
```diff
@@ -186,128 +203,171 @@ public actionCmdMenu(id, key)
...
+			new option = g_menuSelect[id][g_menuPosition[id] * 8 + key];
+			new flags = g_cmdMisc[option][1];
 			
 			if (flags & 1)
-				server_cmd("%s", g_cmdCmd[option])
-			else if (flags & 2)
-				client_cmd(id, "%s", g_cmdCmd[option])
-			else if (flags & 4)
-				client_cmd(0, "%s", g_cmdCmd[option])
+			{
```
Arkshine: Same here, I'm not sure to understand why you "expand". No reason to do that.

plugins/cmdmenu.sma
```diff
@@ -351,97 +417,122 @@ public actionCvarMenu(id, key)
...
-	return PLUGIN_HANDLED
+	return PLUGIN_HANDLED;
 }
 
 loadCvarSettings(szFilename[])
 {
 	if (!file_exists(szFilename)) 
-		return 0
+	{
+		return PLUGIN_CONTINUE;
```
Arkshine: It makes no sense to return `PLUGIN_CONTINUE`, it's a private function where the return value is not used (for now), but still returns 0 or 1 because it makes sense to do that. If something should be changed, I would return a bool.

plugins/cmdmenu.sma
```diff
@@ -468,14 +563,14 @@ loadCvarSettings(szFilename[])
...
-		g_cvarMisc[g_cvarNum][2] = g_cvarCmdNum
-		g_cvarMisc[g_cvarNum][0] = read_flags(szValues[inum])
-		g_cvarNum++
+		g_cvarMisc[g_cvarNum][2] = g_cvarCmdNum;
+		g_cvarMisc[g_cvarNum][0] = read_flags(szValues[inum]);
+		g_cvarNum++;
 	}
 
-	return 1
+	return PLUGIN_HANDLED;
```
Arkshine: Same as above.

Arkshine:
Why did you close the PR?

luxxxoor:
I pressed the wrong button from phone, i think the "expand" was needed to understand the code more easier

Arkshine:
I think honestly it does the reverse. You see all conditions on the same level, which is more readable, and unless it's really needed (like logic issue), I would not change them.

luxxxoor:
For me it is a lot easier to understand the code if it is explained, but if you think the expand will do the reverse i will return to the standard. And i want to know if all the expands are bad (not easy readable) or just some of them ?

Arkshine:
In this context, there is no meaning to do that. I would prefer you revert it. Also this is not of matter of bad or something, when conditions are simple and can be aligned to have a better overview, this is more readable. If conditions are complex or long, you can expand so it's easier to understand ; just a matter of context.

Arkshine:
Why did you revert ALL the changes about the added parenthesis? Did I mention you to revert that? I've SPECIFICALLY pointed above the lines where EXACTLY I would expect some changes... If I did not commented on the others lines, it means It was okay, why are you doing unnecessary changes now?

luxxxoor:
Now is ok ? If it's ok i will squash the commits.

Arkshine:
Yes.

Arkshine:
Thanks!

PR #242, opened 2015-05-06
Add NULL_VECTOR and expose null functions to AMXX API

Nextra:
`NULL_VECTOR` will be required for future engine updates, and is generally nice to have. For module use `get_amxstring_null` and `get_amxvector_null` obviously have to be exposed to the API.


Arkshine:
:ship:

PR #244, opened 2015-05-15
Fix EngFunc_ChangeLevel

Arkshine:
This fixes `EngFunc_ChangeLevel` which never worked from the start.

This function takes 2 arguments, and the second is used on single player game only. To act like a `changelevel` command, second param needs to be `NULL`.  But the original code was passing always a non-null string, even though empty, and since engine disallows this if server is multiplayer, function could not work.

Even though the function is redundant and because we can't delete this, we might want to fix this.
Function is now supporting `NULL_STRING` identifier on the second param. Documentation will be updated in another PR.


Nextra:
Might aswell. :ship:

PR #245, opened 2015-05-15
Plmenu: separation of cstrike version of base version

Freeman-AM:
This split is a requirement for further updates.

This update include:
- fakemeta module as a requirement rather than an optional module, because whatever we do in the future, we will need to work with offsets.
- move cstrike module as a requirement rather an option for cstrike/plmenu
- remove counter-strike related code in base/plmenu
- tell builder and installer about new files

### DIFF
--- a/installer/amxmodx-installer.nsi
+++ b/installer/amxmodx-installer.nsi
@@ -1,4 +1,4 @@
-;(C)2004 AMX Mod X Development Team
+;(C)2004-2015 AMX Mod X Development Team
 ; Licensed under the GNU General Public License
 ; Originally written by -=HaXoMaTiC=-
 !define PRODUCT_NAME "AMX Mod X Installer"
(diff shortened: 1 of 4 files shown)


xPaw:
I would actually propose doing the opposite, merging mod-specific plugins into one, so they are easier to maintain.

Arkshine:
If doing like current code, this is easier to split code for CS only. But if changed code is somehow abstracted so we have well defined block of specific mod-related code, I would not mind yes.

Freeman-AM:
My plans have changed, it's a lot more in Work in Progress state now.
I'm working on something really different for base plugin.
It will be compatible with at least half-life half-life oppossing force and death match classic team system if enabled.

Freeman-AM:
Working on this update paused until 15th June

Freeman-AM:
Updated half-life version.
Comments would be appreciated.

ClaudiuHKS:
Of course it would be better so all PlMenu plug-ins are merged together.
Easier to update, easier to maintain.

Freeman-AM:
I'm closing this one. I don't work anymore on goldsource games.

PR #246, opened 2015-05-18
Correcting msvs12 Build and Platform tool versions

ShootingKing-AM:
For MSVS12,
BuildToolsVersion is 4.0 and
PlatfromToolSetVersion is v110.

Ref: https://msdn.microsoft.com/en-us/library/bb383796.aspx

### DIFF
--- a/compiler/amxxpc/amxxpc.sln
+++ b/compiler/amxxpc/amxxpc.sln
@@ -1,6 +1,6 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 11.00
-# Visual Studio 2010
+# Visual Studio 2012
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "amxxpc", "amxxpc.vcxproj", "{39412290-D01C-472F-A439-AB5592A04C08}"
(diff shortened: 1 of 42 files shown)


Arkshine:
AMXX is built against MSVC 12 (2013), toolset is well v120_xp.
There is nothing to change.

ShootingKing-AM:
So you mean msvc12 is version no not year ? :O

Arkshine:
Yeah, version and year is usually not the same : http://www.wikiwand.com/en/Microsoft_Visual_Studio#/History

ShootingKing-AM:
I thought 12 is year. Also, in .sln's its written '# Visual Studio 2010' isn't it wrong ?

Arkshine:
It's just a comment. It because project files were using MSVC 10 before being upgraded to 12. Probably some bit left, but doesn't matter much.

ShootingKing-AM:
Its not just a comment MSVS decides the year of the solution based on it. You can observe this in solution icon as well.

Arkshine:
The # should be just the product description. If version is retrieved from this file, it's likely from the first line, which shows the internal version.

PR #247, opened 2015-05-18
Correcting years in MSVS project solutions.

ShootingKing-AM:
Correcting years in MSVS project solutions for 2010 to 2013.

### DIFF
--- a/compiler/amxxpc/amxxpc.sln
+++ b/compiler/amxxpc/amxxpc.sln
@@ -1,6 +1,6 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 11.00
-# Visual Studio 2010
+# Visual Studio 2013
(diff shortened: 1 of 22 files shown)


Arkshine:
File format should be `12.00`, if you change year to `2013`. I'm not sure actually that's something you are supposed to change manually. Did you do some test?

Is there actually an issue right now with the project files as it is?

ShootingKing-AM:
Yes. Corrected. Do test yourself Arkshine, its better if more people test it :)

We have found out inconsistent solutions. Solutions files (.sln) ask for Msvs - 2012(showing themselves as 2010 solutions) and Project Files (.vcproj) ask for Msvs - 2013 toolsets. Should we fix it now or should we wait until it causes some issue ?

Arkshine:
Like I said, I have no issue with the current project files. If you provide a PR, you MUST test it.
I don't have VS 2012, and you need anyway VS 2013. You are not going to have issues. Also this is not the official way to compile, even though I'm trying to keep files up to date, so such changes really doesn't matter, but since you did the changes, let's go with that, should be okay.

PR #248, opened 2015-05-19
Update compiler - 3

Arkshine:
Necessary update to be able to use `NULL_STRING` and `NULL_VECTOR` as default value for some natives arguments. Same patch as alliedmodders/sourcemod@d8d13f1.

This includes an extra fix as another commit, since it shares the same code. ([bug 4977](https://bugs.alliedmods.net/show_bug.cgi?id=4977))

### DIFF
--- a/compiler/libpc300/sc1.c
+++ b/compiler/libpc300/sc1.c
@@ -2190,15 +2190,19 @@ static int base;
  *
  *  Global references: litidx (altered)
  */
-static void initials(int ident,int tag,cell *size,int dim[],int numdim,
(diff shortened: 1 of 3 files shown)


dvander:
The initials code is all insane so I'm just going to assume you copied it faithfully :scream:

dvander:
Rest looks good. Sorry for the late review. :ship:

Arkshine:
Aha, yes, and I've tested it anyway. Don't worry, you will be the first to know if your :baby: goes wild.

PR #249, opened 2015-06-01
Fix a clang-3.6 compilation issue

Arkshine:
Reported on the forum: https://forums.alliedmods.net/showthread.php?t=263726


Nextra:
:ship:

PR #250, opened 2015-06-11
Engine Update

Nextra:
Working on the engine documentation I found a bunch of stuff in the module that could be improved, ranging from "slightly odd" to "broken". While completing the docs I created [a list of all of those things](https://gist.github.com/Nextra/c5369d7579664bfa5b1d) with @Arkshine. This PR implements those changes. Got sidetracked a bit while #248 got resolved.

The new lightstyle stuff still needs testing, and the other changes could probably use some scrutiny, too.


Arkshine:
:sheep:

PR #251, opened 2015-06-15
Replace console_print by engclient_print to support connecting players (bug 6372)

WPMGPRoSToTeMa:


Arkshine:
Should be okay. Thanks. :ship:

PR #253, opened 2015-06-29
Fix set_amxstring_utf8 insconsistency

Arkshine:
Originally code was copied from SM, and last argument is max length including null terminator. In AMXX, plugins returns the length without null terminator, so it makes sense that max length doesn't include it. This is also more consistent wiht `set_amxstring`.


Nextra:
:sheep:

PR #254, opened 2015-06-29
Implement a game config parser and expose functionnalities to the module/plugin API

Arkshine:
First step to improve maintainability. A game config parser stolen form SM. Any hardcoded datas will be eventually moved to some gamedata files.

This is essentially the same as SM with some changes:
* Removed useless/unused code for us (sendprop, crc, class, etc..)
* Added a new node "Classes" , mainly to associate a list of offsets to a class.

The config manager is exposed to module API through `MF_GetConfigManager`.

The parser functions is exposed as well to plugin API, so author can easily share game datas:
* `LoadGameConfigFile`
* `GameConfGetOffse`
* `GameConfGetClassOffset`
* `GameConfGetKeyValue`
* `GameConfGetAddress`
* `CloseGameConfigFile`

PR #255, opened 2015-07-08
Update cstrike module - 1

Arkshine:
This is based on #254 .

This patch is a logic step after implementing a game config system to avoid any hardcoding of game datas. While I was on it, I've fixed/improved/cleanup the obvious things.

Essentially:
* All game datas are now external and since we can dump all  the classes headers of the binary, this includes as well offsets of most of classes associated to an entity.
* An approach on-demand of hooks used by some natives/forwards. In others words, if these natives/forwards are not used, module does nothing.
* Some hostages natives support now CZ
* Some natives gain an extra param
* Worth to be noted: `cs_set/reset_user_model`, `cs_set_user_team` won't trigger anymore an overflow resulting a crash when used numerous times in a row. This is an engine bug, and the new workaround in this patch is basically postponing the update of models to the next frame.
* Some minor optimizations

That's all for now.


xPaw:
Just a nit, replace spaces with underscores in links.
e.g: `Gamedata_Updating_(AMX Mod X)` :arrow_right: `Gamedata_Updating_(AMX_Mod_X)`

Arkshine:
Making suddenly such message be hookable by plugins can lead to some unexpected behaviors. So probably not a good idea unless we add a new param or something, but doing that for the related natives seem like silly.

plugins/include/cstrike.inc
```diff
@@ -127,13 +127,14 @@ enum
 #define CSI_SG552               CSW_SG552
 #define CSI_AK47                CSW_AK47
 #define CSI_P90                 CSW_P90
-#define CSI_SHIELDGUN           CSW_SHIELDGUN
+#define CSI_SHIELDGUN           CSW_SHIELDGUN   // The real CS value, use CSI_SHELD instead.
```
Freeman-AM: Typo

Arkshine:
Tested all the related natives and it worked fine. Unless I did not test properly or missed something, this should be fine.

Nextra:
Good that you tested again, should cover the big issues. This is really hard to go through in detail in a reasonable time frame. We can probably :ship:

PR #256, opened 2015-07-10
Plugins: Close Array/Trie handles on map end

DmitriyS:
from `cellarray.inc` and so:
"""
 * @note Plugins are responsible for freeing all Array handles they acquire,
 *       including those from ArrayClone. Failing to free handles will result
 *       in the plugin and AMXX leaking memory.
"""

### DIFF
--- a/plugins/admincmd.sma
+++ b/plugins/admincmd.sma
@@ -1332,3 +1332,9 @@ public cmdLast(id, level, cid)
 	
 	return PLUGIN_HANDLED;
 }
+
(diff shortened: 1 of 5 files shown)


Arkshine:
Usually handles are always freed by AMXX, but it's true you could wasted memory in 1.8.2 depending what you do. In 1.8.3-dev, Trie/Array libraries are completely different, this won't happen. That's said, this is a good practice to close any kind of handles once you don't use it.

PR #258, opened 2015-07-11
Fix mistake

Ni3znajomy:
It was giving me an error during compiling.
```
/home/karol/Pobrane/amxx/amxmodx/amxmodx/CGameConfigs.cpp:904:76: error: operation on ‘returnValue’ may be undefined [-Werror=sequence-point]
```
I think it's just a mistake.


Arkshine:
Yeah, just a typo.

How did you compile just by curiosity? VS 2013 or clang 3.4 ignore this.

Ni3znajomy:
I'm using GCC 5.1

PR #260, opened 2015-07-13
More AMTL conversion - :fire: CString and CVector

Arkshine:


dvander:
Let it burn, let it buuuuuurn :fire: :musical_note: :musical_note:

xPaw:
:fire_engine:

amxmodx/CLang.cpp
```diff
@@ -24,24 +23,29 @@ CVector<FILE *> FileList;
 #define INSERT_NEWLINE		4
 
 template<>
-int Compare<String>(const String &k1, const String &k2)
+int Compare<ke::AString>(const ke::AString &k1, const ke::AString &k2)
 {
-	return k1.compare(k2.c_str());
+	return k1.compare(k2.chars());
```
Nextra: There's an overload for ke::AString so `k1.compare(k2)` does the same thing. There are other places where  this should be used for consistency.

amxmodx/format.cpp
```diff
@@ -114,10 +114,12 @@ const char *translate(AMX *amx, cell amxaddr, const char *key)
...
 		if (debug)
 		{
-			if (status == ERR_BADLANG && (BadLang_Table.AltFindOrInsert(pLangName).last + 120.0f < gpGlobals->time))
+			ke::AString lang(pLangName);
+
+			if (status == ERR_BADLANG && (BadLang_Table.AltFindOrInsert(ke::Move(lang)).last + 120.0f < gpGlobals->time))
 			{
 				AMXXLOG_Error("[AMXX] Language \"%s\" not found", pLangName);
-				BadLang_Table.AltFindOrInsert(pLangName).last = gpGlobals->time;
+				BadLang_Table.AltFindOrInsert(ke::Move(lang)).last = gpGlobals->time;
```
Nextra: If lang is moved twice this won't work.

Nextra:
:ship:

amxmodx/messages.h
```diff
@@ -50,52 +48,45 @@ class RegisteredMessage
...
 
-						m_Forwards.erase(iter);
+						m_Forwards.remove(forward);
 
 						return true;
 					}
 					else
 					{
 						// -1 could be in here more than once
-						m_Forwards.erase(iter);
```
Arkshine: Worth to note (for changelog) that there was a bug here. It removes an iteration without setting `iter` to the new value and with the wrong loop condition where vector should be checked if empty instead.

If you pass there, this results in an infinite loop. It can happen when you call `unregister_message` native in its own hook.

amxmodx/CLang.cpp
```diff
@@ -332,167 +480,60 @@ int CLangMngr::MergeDefinitionFile(const char *file)
...
-						tmpEntry.key = iKey;
-						tmpEntry.definition = new String;
-						multiline = true;
-					} else {
-						//user typed a line with no directives
-						AMXXLOG_Log("Invalid multi-lingual line (file \"%s\" line %d)", file, line);
-					}
-				}
-			} else {
-				if (buf[0] == ':')
```
jacobrshields, line 436: FYI, removing multiline support in dictionary files is an undocumented breaking change.

PR #262, opened 2015-07-13
Remove MakeFile

Arkshine:
Time to :fire: this old build system as well. There is no reason to keep such [outdated, for some] files now we have AMBuild. And yes, I'm lazy, I don't want to maintain them. :japanese_goblin:


Nextra:
good riddance :package:

PR #263, opened 2015-07-14
Fix CS_OnBuyAttempt

Sylvert0:
If i dont have CS_OnBuy on my plugin, CS_OnBuyAttempt doesn't work


Arkshine:
Oops. Thanks.  

Fixes #68.

PR #264, opened 2015-07-17
Fix player not being internally disconnected in some situation and add client_disconnected forward

Arkshine:
Issue reported by @WPMGPRoSToTeMa here #252.

This fixes the situation of a player is marked as initialized by AMXX but has not yet spawned (put in server), if he disconnects/crashes/timeout at this point, `pfnClientDisconnect` won't be triggered by the engine. Result: AMXX will think that this player is still initialized and as consequence code relying on it could provide some unexpected result, also `client_disconnect` won't be called. 

While it's necessary that this change be reflected on `client_disconnect`, instead for backward compatibility reason a new forward `client_disconnected` has been introduced to be used as replacement. `client_disconnect` is marked as deprecated.

Technical note: Core hooks `SV_DropClient`, reHLDS could be supported later, once release is considerate as stable, and/or there is a more friendly approach of the API.


amxmodx/meta_api.cpp
```diff
@@ -811,18 +818,49 @@ BOOL C_ClientConnect_Post(edict_t *pEntity, const char *pszName, const char *psz
...
+
+		if (pPlayer->initialized)
+		{
+			executeForwards(FF_ClientDrop, pPlayer->index, prepareCharArray(buffer, sizeof(buffer), true), sizeof(buffer) - 1);
+		}
+	}
+
+	DETOUR_STATIC_CALL(SV_DropClient)(cl, crash, "%s", buffer);
+
+	pPlayer->Disconnect();
```
xPaw: This should be inside `if (cl->edict)`

Arkshine:
So, after some discussions, it would be more appropriate/convenient/usable to have a replacement instead. There we go, a new forward `client_disconnected` which is merged with `client_drop`.
`client_disconnect` is still there for backward compatibility but is now marked as deprecated.
Any thoughts on this?

RauliTop:
Has the change take into account FM_ClientDisconnect forward from fakemeta?

Also, why client_disconnect is called when player disconnects or map change (1 called for each player connected) and FM_ClientDisconnect only called when player disconnects?

Arkshine:
I'm not sure why you're talking about `FM_ClientDisconnect`, fakemeta is unrelated here or I don't understand.

RauliTop:
You said: "...some unexpected result, also client_disconnect won't be called."
I'm asking if FM_ClientDisconnect it's called or not too.

Also, I asked other thing, but any problem if no answer, it was only for curiosity.

Arkshine:
If you follow #252 this is a situation where pfnClientDisconnect is not called. Essentially, if a client is connecting but has not yet spawned (meaning put in server at least) and if he crashes/timeout at this point, client_disconnect won't be called.

xPaw:
Looks fine to me now.

amxmodx/meta_api.cpp
```diff
@@ -1488,6 +1545,20 @@ C_DLLEXPORT	int	Meta_Attach(PLUG_LOADTIME now, META_FUNCTIONS *pFunctionTable, m
...
+	void *address = nullptr;
+
+	if (CommonConfig && CommonConfig->GetMemSig("SV_DropClient", &address))
+	{
+		DropClientDetour = DETOUR_CREATE_STATIC_FIXED(SV_DropClient, address);
+		DropClientDetour->EnableDetour();
+	}
+	else
+	{
+		AMXXLOG_Log("client_drop forward has been disabled - check your gamedata files.");
```
Nextra: fwd name

plugins/include/amxmodx.inc
```diff
@@ -156,16 +156,26 @@ forward client_connect(id);
 forward client_authorized(id);
 
 /**
- * Called when a client is disconnecting from the server.
+ * @deprecated This function does not catch all cases.
```
Nextra: deprecation does not make the documentation invalid, we haven't removed it in all other cases either

amxmodx/CGameConfigs.cpp
```diff
@@ -886,6 +887,17 @@ CGameConfigManager::~CGameConfigManager()
 {
 }
 
+void CGameConfigManager::OnAmxxStartup()
+{
+	char error[256]; *error = '\0';
```
Nextra: `char error[256] = "";` ?

amxmodx/CGameConfigs.cpp
```diff
@@ -886,6 +887,17 @@ CGameConfigManager::~CGameConfigManager()
...
+
+	if (!LoadGameConfigFile("common.games", &CommonConfig, error, sizeof(error)) && *error != '\0')
```
Nextra: Should this really not log *at all* if there is no error message (not that it could happen, but still)?

Nextra:
:ship:

afwn90cj93201nixr2e1re:
Im really don't understand why we should replace META ClientDisconnect which is based inside core with SV_DropClient, how it's related to each other? Why we can't use what we wanna?
I understand that client_disconnect not gonna be called if we make something like connect ip then instantly press escape for disconnecting.

As you can see neither client_disconnect, neither client_disconnected called sometimes... Just check, i highlighted.

**14** `Client sent 'drop'` messages.
**11** `client_disconnected` calls.

```pawn
public client_disconnected(id){
	server_print("Player %d client_disconnected", id);
}
public client_disconnect(id){
	server_print("Player %d disconnected", id);
}
```

```js
let client_disconnected = 0;
let Reason = 0;
let Dropped = 0;
L 07/15/2020 - 03:26:54: -------- Mapchange to de_chateau --------
Executing ReGameDLL Configuration File
Navigation map loaded.
L 07/15/2020 - 03:26:55: [admin.amxx] [AMXX] Загружено 5 администраторов из файла
[AMXX] Загружено 4 администраторов из файла2
Executing AMX Mod X Configuration File
couldn't exec listip.cfg
couldn't exec banned.cfg
Executing server.cfg
restart
plugin_end
L 07/15/2020 - 03:46:19: -------- Mapchange to de_chateau --------
Executing ReGameDLL Configuration File
Navigation map loaded.
L 07/15/2020 - 03:46:21: [admin.amxx] [AMXX] Загружено 5 администраторов из файла
[AMXX] Загружено 4 администраторов из файла2
Executing AMX Mod X Configuration File
couldn't exec listip.cfg
couldn't exec banned.cfg
Executing server.cfg

ip:27006:reconnect
Player 1 client_disconnected
Dropped test from server
Reason:  Client sent 'drop'

ip:27006:reconnect
Player 1 client_disconnected
Dropped test from server
Reason:  Client sent 'drop'

ip:27006:reconnect
Player 1 client_disconnected
Player 1 disconnected
Dropped test from server
Reason:  Client sent 'drop'

ip:27006:reconnect
Player 1 client_disconnected
Dropped test from server
Reason:  Client sent 'drop'

ip:27006:reconnect
Dropped test from server
Reason:  Client sent 'drop'

ip:27006:reconnect
Player 1 client_disconnected
Dropped test from server
Reason:  Client sent 'drop'

ip:27006:reconnect
Player 1 client_disconnected
Dropped test from server
Reason:  Client sent 'drop'

ip:27006:reconnect
Player 1 client_disconnected
Dropped test from server
Reason:  Client sent 'drop'

ip:27006:reconnect
Dropped test from server
Reason:  Client sent 'drop'

ip:27006:reconnect
Player 1 client_disconnected
Dropped test from server
Reason:  Client sent 'drop'

ip:27006:reconnect
Player 1 client_disconnected
Dropped test from server
Reason:  Client sent 'drop'

ip:27006:reconnect
Player 1 client_disconnected
Dropped test from server
Reason:  Client sent 'drop'

ip:27006:reconnect
Player 1 client_disconnected
Dropped test from server
Reason:  Client sent 'drop'

ip:27006:reconnect
Dropped test from server
Reason:  Client sent 'drop'
```

afwn90cj93201nixr2e1re:
@Arkshine @WPMGPRoSToTeMa check it out, seems like there's some ub.

afwn90cj93201nixr2e1re:
https://github.com/dreamstalker/rehlds/blob/65c6ce593b5eabf13e92b03352e4b429d0d797b0/rehlds/engine/host.cpp#L461

As you can see there s no extra conditions, maybe we should hook steam api disconnected instead this one for disconnected? And leave disconnect as expected? Coz disconnect used only for authed players for example in database saving and other stuff like that, we everytime checking is player was fully connected till putinserver and etc, i mean i know that there's only one case which can bypass this (seem's like client crash), but whatever? We exactly should use steam api for that. @Arkshine

afwn90cj93201nixr2e1re:
Possible some of rehlds meta modules also hook's dropclient and just breaking chain calls.

afwn90cj93201nixr2e1re:
There's some general error's with that.
изображение

Sometimes client_disconnected not called at all, but sometimes client_connect also not called... tf? If client sent drop printed then entity is exactly valid, idk why we should make any internal check's to pPlayer->initialized inside amxmodx. Seems like that's the main reason of current unexpected stuff. Especially it's mostly possible on mapchanges/map ends.
изображение

afwn90cj93201nixr2e1re:
i really don't understand why disconnected is under some condition checks
изображение


as you can see edict exists everytime when we try to connect/disconnect also if we doing it too fast.
изображение
But as you can see there internal check's in amxmodx causing more unexpected shit. Maybe we should remove them?

afwn90cj93201nixr2e1re:
i really don't understand why disconnected is under some condition checks


изображение


as you can see edict exists everytime when we try to connect/disconnect also if we doing it too fast.
изображение
But as you can see there internal check's in amxmodx causing more unexpected shit. Maybe we should remove them?

PR #265, opened 2015-07-18
Retrieve log type in MetaAttach to allow log to be written earlier

Arkshine:
Small issue. Before log type was retrieved only when the map world spawns. By default value is set to 0 and is not reset afterward. To write in the log, you need a value > 0. This means, at first server start, any attempt to write in the log before this time was ignored.
Considering we do now a lot of stuffs at startup, this needs to be fixed.


Nextra:
:runner:

PR #266, opened 2015-07-20
Introduce automatic config file for plugins and two forwards

Arkshine:
Using the new `AutoExeConfig` native, plugin can now generates automatically a config file which will be pre-populated with cvars and theirs associated description/default value, then executed automatically at map change.

The order of execution is the following: _server.cfg_ -> _amxx.cfg_ -> _plugin config_ -> _per-map config_

The name of output file is customizable.
The output folder for these configs file is `amxmodx/configs/plugins`. It's possible to create as well sub-folder with the native in `plugin` directory.

Default output will be similar to this:
```
// This file was auto-generated by AMX Mod X (v1.8.3-dev+4822)
// Cvars for plugin "Test plugin" by "Arkshine" (test-plugin.amxx, v1.0)

// This is a description!
// -
// Default: "42"
// Minimum: "1.000000"
// Maximum: "42.000000"
MyCvar "42"

// This is a description
// with new line!
// -
// Default: "Egg"
MyCvar2 "Egg"
```

In addition of the native, two forwards:
* `OnAutoConfigsBuffered`:  Called after `amxx.cfg` and  all `AutoExecConfig()` exec commands have been added to the server command buffer.
* `OnConfigsExecuted`: Called after all configs are done executing.


amxmodx/CoreConfig.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+#include "amxmodx.h"
+#include "CoreConfig.h"
+#include "CLibrarySys.h"
+#include <amxmodx_version.h>
+
+CoreConfig CoreCfg;
+
+const char *MainConfigFile = "amxx.cfg";
+const char *AutoConfigDir  = "/plugin";
```
xPaw: Why not `/plugins`? `maps` is used as a plural.

amxmodx/CoreConfig.cpp
```diff
@@ -0,0 +1,324 @@
...
+		{
+			return can_create;
+		}
+
+		FILE *fp = fopen(file, "wt");
+
+		if (fp)
+		{
+			fprintf(fp, "// This file was auto-generated by AMX Mod X (v%s)\n", AMXX_VERSION);
+			fprintf(fp, "// Cvars for plugin \"%s\"\n", plugin->getName());
```
xPaw: How about using plugin title and author (and perhaps additionally version it was generated for)?

amxmodx/CoreConfig.cpp
Arkshine: Hmm, yeah, plural sounds more appropriate and consistent.

xPaw:
What about adding an `amxx` command to regenerate a config for plugin? After a huge plugin update for example.

amxmodx/CoreConfig.cpp
Arkshine: Sounds like a good idea. Feel free to propose a format.

Arkshine:
I don't know, you could simply delete all files, and it will do the same too.

amxmodx/CoreConfig.cpp
xPaw: `// This file contains cvars for "%title%" by "%author%" (%name%, %version%)`

amxmodx/CPlugin.cpp
```diff
@@ -443,6 +444,45 @@ void CPluginMngr::CPlugin::unpausePlugin()
...
 }
 
+void CPluginMngr::CPlugin::AddConfig(bool create, const char *name, const char *folder)
+{
+	// Do a check for duplicates to prevent double-execution
+	for (size_t i = 0; i < m_configs.length(); ++i)
+	{
+		auto config = m_configs[i];
+
+		if (config->autocfg.compare(name) == 0 && config->folder.compare(folder) == 0 && config->create == create)
```
Nextra: the create flag should probably irrelevant for this check

Nextra:
bcompat is unfortunately hacky, but lgtm. :ship:

Sylvert0:
I have an issue with this:

Plugin:
```
#include <amxmodx>

public plugin_init()
{
	register_plugin("cfg test", "1", "Sylvert");
	create_cvar("cvar_test", "1");
	AutoExecConfig();
}
```
The file cstrike/addons/amxmodx/configs/plugins/plugin.cfg.cfg exists but i have this error on console:
```
exec cstrike/addons/amxmodx/configs/plugins/plugin.cfg.cfg: invalid filename.
```

PR #267, opened 2015-07-20
Add enable/disable_event() native (bug 6360)

Arkshine:
Nothing much to say. This uses a state system for easiness which seems appropriate in the context of the event system dealing with reentrancy..

### DIFF
--- a/amxmodx/amxmodx.cpp
+++ b/amxmodx/amxmodx.cpp
@@ -19,6 +19,7 @@
 #include "nongpl_matches.h"
 #include "format.h"
 #include <amxmodx_version.h>
+#include "CEvent.h"
 
 extern CFlagManager FlagMan;
 ke::Vector<CAdminData *> DynamicAdmins;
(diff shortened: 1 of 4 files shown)


plugins/include/amxmodx.inc
```diff
@@ -496,6 +496,15 @@ native console_cmd(id, const cmd[], any:...);
 native register_event(const event[], const function[], const flags[], const cond[] = "", ...);
 
 /**
+ * Unregisters a function hook of a game event which has been previously registered with register_event().
+ *
+ * @param handle    Value returned from register_event()
+ *
+ * @return          1 on successfully uneregistering, 0 on failure
```
xPaw: Fix wording

amxmodx/amxmodx.cpp
```diff
@@ -1722,26 +1722,36 @@ static cell AMX_NATIVE_CALL get_concmdsnum(AMX *amx, cell *params) /* 1 param */
...
 		return 0;
 	}
 
-	sTemp = get_amxstring(amx, params[2], 0, len);
-	iFunction = registerSPForwardByName(amx, sTemp, FP_CELL, FP_DONE);
+	const char* callback = get_amxstring(amx, params[2], 0, len);
+
+	auto event = g_events.findEvent(plugin, evenId, callback);
+
+	if (event)
```
Nextra: This will cause bcompat issues and I think we should go with the Ham-style enable/disable API.

Nextra:
Fix the @return in the doc and :ship:

PR #270, opened 2015-07-28
Fix a ML parser crash on key with an associated empty value

Arkshine:
Reported by `New.ZM.Life`.
Related to #260.

Fix a crash with ML parser not expecting a null value. 
For now, this will just trigger the same "Invalid multi-lingual line" error message.


Nextra:
Can `ReadINI_NewSection` also pass null? If not, :ship:

Arkshine:
No.

PR #271, opened 2015-07-29
Fix ns_remove_upgrade() native (bug 3394)

Arkshine:
Bug is essentially that the item upgrade state is not updated, so even if item is deleted from player's related internal lists, game will still think upgrade is "done". Native partially works, the glitch is at next spawn, upgrade is removed, but you won't be able to select this upgrade anymore. 

Fix is just about calling `mCombatNodes.SetResearchDone(impulseid, false)` on the related upgrade impulse index.

It was a pain to understand how NS is working, but luckily NS released the source code in 2014. This bug was bothering me.

Note: The whole NS module needs to be refactored, it was traumatizing to do the same. :scream: 
Note2: This has been tested, should be fine.


modules/ns/natives/player_memory.cpp
```diff
@@ -378,6 +388,9 @@ static cell AMX_NATIVE_CALL ns_remove_upgrade(AMX *amx, cell *params)
 
 	if (bfound)
 	{
+		void *pTechTree = reinterpret_cast<char*>(player->GetEdict()->pvPrivateData) + MAKE_OFFSET(COMBAT_TECHTREE);
```
Arkshine: Note: This is to access player's private data `AvHTechTree  mCombatNodes;` Then, we can do afterward `mCombatNodes.SetResearchDone()`

Nextra:
So I *think* I understand the code, and it seems to be fine. Given that it's quite a small change and you tested it: :ship:

PR #273, opened 2015-07-29
Fix typo in newmenus (bug 6394)

Arkshine:
I can't believe this has not been spotted after several reviews nor compiler would not whine about it. Too bad we can't do += directly.

Related to #260.


Nextra:
oops, :ship:

Myszax:
Anyway Thanks guys! :+1:

PR #276, opened 2015-07-31
Fix "ham hooks" command not displaying all current hooks of a plugin

Arkshine:
:scream:


Nextra:
the typos are real. :ship:

PR #278, opened 2015-07-31
Update cstrike module - 2

Arkshine:
This patch does the following
* Added some known constants from CS, a `cstrike_const.inc` has been created for the occasion.
* Added `cs_get_item_id` native to get a `CSI_` value from a given alias, `weapon_` or `item_`
* Added `cs_get_translated_item_alias()` native to get an item name (classname) from a given alias
* <strike>Added `CS_OnGetItemPrice` forward to be called after an user bought an item and got it, but before money is used. That's it, you can change the original price value. This param by ref could have been implemented in `CS_OnBuy` but for practical, design, and technical reasons, this is a bad idea.</strike>
* Fixed an issue with `CS_OnBuy`, while testing buying ammos, I've noticed forward would be called only one time and not for all ammos given. It's because the first time forward is triggered by the command, but afterwards game is looping internally to give as much as possible.
* Fixed more things, check 2964d261038156d1625ab576b5d1381987346f1a
* Added `cs_get_weapon_info()` native which gives various information about a weapon (only), like weapon/clip cost, clip size on buy, clip size of gun, max rounds and ammo type. It's likely not worth to add a set_ version since such data comes from a static array used for alias command only.
* Added `cs_find_ent_by_owner()` native because the engine version is used often and this makes sense we add the same for CS.
* Added `cs_get_weapon_class()` and `cs_is_valid_itemid()` stocks for convenience

I need to check/test this more, especially the around the alias to make sure values are correct, but created PR in hoping my retardness will be caught and fixed.


plugins/include/amxconst.inc
```diff
@@ -154,6 +155,18 @@ public stock const Float:NULL_VECTOR[3];
 #define CSW_VEST            31  // Custom
 #define CSW_VESTHELM        32  // Custom
 #define CSW_SHIELDGUN       99
+#define CSW_LAST_WEAPON     CSW_P90
+
+stock const CSW_ALL_WEAPONS      = (~(1<<CSW_VEST));
```
xPaw: Why aren't these in cstrike_const?

Arkshine: Well, you have native like `get_user_weapon` in core which doesn't need cstrike be included and I thought that it's too late to move them, since some old plugins could fail to be recompiled.
We could `cstrike_const.inc` could be included in this file, so sure if it makes sense though.

xPaw: Yeah, moving all cstrike related consts into cstrike_const.inc and just including it would make sense.

gamedata/modules.games/game.cstrike.txt
```diff
@@ -99,5 +107,338 @@
...
+				"c90"
+				{
+					"itemid"    "30"  // CSW/I_P90
+					"classid"    "4"  // CS_WEAPONCLASS_SUBMACHINEGUN
+				}
+			}
+
+			"BuyEquip"
+			{
+                "hegren"
```
Nextra: tab/space inconsistency

modules/cstrike/cstrike/CstrikeNatives.cpp
```diff
@@ -1737,6 +1739,164 @@ static cell AMX_NATIVE_CALL cs_find_ent_by_class(AMX* amx, cell* params)
...
+		edict_t *pev = G_HL_TypeConversion.cbase_to_edict(pEntity);
+
+		if (!FNullEnt(pev) && pev->v.owner == pOwner)
+		{
+			int index = ENTINDEX(pev);
+
+			if (index != -1)
+			{
+				return index;
+			};
```
Nextra: style

Nextra:
should be fine now

Arkshine:
I actually did not finish to refactor code around forward as I've spotted some specific issues related to ammos and things  not being blocked when forward is in blocking-mode. Will push soon.

Nextra:
CS really doesn't want you messing with this.

Sylvert0:
@Arkshine  what happends with CS_NORESET from #287 ?

Arkshine:
@Sylvert0 Do you mean the constant which has magically disappeared? Why it's gone, I'm not sure...

Sylvert0:
@Arkshine yes

modules/cstrike/cstrike/CstrikeHacks.cpp
```diff
@@ -207,182 +219,199 @@ DETOUR_DECL_MEMBER0(GiveDefaultItems, void)  // void CBasePlayer::GiveDefaultIte
...
 	{
-		DETOUR_MEMBER_CALL(AddAccount)(amount, bTrackChange);
+		BlockMoneyUpdate = false;
+		return;
 	}
-	// Shield is blocked.
-	// We need to hook EmitSound to block pickup sound played right after.
-	else if (CurrentItemId == CSI_SHIELDGUN)
+
+	DETOUR_MEMBER_CALL(AddAccount)(amount, bTrackChange);
```
Sylvert0: In the ReGameDLL_CS the function AddAccount have an additional param, causing the server to crash.

AMXX: AddAccount(int amount, bool bTrackChange)
ReGameDLL_CS: AddAccount(int amount, RewardType type, bool bTrackChange)

https://github.com/s1lentq/ReGameDLL_CS/search?p=1&q=AddAccount

I use @_ZN11CBasePlayer10AddAccountEi10RewardTypeb for AddAccount on the file amxmodx/data/gamedata/modules.games/custom/game.cstrike.txt

Mistrick: With ReGameDLL try to use [ReAPI](https://github.com/s1lentq/reapi) instead cstrike.

Arkshine: Not sure why s1lent broke the compatibility with the original version.

Not something you can fix with gamedata alone, the module needs to be updated. That's not going to happen, for now, therefore your choices are either you add yourself the missing argument where the function is detoured in the code, you use ReGameDLL API, or as said above you're using ReAPI.

Sylvert0: I try to compile it adding the additional param, but i have an error:

```
[7536] c++ -pipe -fno-strict-aliasing -Wall -Werror -Wno-uninitialized -Wno-unused -Wno-switch -Wno-format -Wno-format-security -m32 -fvisibility=hidden -Wno-parentheses -g3 -ggdb3 -Wno-invalid-offsetof -std=c++11 -fvisibility-inlines-hidden -fno-exceptions -fno-rtti -D_LINUX -DPOSIX -DLINUX -DAMX_NOPROPLIST -DPAWN_CELL_SIZE=32 -DAMXMODX_BUILD -DAMXX_USE_VERSIONLIB -DAMXX_GENERATED_BUILD -I /root/amxx/build/includes -I /root/amxx/amxmodx/support/versionlib -I /root/amxx/amxmodx/public -I /root/amxx/amxmodx/public/sdk -I /root/amxx/amxmodx/public/amtl -I /root/amxx/amxmodx/public/amtl/amtl -I /root/amxx/amxmodx/public/memtools -I /root/amxx/amxmodx/third_party -I /root/amxx/amxmodx/third_party/hashing -I /root/amxx/amxmodx/third_party/zlib -I /root/amxx/amxmodx/third_party/hashing -H -c /root/amxx/amxmodx/third_party/hashing/hashers/sha256.cpp -o hashers_sha256.o
Traceback (most recent call last):
  File "/usr/bin/ambuild", line 27, in <module>
    main()
  File "/usr/bin/ambuild", line 23, in main
    if not run.Build(os.path.abspath(folder), options, argv):
  File "/usr/lib/python2.6/site-packages/ambuild2/run.py", line 103, in Build
    return cx.Build()
  File "/usr/lib/python2.6/site-packages/ambuild2/context.py", line 128, in Build
    return self.build_internal()
  File "/usr/lib/python2.6/site-packages/ambuild2/context.py", line 163, in build_internal
    if not builder.update():
  File "/usr/lib/python2.6/site-packages/ambuild2/builder.py", line 108, in update
    success = tm.run()
  File "/usr/lib/python2.6/site-packages/ambuild2/task.py", line 553, in run
    self.spewResult(message)
  File "/usr/lib/python2.6/site-packages/ambuild2/task.py", line 489, in spewResult
    util.WriteEncodedText(sys.stderr, message['stderr'])
  File "/usr/lib/python2.6/site-packages/ambuild2/util.py", line 428, in WriteEncodedText
    fd.write(text)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xf3' in position 38: ordinal not in range(128)
```

PR #279, opened 2015-08-04
Fix an SortADTArray() issue where strings are not properly sorted

Arkshine:
Related to 5a6d3fde61d5d5993bd5c0995b46c42e79f57ccd.

A situation of mixed `char*` and `cell*`. The arguments `str1` and `str2` were returning only the first character of the string, resulting an array partially sorted.


Nextra:
:ship:

PR #280, opened 2015-08-07
Cleanup/Fix CS gamedata of entities offsets and add support for HLDM, DoD, TFC and OpFor

Arkshine:
I've fixed a bit my parser, so here an update.

* Fixed one typo in a CS class
* FIxed some missing CS classes
* Fixed `PackageString` and CS master file
* Added class hierarchy as comment at the end of each file
* Moved anything unrelated to entities directly to its own directories. `entities.games` directory should contain only classes derived of `CBaseEntity` 
* Add support for HLDM, TFC, DoD, and OpFor. Will try later Svencoop, Ricochet and TS, this should cover more of played games.


NiHiLaNTh2:
Any plans on Half-Life Deathmatch?

Arkshine:
Ah silly me, forgot `valve`.

PR #282, opened 2015-08-10
Modify game config parser to support more offset data and reflect changes

Arkshine:
This hardcodes more infos about an offset, for now: type (integer, float, etc.), array size and whether value is unsigned. 
This will allow to create a bunch of safe natives to read/write values from a given offset class/name where its data type can be checked for.


Nextra:
Code lgtm. I can't reasonably review all of the gamedata.

Arkshine:
Since gamedata has been generated from a dump of pahole, the only thing which could be wrong is the linux extra offset and the type. But since I've full reviewed at least CS and maybe DoD, this should be fine, I hope!

Nextra:
:shipit:

PR #283, opened 2015-08-13
Fix MPROP_NOCOLORS menu property not working

Arkshine:
Reported by GordonFreeman (RU) here: https://forums.alliedmods.net/showthread.php?t=269968

Just a logic issue where `true`/`false` is reversed.

Bug existed since the creation of API. :scream:


Nextra:
Yay \o/
:sheep:

PR #284, opened 2015-08-13
Introduce new natives to read/write on entity's private data based off class/member name using gamedata files

Arkshine:
Based on #282.

This introduces a new set of natives to handle an entity's private data. 

Unlike the `[get|set]_pdata_*` natives that require compiling the class member offset into the plugin, those new ones instead retrieve the necessary offset from the AMXX gamedata files at runtime, based on the provided class and member name.

The use of gamedata files has the advantage to make these natives much safer than `[get|set]_pdata_*` as it can perform stricter offset and typing checks.

### DIFF
--- a/modules/fakemeta/pdata_gc.cpp
+++ b/modules/fakemeta/pdata_gc.cpp
@@ -1,0 +1,537 @@
+// vim: set ts=4 sw=4 tw=99 noet:
+//
+// AMX Mod X, based on AMX Mod by Aleksander Naszko ("OLO").
+// Copyright (C) The AMX Mod X Development Team.
+//
+// This software is licensed under the GNU General Public License, version 3 or higher.
+// Additional exceptions apply. For full license details, see LICENSE.txt or visit:
+//     https://alliedmods.net/amxmodx-license
+
+//
+// Fakemeta Module
+//
+
+#include "fakemeta_amxx.h"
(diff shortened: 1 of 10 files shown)


modules/fakemeta/pdata_gc.cpp
```diff
@@ -0,0 +1,537 @@
...
+		}
+	}
+
+	return BaseFieldType::None;
+}
+
+#define GET_TYPE_DESCRIPTION(position, data, baseType)                                     \
+	int classLength, memberLength;                                                         \
+	const char *className  = MF_GetAmxString(amx, params[position + 1], 0, &classLength);  \
+	const char *memberName = MF_GetAmxString(amx, params[position + 2], 1, &memberLength); \
```
Nextra: Why is this offset by 1/2 instead of 0/1? It looks like you are retrieving the same parameter twice in code:

```C++
int entity = params[1];
CHECK_ENTITY(entity);

TypeDescription data;
GET_TYPE_DESCRIPTION(1, data, BaseFieldType::Integer);
```

Arkshine: Don't know why I've passed the position zero-based. It should be for sure more understandable that position index starts to 1, likes params.

Nextra:
:ship:

PR #285, opened 2015-08-18
Fix cs_set_user_team

Sylvert0:
Dont update the model :+1:


Arkshine:
This is already the case by default.

`CS_DONTCHANGE` means model is not updated.

PR #286, opened 2015-08-18
Fix cs_set_user_team

Sylvert0:
When the player change of team with CS_DONTCHANGE its change anyway! 
Plugin to test: http://pastebin.com/HCixm6UL

### DIFF
--- a/plugins/include/cstrike.inc
+++ b/plugins/include/cstrike.inc
@@ -20,7 +20,7 @@
 #if !defined AMXMODX_NOAUTOLOAD
 	#pragma loadlib cstrike
 #endif
-
+CS
 /**
  * @section Team and team model constants, used by cs_[get|set]_user_team().
  */
@@ -475,6 +475,7 @@ native cs_set_user_plant(index, plant = 1, showbombicon = 1);
  * @param index         Client index
  * @param team          Team id
  * @param model         Internal model id, if CS_DONTCHANGE the game will choose the model
+ *			or use CS_NORESET if you dont want update the model.
  * @param send_teaminfo If true, a TeamInfo message will be sent
  *
  * @noreturn
--- a/plugins/include/cstrike.inc
+++ b/plugins/include/cstrike.inc
@@ -32,7 +32,7 @@
  */
 enum CsInternalModel
 {
-	CS_AUTOCHANGE = -1,
+	CS_NORESET = -1,
 	CS_DONTCHANGE = 0,
 	CS_CT_URBAN = 1,
 	CS_T_TERROR = 2,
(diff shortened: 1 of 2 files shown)


plugins/include/cstrike.inc
```diff
@@ -32,6 +32,7 @@
  */
 enum CsInternalModel
 {
+	CS_AUTOCHANGE = -1
```
Arkshine: Come one, at least try to compile before doing a PR.

Arkshine:
This breaks backward compatibility as it is. Old compiled plugins which have `CS_DONTCHANGE` set, should expect `ResetModel()` be called by default. 

Same for default value in native header, this will change behavior for existing plugin relying on default value, coder should not have to modify their plugin.

Though, this is a specific situation where you set model before team at the same frame, this can be checked, and for sure this doesn't make sense to reset model if you set before a new one.

I see three ways:
* You change the default value to `CS_AUTOCHANGE`, but then, this would require more checks to match only context of having setting model right before team.  Basically resetting should happen only if you pass `CS_DONTCHANGE` and above, or player doesn't have custom model, or player haven't just got a custom model (not applied yet).
* You rename `CS_AUTOCHANGE` to something like maybe `CS_NORESET`, then all you need to do change is to allow reset with any values `>= 0`. 
* We could argue that this can be just documented that setting a model should be done after setting a team. 

First will require some more code to make sure it fixes only this issue. I think it needs to track when `UpdateModel` is invoked. 
Second is easier and safer, there is no bcompat issues, and we still give a way to fix such situation.
Third is the lazy way but I don't see any reasons not to provide a way to deal with this behavior.

I don't have preferences for now, just posting some thoughts.

Arkshine:
Using the second way is fine.

For doc, I would use more this for consistency: "or if CS_NORESET the game will not update it".

By the way, you should avoid editing directly on github, but instead working on a branch on your computer, then pushing this branch to create a PR. This way, if you need to modify your commits you can do it easily. Some help https://wiki.alliedmods.net/Git_Tutorial#Contributing_via_Forks. You have also software like https://desktop.github.com/ or http://www.syntevo.com/smartgit/.

Anyway, you have 7 commits for small changes, please squash them to one commit.(some help: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html)

Arkshine:
Can you squash your commits as said above, please?

Sylvert0:
I delete for mistake the repository trying to squash the commits xD So i forked again the repository and make a new branch https://github.com/Sylvert0/amxmodx/tree/Fix-cs_set_user_team you can close this?

Arkshine:
I'm not sure how you did that by mistake as to delete a repo you need to go in options and confirming it. If you wanted some help you could have asked. But well, okay.

PR #287, opened 2015-08-22
Add a CS_NORESET constant for use with cs_set_user_team for skipping the model reset

Sylvert0:
When the player change of team with CS_DONTCHANGE its change anyway! 
Plugin to test: http://pastebin.com/HCixm6UL


Arkshine:
Should be good now. Thanks.

PR #289, opened 2015-08-27
Fix typo in AutoExecConfig native

Arkshine:
Related to #266. Looks like I've changed the prefix afterward at some point for some reason, but engine is retarded and with `plugin.something.cfg` it will understand extension is `something` instead  of `cfg` ; and will fail to execute as it's an invalid extension (only `cfg` and `rc` are allowed).

Prefix is now `plugin-`.


Nextra:
silly goldsrc. :ship:

PR #290, opened 2015-08-30
Added CSI_KNIFE (CSW_KNIFE) and CSI_MAX_VALUE

collinsmith:
I believe CSI_KNIFE was missed when copying over the constants, as there is no corresponding value for CSW_KNIFE. I also felt that defining the max value is useful for iterations and declaring an array where the CSI_* constants can be used to index. The benefit of using this over CSI_SHIELD is that if more are added in the future (similar to how CSI constants were added in preference to the CSW constants), then this value will only need to be changed to represent the new highest constant.


Arkshine:
Thanks, but I've actually already fixed in the opened PR #278.

collinsmith:
Ahh, I see. Thanks for referencing it for me.

PR #291, opened 2015-09-05
Fix crash when dynamic native is executed with the max number of params

Arkshine:
This fixes a buffer overflow with `g_Params`, which has a size of 16 but params index starts to 1. As consequence when 16 params are passed, `g_Params[16]` is indexed and likely overwrites something important which leads to a crash at some point.

It doesn't happen in 1.8.2. Possible reasons why it happens in dev version is this uses newer compiler and compiling optimization are better , resulting crash is triggered now. 

Reported here https://forums.alliedmods.net/showthread.php?t=271103.


Nextra:
oops. :sheep:

PR #292, opened 2015-09-08
Fixed possible issue in native params

collinsmith:
See https://bugs.alliedmods.net/show_bug.cgi?id=6412

### DIFF
--- a/amxmodx/debugger.cpp
+++ b/amxmodx/debugger.cpp
@@ -935,7 +935,7 @@ static cell AMX_NATIVE_CALL dbg_trace_info(AMX *amx, cell *params)
 	pDebugger->GetTraceInfo(pTrace, lLine, function, file);
 
 	set_amxstring(amx, params[3], function ? function : "", params[4]);
-	set_amxstring(amx, params[5], file ? file : "", params[5]);
+	set_amxstring(amx, params[5], file ? file : "", params[6]);
 	*line_addr = (cell)lLine + 1;
 
 	return 1;


collinsmith:
After compiling with this change made, the hang still occurs when the native is executed. I will investigate further.

collinsmith:
Debugger pointed me at https://github.com/alliedmodders/amxmodx/blob/master/amxmodx/debugger.cpp#L341

collinsmith:
I've noticed some interesting behavior after playing around with my pawn code. If I execute only:

new trace = dbg_trace_begin();
server_print("Trace = %d", trace);

The result trace handle is 0 (invalid trace), however if I execute the entirety:

    new trace = dbg_trace_begin();
    server_print("Trace = %d", trace);
    new line, function[32], file[32];
    dbg_trace_info(
            trace,
            line,
            function,
            31,
            file,
            31);

The result is 1004 (i.e., non-zero, and a valid handle). At minimum, if I remove the dbg_trace_info native, leaving only the declarations, I also receive a non-zero return.

My assumption is that this is an invalid function in the code to perform a trace in (thus the crash), however some other issue is causing a "valid" trace handle value to be returned.

Nextra:
It's hard to tell what you're trying to do from your bug report, but do you actually use this native [where it is supposed to be used](https://github.com/alliedmodders/amxmodx/blob/master/plugins/include/amxmodx.inc#L2742)? You can only debug a trace if you have a trace, which is to say that you need to be inside an error handler that gives you one. These debug natives don't work in ordinary functions.

Arkshine:
Looking at code, the handle returned by `dbg_trace_begin` is only set when a log error occurs from a native. What you get is likely garbage as the member which holds the handle doesn't seem to be initialized.

collinsmith:
I agree, it looks like it could be garbage. Can we amend the doc to state that this is only valid when an error has been thrown and fix the value to not return the garbage?

As a workaround I think I will try and catch my own error.

Nextra:
The documentation for `dbg_trace_begin()` states that it is intended to be used inside an error filter set by `set_error_filter()`. If your error filter is called it is guaranteed an error has been trapped, and therefore a trace will be ready. I don't know why it apparently returns a valid trace outside of error functions, so that is something worth looking into.

collinsmith:
My assumption based on the documentation was that dbg_trace_begin was specifically created for use within an error filter, however it could also be use outside of one to get the current call stack. It did not specify that it could/should not be used in other functions, similar to other natives, e.g., dynamic parameter natives which throw errors if used outside a dynamic native body. This is what I meant by amending the documentation.

I'm typing this on my phone, please forgive poor grammar

dvander:
It seems like a simple solution to this problem would be to have all your natives take in a required "context" parameter which describes their caller. It could just be a string, and (I think) you already have the invoking plugin. If Pawn had features from other languages like RAII or with/using-scopes there would be more solutions; but alas.

I'm not a huge fan of grabbing callstacks to find the nearest caller. Wrapper or helper functions can  obscure the callstack, and working around that encourages some weird designs.

collinsmith:
I completely agree. This is the kind of design I'm trying to avoid. I am currently passing a context param (function name), however this is not ideal.

Sylvert0:
My server crash with only setting the error filter and after calling the function who have an error

```
public plugin_precache()
{
    set_error_filter("plugin_error_filter");
}

public plugin_error_filter(err, debugging, msg[])
{
	server_print("[ERROR FILTER] err: %d - debugging: %d - msg: %s", err, debugging, msg);
}
```

HLDS Version: 6153
AMXX Version: 1.8.3-dev+4882
Metamod version: 1.27p37
SO: Windows XP

PR #293, opened 2015-09-20
Sockets update

Javivi:
Finished working in this some months ago but didn't end up sending a pull request.

Changelog:
- WinSock version changed from 1.1 to 2.2.
- Properly check for WinSock initialization on OnAmxxAttach/Detach.
- Now, natives will not be added if we can't start up WinSock.
- socket_open is now IP version agnostic (both IPv4 and IPv6 are
supported).
- Error reporting has been changed on socket_open, now it returns the
actual libc errors instead of the previous made-up errors.
- socket_close now returns a value on success/failure.
- Code rewritten to be more readable, it should be self-explaining now.

### DIFF
--- a/amxmodx/CFileSystem.h
+++ b/amxmodx/CFileSystem.h
@@ -12,7 +12,6 @@
 
 #include <FileSystem.h> // IFileSystem, FileSystemSeek_t, FileHandle_t (HLSDK)
 #include <stdio.h>      // FILE*
-#include <am-cxx.h>     // KE_OVERRIDE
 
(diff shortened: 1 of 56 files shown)


Arkshine:
> Error reporting has been changed on socket_open, now it returns the actual libc errors instead of the previous made-up errors.
Native should be backward compatible. If old plugin relies on specific error and you changed that, this may have some unexpected and unwanted behavior. Either you deal with an existing param, or you add a new one, or ultimately you add a new native.

If you update documentation, please do it properly. Check amxmodx.inc for example.

PR #294, opened 2015-09-25
Customizable Logger

collinsmith:
I've added a customizable Logger into the AMXX core and would appreciate if someone could now take a look, as I believe it is ready for general testing/review. Please let me know any improvements, questions, comments or changes you would like to make.

**What does it do:**
This patch adds the Logger data structure into the AMXX core. In each logger you can customize a variety of options, such as:

| Item        	|         Default        	| Description                                                              	|
|-------------	|:----------------------:	|--------------------------------------------------------------------------	|
| verbosity   	|      Severity_Warn     	| Only log messages greater or equal to this verbosity will be logged      	|
| name        	|          %p_%d         	| Name of the logger, this example will create one log file per unique day 	|
| message     	| [%5v] [%t] %n::%f - %s 	| Format of a log message                                                  	|
| path        	|                        	| This will create a log file in */cstrike/addons/amxmodx/logs/*           	|
| stack trace 	|     at %n::%f : %l     	| Format of a stack trace                                                  	|
| date        	|        %Y-%m-%d        	| Format of the date (standard C format)                                   	|
| time        	|        %H:%M:%S        	| Format of the time (standard C format)                                   	|

Severity | Value
------------ | -------------
Severity_None | 0
Severity_Debug | 1
Severity_Info | 101
Severity_Warn | 201
Severity_Error | 301

Specifier | Function
------------ | -------------
%d | date
%f | function
%i | counter
%l | line #
%m | map
%n | script (with .sma)
%p | plugin file (without .amxx)
%s | log message
%t | time
%v | verbosity/severity
%% | % literal 

**Why does AMXX need this type of addon?**
It's simple, this makes debugging and creating effective logging easier for the developer. While the current logging mechanism works great for some cases, it leaves the AMXX logs feeling bloated for large plugins which require a large amount of logging. This this data structure, developers and server admins can choose what messages they want to see. Some might want to develop their modules in debug mode and then disable logging completely for a performance boost. This module was designed to cut out of the native as soon as the message is not within the severity parameters of the server.

**Some quick notes:**
- I chose to double up on the natives to try and promote speed and lightweightedness of the module, so there wouldn't be any stocks performing argument formatting (vformat), and the log native will cease execution ASAP.
- documentation is still required for the logger natives, particularly the specifiers, however I wanted to nail down the specification and get general review before doing so, general info can be found [here](https://forums.alliedmods.net/showpost.php?p=2346066&postcount=20)
- I already added a test suite to test the majority of the natives
- I've added a CVAR manager plugin that would need to be added to the core plugins, which creates and manages CVARs for logging verbosities (useful for server owners who want to adjust on the fly or disable all together).
- This has also committed my msvs2015 project, which is a conversion of the included msvs2012, however you may wish to remove it.


xPaw:
Your pull request updates all vcproj files with changes that are not related to your pull request, you should revert these so the PR would be smaller and easier to review.

collinsmith:
I understand conceptually how that might be done, but I don't know how to do it in actuality. I'll look into how to either ignore or revert those files to their original state. I did not realize it included changes to each vcxproj

collinsmith:
Okay, I figured it out :). Let me know if you would like me to also remove /amxmodx/msvc15/ . I noticed that the current master doesn't include one for this version of MS VS

collinsmith:
I'm not sure why this occurs by the CI build. Can someone explain to me what changes need to be made to resolve it? I don't receive this error when building locally.
clipboard-1

Arkshine:
AMXX uses AMBuild to compile. You need to modify https://github.com/alliedmodders/amxmodx/blob/master/amxmodx/AMBuilder as well.
And for any extras files which need to be included in released package (includes, plugins, etc.), you need to modify https://github.com/alliedmodders/amxmodx/blob/master/support/PackageScript.

It would be welcomed you described in first post what does this patch to offer a better overview.

amxmodx/logger.cpp
```diff
@@ -0,0 +1,1164 @@
+#include <algorithm>
```
Arkshine: Please do without STL. I did not check your code, but if only for `min`/`max`, you can use `ke::Min`/`Max`.

collinsmith:
@Arkshine I've updated the original commit comment with a description. I will also work on adding in comments for the natives as soon as I get time

collinsmith:
@Arkshine I've updated both of the files you mentioned with the new dependencies. Please let me know of any API change requests or if anything was done which doesn't obey project guidelines.

The last change that I think I need to make is with *iota*, is there a recommended replacement?

collinsmith:
I'm toying around with the idea to add rudimentary exception handling and making use of this Logger. Exceptions would need to be identified using strings, e.g., *throw("IllegalArgumentException")*, but I think it could be done, albeit *rudimentarily*. I will probably wait and make this a separate patch, as I want to get this one okay'd first.


E.g.,
```
setName(33, "Name"); // 33 is invalid index, this native calls *throw("IllegalArgumentException")*
if (catch("IllegalArgumentException")) { // module would store the current exception string that was thrown
    handled(); // would clear the exception and not log it
}

// The trick here would to then have AMXX throw the error if handled() was not called,
// This may not be possible without calling another native
```

Or:
```
try(); {
    setName(33, "Name");
    if (catch("IllegalArgumentException")) {
        handled();
    }
} endTry();
```

The idea with this would be to allow for greater flexibility in plugin APIs for reporting errors and giving  plugins a chance to resolve them instead of just setting a fail state or whatnot. Personally, I've tried to make a few APIs where it's not exactly clear what *true* or *false* mean, since if an error is thrown the native will return *false*, and this might not be accurate.

I understand that there is *set_error_filter*, however this is meant to be a solution that is in-line with where the exception is thrown, versus a separate function.

Is this something you'd be interested in seeing? Has anyone tried this before? I understand this is off-topic, I just wanted to mention it here because the exception handler may be tightly coupled with the Logger.

Arkshine:
For `itoa` question, I guess `ke::SafePrintf` with `%d` should do the trick?

collinsmith:
@Arkshine I made that change, however I noticed a couple others, namely *atoi* and *atoll*. These are all related to *logger.cpp::onMapChange* for saving and loading map counters which reset every day (so count and next reset time are saved). I think maybe *logger.cpp::onMapChange* could be rewritten or improved so they are no longer necessary, I'll look into making these changes.

Arkshine:
These functions should be fine I guess.

Worth to be noted: I've merged yesterday an update to AMTL and I've removed `UTIL_Format` since `ke::SafePrintf` has been introduced. You will need to update/rebase your `master` from upstream.

collinsmith:
I think I rebased it correctly :)

collinsmith:
I've noticed an issue where the current build will crash the server if the debug option is not set on the plugin using the Logger. This is caused by a nullptr in pDebugger, i.e., *amx->userdata[UD_DEBUGGER];*. I'm going to look into this. Obviously I'd like to avoid forcing all plugins to be in debug mode, however is it a possibility that if a plugin registers a logger to then set it in debug mode?

collinsmith:
I'm having trouble figuring out how to set debug mode on a plugin after it's already been loaded. I've been looking at the implementation in *load_amxscript* for some guidance, however in my case I don't have access to some of those pointers (e.g., *program*). In the meantime I just added an error message an an early quit if debug mode is not set (as in *LogError*). This has resolved the crashing issue.

Arkshine:
Looks like you failed to rebase properly, or at least you did not push with --force as warned, you have duplicated commits now. Your PR starts to be quite the mess. I have no thoughts about this yet nor did check really code, and it would seem this needs more work/thinking ; at some point you will  need to rebase again and cleaning your commits list, merging/renaming some to avoid unnecessary noises. Anyway, there is no rush, take your time. Also Nextra is likely to take a look soon.

collinsmith:
I'm creating a new pull request with a different branch. This should resolve all of the aforementioned issues.

PR #295, opened 2015-09-29
Update Function for Load Admins

9iky6:
Add No Admins Message to load admins in file

### DIFF
--- a/plugins/admin.sma
+++ b/plugins/admin.sma
@@ -354,57 +354,49 @@ AddAdmin(id, auth[], accessflags[], password[], flags[], comment[]="")
 
 loadSettings(szFilename[])
 {
-	new File=fopen(szFilename,"r");
+	new File = fopen(szFilename, "r");
 	
 	if (File)
 	{
-		new Text[512];
-		new Flags[32];
-		new Access[32]
-		new AuthData[44];
-		new Password[32];
-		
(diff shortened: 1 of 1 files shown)


luxxxoor:
Please remove the prefixes like "sz"... it looks so ugly...and i don't see any reason to do that.
We should reduce the styles from the plugins to only one, why we have 100styles for only 1 plugin ?

plugins/admin.sma
```diff
@@ -354,57 +354,49 @@ AddAdmin(id, auth[], accessflags[], password[], flags[], comment[]="")
...
 		{
-			fgets(File, Text, charsmax(Text));
+			fgets(File, szString, charsmax(szString));
 			
-			trim(Text);
+			trim(szString);
 			
-			// comment
-			if (Text[0]==';') 
+			if(!szString[0] || szString[0] == ';' || szString[0] == '/' && szString[1] == '/')
```
ClaudiuHKS: if (!szString[0] || szString[0] == ';' || szString[0] == '/' && szString[1] == '/')

You probably mean

if (szString[0] == '^0' || szString[0] == ';' || **(**szString[0] == '/' && szString[1] == '/'**)**)

Arkshine:
Are you still alive?

PR #296, opened 2015-10-01
Update to latest AMTL

Arkshine:


Nextra:
UTIL_FormatArgs should probably be UTIL_VarArgs in that commit message. Otherwise this should be fine. There are a bunch of incosistencies regarding the maxlength parameter (nothing critical as far as I can tell), but that can be done in another PR.

Nextra:
:ship:

PR #297, opened 2015-10-06
Fix some natives relying on a known engine bug related to last player's edict being invalid

Arkshine:
It's an old engine bug (and it looks like still not fixed yet?) where edict of the last player can't be retrieved with `INDEXENT`. 
That's why AMXX uses another way to retrieve a player's edicts list, but it would seem that some natives have been forgotten.

* `emit_sound`: Last player won't hear the sound
* `e/message_bein`,  `dod_weaponlist`: Lead to a crash since engine doesn't like `MSG_ONE_*` with null entity.
* `TFC_[Clear/Set]Model`: Lead to a crash because of null edict.
* `set_tr/2`: Will not work when used with `TR_pHit`
* `pev`: Will return 0 used with `pev_edict`
* `pev_valid`: Will return 0.

Patch includes some consistency changes and to avoid to duplicate each time macro to retrieve a player's edict. But this would need to be more centralized. A real solution will be done in another PR.


Nextra:
lgtm now, :ship:

PR #298, opened 2015-10-09
Introduce get/set_gamerules_*() natives to read/write on the gamerules object

Arkshine:


modules/fakemeta/fakemeta_amxx.cpp
```diff
@@ -39,13 +42,27 @@ void OnAmxxAttach()
...
 
-	char error[256];
-	error[0] = '\0';
+	char error[256] = "";
 
 	if (!ConfigManager->LoadGameConfigFile("common.games", &CommonConfig, error, sizeof(error)) && error[0] != '\0')
 	{
 		MF_Log("Could not read common.games gamedata: %s", error);
-		MF_Log("get/set/find_ent_data* natives have been disabled");
+		MF_Log("get/set_ent_data* natives have been disabled");
```
xPaw: These two logs could be a single one.

modules/fakemeta/fakemeta_amxx.cpp
```diff
@@ -39,13 +42,27 @@ void OnAmxxAttach()
...
+		return;
+	}
+
+	*error = '\0';
+
+	if (!ConfigManager->LoadGameConfigFile("common.games/gamerules.games", &GamerulesConfig, error, sizeof(error)) && error[0] != '\0')
+	{
+		MF_Log("Could not read common.games/gamerules.games gamedata: %s", error);
+		MF_Log("get/set_gamerules_* natives have been disabled");
```
xPaw: Same here.

modules/fakemeta/fakemeta_amxx.cpp
Arkshine: On purpose because `MF_Log` includes automatically EOL and that a line in the log the date and time is automatically inserted before, so if you add manually EOL here it kind of breaks visually the log.

modules/fakemeta/fakemeta_amxx.cpp
Arkshine: Same answer :P

modules/fakemeta/fakemeta_amxx.cpp
xPaw: You don't really need a new line here though. Same as line 65. If you `error` to be at the end, it could formatted like `get/set_ent_data* natives have been disabled because common.games gamedata could not be read: %s`

Arkshine: Well I can change if it traumatizes that much. I personally don't care.

xPaw: I just think there is no need to spread out error messages like this over multiple lines.

modules/fakemeta/pdata_gc.cpp
```diff
@@ -112,112 +132,89 @@ static BaseFieldType GetBaseDataType(TypeDescription &data)
...
+#define CHECK_GAMERULES()                                                                  \
+	if (!GameRulesAddress)                                                                 \
+	{                                                                                      \
+		MF_LogError(amx, AMX_ERR_NATIVE, "%s is disabled. Check your AMXX log.", __FUNCTION__);  \
+		return 0;                                                                          \
+	}
 
-	auto pEntity = TypeConversion.id_to_edict(entity);
 
+cell GetData(void *pObject, TypeDescription &data, int element)
```
Nextra: All these helper functions could move to a separate file to clean this up.

modules/fakemeta/pdata_gc.cpp
```diff
@@ -75,20 +75,40 @@ static BaseFieldType GetBaseDataType(TypeDescription &data)
 	return BaseFieldType::None;
 }
 
-#define GET_TYPE_DESCRIPTION(position, data, baseType)                                     \
+
+#define GET_TYPE_DESCRIPTION(position, data, baseType, list)  \
```
Nextra: I love me some macro trickery but I think with these changes we can remove this now?

Nextra:
Otherwise lgtm. :+1:

Sylvert0:
Arkshine, you can do a plugin showing how to use this? Or for what we can use this

Arkshine:
The new sets of `get/set_ent_data*` and `get/set_gamerules*`natives are similar to `get/set_pdata_*` natives to get/set a value from entity's private data where you need to hard-code in plugin the offset, like:
``` PAWN
const m_iTeam = 114;
new team = get_pdata_int(index, m_iTeam);
```

Both new sets of natives use offsets which are hard-coded in gamedata files and retrieved at runtime (`/data/gamedata/common.games/`, one file per class) and to specify what offset you want to use you would need to specify its name and its associated class name. For example, `m_iTeam` is a player's offset, and the base class for players is [`CBasePlayer`](https://github.com/alliedmodders/amxmodx/blob/master/gamedata/common.games/entities.games/cstrike/offsets-cbaseplayer.txt) so you would write instead:
``` PAWN
new team = get_ent_data(index, "CBasePlayer", "m_iTeam");
```

The gamerules natives are not related to an entity. It's global to game and it's basically used to manage the gameplay and its rules. For example, you could get the CT scores this way:
```PAWN
new score = get_gamerules_int("CHalfLifeMultiplay", "m_iNumCTWins");
```

Do you understand better?

Sylvert0:
Yes, thanks

Arkshine:
Yes, but be careful that the variable name which may be used in existing plugins, may be not the same as defined in gamedata files. Those files have been generated from the game binary so it uses the real  names. You will get anyway an error message if name could not be found, then just a matter to search name in gamedata file from the offset value.

PR #301, opened 2015-10-11
Update sockets module

Javivi:
Been working on this some time. Changelog:
- WinSock version changed from 1.1 to 2.2.
- Properly check for WinSock initialization on OnAmxxAttach/Detach.
- Now natives will not be added if we can't start up WinSock.
- socket_open() is now IP version agnostic (both IPv4 and IPv6 are
  supported).
- Error reporting has been changed on socket_open(), a new parameter
  called _libc_errors has been added, and, if enabled, libc errors will be
  returned instead of the previous made-up errors.
- socket_close() now returns a value on success/failure.
- Added non-blocking sockets at socket_open_nb().
- Added socket_is_writable() to check if a socket is ready for write.
- Added socket_is_readable() as an alias to socket_change().
- Code rewritten to be more readable, it should be self-explaining now.

Sorry for the ugly diff @Arkshine

### DIFF
--- a/modules/sockets/sockets.cpp
+++ b/modules/sockets/sockets.cpp
@@ -3,10 +3,6 @@
 // AMX Mod X, based on AMX Mod by Aleksander Naszko ("OLO").
 // Copyright (C) The AMX Mod X Development Team.
 //
-// Codebase from Ivan, -g-s-ivan@web.de (AMX 0.9.3)
-// Modification by Olaf Reusch, kenterfie@hlsw.de (AMXX 0.16, AMX 0.96)
-// Modification by David Anderson, dvander@tcwonline.org (AMXx 0.20)
-//
 // This software is licensed under the GNU General Public License, version 3 or higher.
 // Additional exceptions apply. For full license details, see LICENSE.txt or visit:
 //     https://alliedmods.net/amxmodx-license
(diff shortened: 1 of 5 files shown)


modules/sockets/sockets.cpp
```diff
@@ -15,233 +11,370 @@
...
+
+	struct timeval tv;
+	tv.tv_sec = 0;
+	tv.tv_usec = timeout;
+
+	fd_set readfds;
+	FD_ZERO(&readfds);
+	FD_SET(sockfd, &readfds);
+
+	return (select(sockfd + 1, &readfds, nullptr, nullptr, &tv) > 0) ? 1 : -1;
```
Arkshine: You will break likely compatibility by returning -1 instead of 0 originally.

modules/sockets/sockets.cpp
```diff
@@ -15,233 +11,370 @@
...
+
+	if(fcntl(sockfd, F_SETFL, flags | O_NONBLOCK) == -1)
+		return errno;
+
+	return 0;
+#endif
+}
+
+// native socket_open_nb(_hostname[], _port, _protocol = SOCKET_TCP, &_error);
+static cell AMX_NATIVE_CALL socket_open_nb(AMX *amx, cell *params)
```
xPaw: This whole function is practically a copypaste of `socket_open`, adding a separate function like `socket_set_nonblock` would be more desirable.

modules/sockets/sockets.cpp
```diff
@@ -15,233 +11,370 @@
...
-    FD_ZERO(&rfds);
-    FD_SET(socket, &rfds);
-    tv.tv_sec = 0;
-    tv.tv_usec = timeout;
-    // Now we "select", which will show us if new data is waiting in the socket's buffer
-    if (select(socket+1, &rfds, NULL, NULL, &tv) > 0) 
-       return 1; // Ok, new data, return it
-     else 
-       return 0; // No new data, return it
+	return (close(params[1]) == -1) ? -1 : 1;
```
xPaw: As @Arkshine mentioned previously, you're returning -1 instead of 0 here.

plugins/include/sockets.inc
```diff
@@ -24,43 +21,159 @@
...
+ * @note Set _timeout to 0 avoid blocking the call
+ * @note A socket will become readable if there's any data or an EOF
+ *
+ * @param _socket    Socket descriptor
+ * @param _timeout   Ammount of time to block the call waiting for the socket to be marked as readable or for the timeout to expire, in µSeconds (1 sec = 1000000 µsec)
+ *
+ * @return           1 if the socket is marked as readable
+ *                   -1 otherwise
+ */
+native socket_is_readable(_socket, _timeout = 100000);
```
xPaw: Unsure if having duplicate function with duplicate description is a good idea. Perhaps mark `socket_change` as deprecated?

modules/sockets/sockets.cpp
Arkshine: `socket_close` should be fine since it was not returning value before.

xPaw: Consistency though. `socket_recv` can return -1 too.

modules/sockets/sockets.cpp
Javivi: It was changed from 0 to -1 to maintain consistency with the other function's return value, because having some functions returning 0 and others returning -1 could be confusing.
Also I didn't thought anyone would check the socket against 0, I thought they would just do !socket.

Return values should be standardized across the module I think, so maybe changing all of them to 0 would be better, even if some of them return -1 by design.

modules/sockets/sockets.cpp
Javivi: Maybe be can use something like:
```SourcePawn
socket_open(_hostname[], _port, _protocol = SOCKET_TCP, &_error, _libc_errors = DISABLE_LIBC_ERRORS, _socket_type = BLOCKING);
```

It can't be done the way you said because 'connect' can block, so we can't call socket_open first and then set that socket a non-blocking. The socket should be set as non-blocking right away as it is created,

modules/sockets/sockets.cpp
Javivi: Yes, as I said before, having consistency is good. Return values should be standardized across the module, so maybe changing all of them to 0 would be better in order to maintain backwards compatibility with socket_change().

modules/sockets/sockets.cpp
xPaw: In that case, shouldn't a boolean be enough then? And put it before `_libc_errors` param.

modules/sockets/sockets.cpp
xPaw: Yep, 0 is reasonable.

plugins/include/sockets.inc
Javivi: Yes, that's a good idea, I vote in favor of having socket_change() deprecated and using socket_is_readable() too.

modules/sockets/sockets.cpp
Javivi: I am not sure I'm following you, can you explain a bit more your idea?

xPaw: Your new param is `_socket_type = BLOCKING`, but what else would socket type be? `is_blocking = true` should be enough.

Javivi: We don't have boolean values on Pawn, so I think ```_socket_type = BLOCKING/NON_BLOCKING``` would look better than ```_is_blocking = 0/1```

And using BLOCKING / NON_BLOCKING would be more readable when writing a plugin.

xPaw: > We don't have boolean values on Pawn

What? It does have booleans.

Javivi: Sorry, I didn't express myself correctly.

I meant that I remember that @Arkshine told me not to use bools on natives, or something like that.

I may be mistaken, it has been a long time since this PR was done, if Arkshine can shed some light it would be good.

modules/sockets/sockets.cpp
Arkshine: Just doing `if (socket_change())` will still break with -1 since expression will be considerated as boolean and because it's a non-zero value, condition will be true.

So yes it should be standardized as much as possible, but unfortunately you can't change suddenly 0 -> -1 or -1 -> 0.

modules/sockets/sockets.cpp
Arkshine: I honestly don't remember saying that to you. You would not explicitly use `bool:` if later others modes could be added. Is there the case here?  Either way both are fine really. 

If you have 2 choices only, a boolean would be logic & consistent though, like `_libc_errors = false, _blocking_mode = true`. I would slightly favor booleans, it seems more appropriate and fairly readable, but I suck at design.

Javivi: I prefer reading something like
```socket_open("1.1.1.1", "80", SOCKET_TCP, errorcode, ENABLE_LIBC_ERRORS, NON_BLOCKING)```
instead of
```socket_open("1.1.1.1", "80", SOCKET_TCP, errorcode, true, true)```

Improves readability and looks more descriptive.

Let's get some more opinions so we can decide. @IgnacioFDM @Nextra @Freeman-AM @WPMGPRoSToTeMa

Comment (xPaw) [inline modules/sockets/sockets.cpp]: If you really want that, you can do `#define NON_BLOCKING false`

Comment (Arkshine) [inline modules/sockets/sockets.cpp]: socket_open("1.1.1.1", "80", SOCKET_TCP, errorcode, .libc_errors = true, .blocking_mode = true)  is for me too. :P

File: modules/sockets/sockets.cpp (hunk 1 above)
Comment (Javivi) [inline modules/sockets/sockets.cpp]:
```/**
 * Receives data.
 *
 * @note The ammount of bytes than you end up receiving can be less than the one you expected
 *
 * @param _socket    Socket descriptor
 * @param _data      Array to save the data
 * @param _length    Length of the array
 *
 * @return           Ammount of bytes received
 *                   0 if the peer closed the connection
 *                   -1 on failure
 */
native socket_recv(_socket, _data[], _length);
```

Every return value but this one can be changed to 0 without any problem.

Changing socket_recv() to 0 would remove the possibility of knowing if the peer closed the connection.

Should we remove this feature in order to keep everything consistent ? I'm not sure it would be of any use, and the socket's status can be check using other functions.

modules/sockets/sockets.cpp
Javivi: Haha, I knew someone was going to point that, @Arkshine.

I think @xPaw idea has the best of both worlds, so if no one objects I will be doing it that way.

Arkshine: That would be really kind of silly. If you don't want to use boolean, then I would prefer a single flag parameter instead.so you can do for example:  `socket_open("1.1.1.1", "80", SOCKET_TCP, errorcode, SOCKET_LIBC_ERRORS | SOCKET_NON_BLOCKING)` or something. It opens room for more flags if needed.

modules/sockets/sockets.cpp
Arkshine: You don't want to break backward compatibility, so no you can't change return value here.

Javivi: I just took a look again at it[1] and yeah, it has to stay at -1. It could return 0 previously, I just documented it.

[1] https://github.com/alliedmodders/amxmodx/blob/master/modules/sockets/sockets.cpp#L143

modules/sockets/sockets.cpp
Javivi: Flags is it then, as that would allow using some TCP flags that may come handy.

Will update next monday probably.

plugins/include/sockets.inc
Javivi: @Arkshine do we have your blessing

Arkshine: I don't know. Kind of silly, but for the sake of consistency and readability, it could make sense.

Javivi:
Simple test plugin

```SourcePawn
#include < amxmodx >
#include <sockets>

#define NODE "httpbin.org"
#define SERVICE 80

new request[] = "GET /get HTTP/1.1^nHOST: httpbin.org^nConnection: close^n^n^n^n"

public plugin_init( )
{
	register_plugin("Socket test", "1.0", "Javivi")

	register_clcmd("say /s1", "test_blocking")
	register_clcmd("say /s2", "test_nonblocking")
}

public test_blocking(id)
{
	new sockfd, socket_error
	new bytes_sent, bytes_recv, bytes_recv_buffer[500]

	sockfd = socket_open(NODE, SERVICE, SOCKET_TCP, socket_error, SOCK_LIBC_ERRORS)

	if(sockfd != -1)
	{
		client_print(id, print_chat, "Got socketfd %d", sockfd)

		bytes_sent = socket_send(sockfd, request, charsmax(request))
		client_print(id, print_chat, "[%d/%d] bytes sent", bytes_sent, charsmax(request))

		bytes_recv = socket_recv(sockfd, bytes_recv_buffer, charsmax(bytes_recv_buffer))
		client_print(id, print_chat, "[%d/%d] bytes received", bytes_recv, charsmax(bytes_recv_buffer))
		client_print(id, print_chat, "%s", bytes_recv_buffer)

		socket_close(sockfd)
	}
	else
	{
		client_print(id, print_chat, "Error %d", socket_error)
	}
}

new nb_sockfd

public test_nonblocking(id)
{
	new socket_error

	nb_sockfd = socket_open(NODE, SERVICE, SOCKET_TCP, socket_error, SOCK_NON_BLOCKING|SOCK_LIBC_ERRORS)

	if(nb_sockfd != -1 && SOCK_ERROR_EINPROGRESS(socket_error))
	{
		client_print(id, print_chat, "Got nonblocking socket %d", nb_sockfd)

		set_task(1.0, "check_nbsocket", 1234, .flags="b")
	}
	else
	{
		client_print(id, print_chat, "Error %d", socket_error)
	}
}

public check_nbsocket(task)
{
	if(socket_is_writable(nb_sockfd))
	{
		remove_task(task)

		client_print(0, print_chat, "Socket %d is writable", nb_sockfd)

		new bytes_sent = socket_send(nb_sockfd, request, charsmax(request))
		client_print(0, print_chat, "[%d/%d] bytes sent", bytes_sent, charsmax(request))

		set_task(1.0, "recv_nbsocket", 1234, .flags = "b")
	}
}

public recv_nbsocket(task)
{
	new bytes_recv, bytes_recv_buffer[500]

	bytes_recv = socket_recv(nb_sockfd, bytes_recv_buffer, charsmax(bytes_recv_buffer))

	if(bytes_recv)
	{
		remove_task(task)

		client_print(0, print_chat, "[%d/%d] bytes received", bytes_recv, charsmax(bytes_recv_buffer))
		client_print(0, print_chat, "%s", bytes_recv_buffer)

		socket_close(nb_sockfd)
	}
}
```

**NOTE**: The first call to /s2 will probably block, as the name resolution server will have to be called. To avoid that a numeric ip address has to be used, as noted in the docs.


**FOR THE FUTURE**:
- Add a new function that lets you set more flags on the socket (like TCP_CORK, TCP_NODELAY...)
- Adding SSL support
- Adding forwards (can be done easily but there's problem with it being multiplatform)

WPMGPRoSToTeMa:
@Javivi ^n -> ^r^n?
You can resolve name in another thread. (non-blocking socket_open)

Javivi:
Both ways adhere to the HTTP requisite of a blank line at the end, but yes, if the test plugin doesn't work for someone try changing ^n^n for ^r^n.

Do you mean creating a new function that spawns a thread and then callbacks when the socket is "ready" ?

We discussed at the beginning of the pull against having 2 separate functions for connect when just 1 would do, and also keep in mind that once getaddrinfo resolves an address it stays on the dns cache until it gets flushed, and that may never happen.

Using a numeric IP or having the dns address already cached makes the socket_open function with the NON_BLOCKING flag not to block, adding a new function with threads just for doing a resolve adds some unneeded complexity in my opinion.

WPMGPRoSToTeMa:
@Javivi hm, you can just add function for domain resolving. But what about non-blocking TCP connect?

Javivi:
You mean doing something like:
1 - socket_dnspreload("domain.com")
2 - spawn thread and call getaddrinfo
3 - callback

That can be done, but I don't think it is needed. You only need to resolve the hostname one time and it stays cached until the dns cache gets flushed.

Calling socket_open just one time at plugin_init or at any event that doesn't interfere with a current game would do the same, even though it is not as pretty

Unless the dns of the domain is really misconfigured the query should not take more than a few ms and it can even return instantly, I would like some feedback on this if possible too https://github.com/javivi/amxmodx/tree/sockets-update


Having an event based module is obviously the end game, and there won't be any problems with this when that happens, but meanwhile a work around is needed.

Arkshine:
The changes look overall okay.

- Code could have modernized a little more with C++11/AMTL but it's okay as it is.
- I'm not totally sure about `socket_is_readable` as native. If just an alias of `socket_change`, a stock would be enough I think, but it's more consistent and readable this way and if following this way, you should probably deprecate `socket_change` similar to https://github.com/alliedmodders/sourcemod/blob/master/plugins/include/string.inc#L92-L100.
- Documentation needs likely some adjustments. Comment on constants should use  /** */ (i.e: https://github.com/alliedmodders/amxmodx/blob/master/plugins/include/amxconst.inc). English is not my forte. 'Check should be `Checks`. You should probably cut the lines (~80/120 characters?). Klippy did some update in another PR https://github.com/rsKliPPy/amxmodx/blob/2165ffb4b5ef84cc84e03744e41b831627a272bf/plugins/include/sockets.inc, you could probably take some inspiration from there and mixing things.

Are there others things you wanted to add?

Javivi:
Should be ready for merge if nothing wrong pops.

Spaguetti test code.

/s1 uses socket_open() + socket_send()
/s2 uses socket_open() with the new nonblocking and libc errors flags + socket_send() + socket_is_writable(). Messing around with the NODE/SERVICE on socket_open() will show some of the new libc errors
/s3 uses socket_open() + socket_send2() with a null byte
/s4 uses socket_open() + socket_send() with a null byte

**/s1** and **/s2** should print something like this on console

>HTTP/1.1 200 OK
>Server: nginx
>Date: Wed, 22 Feb 2017 23:47:01 GMT
>Content-Type: application/json
>Content-Length: 155
>Connection: close
>Access-Control-Allow-Origin: *
>Access-Control-Allow-Credentials: true
>
>{
>  "args": {}, 
>  "headers": {
>    "Ho
>st": "httpbin.org", 
>    "Test": "a\u00ffc"
> }, 
>  "origin": "1.2.3.4", 
>  "url": "http://httpbin.org/get"
>}

**/s3** should print

>HTTP/1.1 400 Bad Request
>Server: nginx
>Date: Wed, 22 Feb 2017 23:54:25 GMT
>Content-Type: text/html
>Content-Length: 166
>Connection: close
>
><html>
><head><title>400 Bad Request</title></head>
><body bgcolor="white">
><center><h1>400 Bad Request</h1>
></center>
><hr><center>nginx</center>
></body>
></html>

**/s4** should crash the server

```SourcePawn
#include <amxmodx>
#include <sockets>

#define NODE "httpbin.org"
#define SERVICE 80

public plugin_init( )
{
    register_clcmd("say /s1", "test_blocking")
    register_clcmd("say /s2", "test_nonblocking")
    register_clcmd("say /s3", "test_send2")
    register_clcmd("say /s4", "test_send12")
}

public test_blocking(id)
{
    new sockfd, socket_error
    new bytes_sent, bytes_recv, bytes_recv_buffer[500], bytes_recv_buffer_printed

    sockfd = socket_open(NODE, SERVICE, SOCKET_TCP, socket_error, SOCK_LIBC_ERRORS)

    if(sockfd != -1)
    {
        client_print(id, print_chat, "Got socketfd %d", sockfd)

        new request[100]
        formatex(request, charsmax(request), "GET /get HTTP/1.1^nHOST: httpbin.org^nTEST: %c%c%c^nConnection: close^n^n^n^n", 'a', -1, 'c')
        
        bytes_sent = socket_send(sockfd, request, strlen(request))
        client_print(id, print_chat, "[%d/%d] bytes sent", bytes_sent, charsmax(request))

        bytes_recv = socket_recv(sockfd, bytes_recv_buffer, charsmax(bytes_recv_buffer))
        client_print(id, print_chat, "[%d/%d] bytes received", bytes_recv, charsmax(bytes_recv_buffer))

        bytes_recv_buffer_printed = server_print("%s", bytes_recv_buffer)
        server_print("%s", bytes_recv_buffer[bytes_recv_buffer_printed-1])

        socket_close(sockfd)
    }
    else
    {
        client_print(id, print_chat, "Error %d", socket_error)
    }
}

new nb_sockfd

public test_nonblocking(id)
{
    new socket_error

    nb_sockfd = socket_open(NODE, SERVICE, SOCKET_TCP, socket_error, SOCK_NON_BLOCKING|SOCK_LIBC_ERRORS)

    if(nb_sockfd != -1)
    {
        client_print(id, print_chat, "Got nonblocking socketfd %d", nb_sockfd)

        set_task(1.0, "check_nbsocket", 1234, .flags="b")
    }
    else
    {
        client_print(id, print_chat, "Error %d", socket_error)
    }
}

public check_nbsocket(task)
{
    if(socket_is_writable(nb_sockfd))
    {
        remove_task(task)

        client_print(0, print_chat, "Socket %d is writable", nb_sockfd)

        new request[100]
        formatex(request, charsmax(request), "GET /get HTTP/1.1^nHOST: httpbin.org^nTEST: %c%c%c^nConnection: close^n^n^n^n", 'a', 'b', 'c')

        new bytes_sent = socket_send(nb_sockfd, request, strlen(request))
        client_print(0, print_chat, "[%d/%d] bytes sent", bytes_sent, charsmax(request))

        set_task(1.0, "recv_nbsocket", 1234, .flags = "b")
    }
}

public recv_nbsocket(task)
{
    new bytes_recv, bytes_recv_buffer[500], bytes_recv_buffer_printed

    bytes_recv = socket_recv(nb_sockfd, bytes_recv_buffer, charsmax(bytes_recv_buffer))

    if(bytes_recv)
    {
        remove_task(task)

        client_print(0, print_chat, "[%d/%d] bytes received", bytes_recv, charsmax(bytes_recv_buffer))

        bytes_recv_buffer_printed = server_print("%s", bytes_recv_buffer)
        server_print("%s", bytes_recv_buffer[bytes_recv_buffer_printed-1])

        socket_close(nb_sockfd)
    }
}

public test_send2(id)
{
    new sockfd, socket_error
    new bytes_sent, bytes_recv, bytes_recv_buffer[500], bytes_recv_buffer_printed

    sockfd = socket_open(NODE, SERVICE, SOCKET_TCP, socket_error, SOCK_LIBC_ERRORS)

    if(sockfd != -1)
    {
        client_print(id, print_chat, "Got socketfd %d", sockfd)

        new request[100]
        formatex(request, charsmax(request), "GET /get HTTP/1.1^nHOST: httpbin.org^nTEST: %c%c%c^nConnection: close^n^n^n^n", 'a', 0, 'c')
        
        bytes_sent = socket_send2(sockfd, request, charsmax(request))
        client_print(id, print_chat, "[%d/%d] bytes sent", bytes_sent, charsmax(request))

        bytes_recv = socket_recv(sockfd, bytes_recv_buffer, charsmax(bytes_recv_buffer))
        client_print(id, print_chat, "[%d/%d] bytes received", bytes_recv, charsmax(bytes_recv_buffer))

        bytes_recv_buffer_printed = server_print("%s", bytes_recv_buffer)
        server_print("%s", bytes_recv_buffer[bytes_recv_buffer_printed-1])

        socket_close(sockfd)
    }
    else
    {
        client_print(id, print_chat, "Error %d", socket_error)
    }
}

public test_send12(id)
{
    new sockfd, socket_error
    new bytes_sent, bytes_recv, bytes_recv_buffer[500], bytes_recv_buffer_printed

    sockfd = socket_open(NODE, SERVICE, SOCKET_TCP, socket_error, SOCK_LIBC_ERRORS)

    if(sockfd != -1)
    {
        client_print(id, print_chat, "Got socketfd %d", sockfd)

        new request[100]
        formatex(request, charsmax(request), "GET /get HTTP/1.1^nHOST: httpbin.org^nTEST: %c%c%c^nConnection: close^n^n^n^n", 'a', 0, 'c')
        
        bytes_sent = socket_send(sockfd, request, charsmax(request))
        client_print(id, print_chat, "[%d/%d] bytes sent", bytes_sent, charsmax(request))

        bytes_recv = socket_recv(sockfd, bytes_recv_buffer, charsmax(bytes_recv_buffer))
        client_print(id, print_chat, "[%d/%d] bytes received", bytes_recv, charsmax(bytes_recv_buffer))

        bytes_recv_buffer_printed = server_print("%s", bytes_recv_buffer)
        server_print("%s", bytes_recv_buffer[bytes_recv_buffer_printed-1])

        socket_close(sockfd)
    }
    else
    {
        client_print(id, print_chat, "Error %d", socket_error)
    }
}
```

PR #306, opened 2015-11-06
Allow wordspawn in edict/entvars->index type conversion  (bug 6429)

IgnacioFDM:
FNullEnt is too strict, and causes worldspawn to fail these checks.

Related to #297.

### DIFF
--- a/public/HLTypeConversion.h
+++ b/public/HLTypeConversion.h
@@ -77,7 +77,7 @@ class HLTypeConversion
 
 		int edict_to_id(edict_t *pEdict)
 		{
-			if (FNullEnt(pEdict))
+			if (!pEdict)
 			{
 				return -1;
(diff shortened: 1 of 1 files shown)


Arkshine:
You're right. It looks like I've mistaken the purpose of `FNullEnt`, which should be used when you want to ignore worldspawn.

I think you can remove the check in EHANDLE class below, as well. 

And you would make me happy if you could squash your commits a the end.

IgnacioFDM:
There it is. Had to get into a little fight with git. Not sure about that AppVeyor error.

Arkshine:
It happens sometimes when no changes have been detected or something.It can be ignored. Probably some small issue in AMBuild.

Anyway, looks good now. Changes should be safe. Thanks.

PR #311, opened 2015-11-26
update admin.sma

9iky6:
add stock ShowLoadAdminsMessage(const client, const count, bool:mysql = false) and remove client_infochanged(id) (here there was a conflict with a plugin which changes the names of the players, for example, a kill assist) and remove AdminCount and bool:g_CaseSensitiveName[MAX_PLAYERS + 1]

### DIFF
--- a/plugins/admin.sma
+++ b/plugins/admin.sma
@@ -22,17 +22,13 @@
 
 //new Vector:AdminList;
 
-new AdminCount;
-
 new PLUGINNAME[] = "AMX Mod X"
 
 #define ADMIN_LOOKUP	(1<<0)
 #define ADMIN_NORMAL	(1<<1)
-#define ADMIN_STEAM		(1<<2)
+#define ADMIN_STEAM	(1<<2)
 #define ADMIN_IPADDR	(1<<3)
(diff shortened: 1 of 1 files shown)


Arkshine:
You can't just remove `client_infochanged`. It's used to update access on player's name change. Removing this behavior is not really something we would want to see.

9iky6:
Remove client_infochanged need for some plugins that replace the players names. For example: kill_assist.sma
https://forums.alliedmods.net/showthread.php?p=817621

It produces 2 times the change of the name of the player, but client_infochanged is triggered only 1 time and the player is not admin return the access flags

Arkshine:
Yes, I understand, but this is anyway not a good reason to remove the forward. You have to keep backward compatibility in mind and breaking old behavior is not a good solution.

I want to update things around Admin at some point, stealing some ideas from SourceMod. Having a forward to have control whether you want to block admin check could be a solution. Something to consider later.

Anyways for now, if you still want to maintain this PR, you will have to revert this change.

Arkshine:
You are not going to like it, but could you merge your two commits into one, please?

PR #313, opened 2015-12-02
Fix translations for %l specifier

Ni3znajomy:
Related to #229.

This bug occurs when we use ```%l``` format specifier. It sets server's language when it is used instead of player's.
The problem lies in the __translate__ function. When we use ```%l``` specifier it passes player's or server's index as amxaddr (but it's not amxaddr really) to the function so if we call __get_amxaddr__ for player/server index it'll return 0 (set server's language).

### DIFF
--- a/amxmodx/format.cpp
+++ b/amxmodx/format.cpp
@@ -52,7 +52,13 @@ const char *translate(AMX *amx, cell amxaddr, const char *key)
 	const char *pLangName = NULL;
 	const char *def = NULL;
 	int status;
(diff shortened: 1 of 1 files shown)


Nextra:
Good catch, no idea how I missed that. The fix is a bit too hacky for my taste, though. We should come up with something cleaner.

Ni3znajomy:
I agree that the solution isn't the best.
My first solution was to add new parameter to translate function (bool type) and set it to false when it is amxaddr or true when we pass player/server index directly to the function. What about that?

PR #314, opened 2015-12-02
Fix strncmp case sensitivity

Th3-822:
strncmp was using case insensitive call when ignorecase == false


Arkshine:
Ah, damn. Thanks.

Related to a86ca1491f97961cd58ac1492272130742bb5d5b.

PR #315, opened 2015-12-05
Make sure client is alive before checking buying commands

Arkshine:
This might also prevent a potential crash when client's private data is checked right before he enters in game.


Nextra:
:+1:

PR #317, opened 2015-12-09
Hamsandwich: fix HAM_OVERRIDE in post hooks (bug 6443)

WPMGPRoSToTeMa:


Arkshine:
Looking at your message https://bugs.alliedmods.net/show_bug.cgi?id=6443 ; it's expected first example doesn't work. You can't know original value in pre, nor overriding is supposed to be used at this point. Though you can change return value and superceding, it will work.

That's said, documentation says well about `HAM_OVERRIDE` : 
> Still calls the target function, but returns whatever is set with SetHamReturn

Right now, this is indeed missing. I'm not sure if there will be a lot of usage where you would need to know original value before changing it, but it's okay to have this possibility either way.

Thanks for the patch.

WPMGPRoSToTeMa:
@Arkshine I have post hook in first example, not pre hook.
Metamod post hooks are able to change return value via MRES_OVERRIDE (but MRES_SUPERCEDE not valid in post functions).
Also I think that it will be useful to add this feature to fakemeta module, isn't it?

WPMGPRoSToTeMa:
>**MRES_OVERRIDE**
>The plugin is providing a return value for the routine, which should be used in place of the return value from the gameDLL's routine (the plugin "overrides" the gameDLL's return value). Unless otherwise specified by a later plugin, the gameDLL routine will still be called. Note this only makes sense for non-void routines. This is valid in both normal and post routines.
> - http://metamod.org/coding.html

Arkshine:
I remember reading a comment somewhere that override doesn't work well with metamod. But I can't say more than that. That would be need some investigation.

PR #318, opened 2015-12-11
Enable hook detour on cvar with bounds regardless current value

Arkshine:
Reported by PartialCloning here : https://forums.alliedmods.net/showthread.php?t=275918

Related to #185.


Nextra:
:gem:

PR #319, opened 2015-12-13
Enhance "amxx" client-side command

Ni3znajomy:
The idea was stolen from SM. SM allows us (as a player) to check what plugins and extensions are running on a server. I think it would be nice to have similar feature in amxx. Also added missing core mode "JIT+ASM32".

### DIFF
--- a/amxmodx/meta_api.cpp
+++ b/amxmodx/meta_api.cpp
@@ -958,11 +958,48 @@ void C_ClientCommand(edict_t *pEntity)
 	// Handle "amxx" if not on listenserver
 	if (IS_DEDICATED_SERVER())
 	{
+		static char buf[1024];
+		size_t len = 0;
+
 		if (cmd && stricmp(cmd, "amxx") == 0)
 		{
+		    int argc = CMD_ARGC();
+		    if(argc > 1 && stricmp(arg, "plugins") == 0)
+		    {
+				if(!g_plugins.getPluginsNum())
+				{
+					CLIENT_PRINT(pEntity, print_console, "[AMXX] No plugins found.");
+					RETURN_META(MRES_SUPERCEDE);
+				}
+
+				CPluginMngr::iterator a = g_plugins.begin();
+				size_t i = 0;
+
+				while(a)
+				{
+					if((*a).getStatusCode() == ps_running)
+					{
+						len = ke::SafeSprintf(buf, sizeof(buf), "[%d] \"%s\"", ++i, (*a).getTitle());
+
+						if(*(*a).getVersion())
+						{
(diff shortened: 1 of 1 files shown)


Arkshine:
Is it really useful as simple user to know such informations?

Ni3znajomy:
For simple user, hmm... probably not, but not everyone is simple user. For someone who want to know, what plugin(s) is/are running on a server, it's easier to type "amxx plugins/modules" than write to server's operator to ask and wait, we don't know that he will answer (give a plugin's name or tell "search on your own") or even not answer. I've proposed this to make life easier for people who aren't as you said simple users :smiley: .

ClaudiuHKS:
I think it is, Arkshine. These builds are free, and everyone can have/ use them.
The published AMXX scripts are free as well - see http://www.amxmodx.org/compiler.php

amxmodx/meta_api.cpp
```diff
@@ -958,29 +958,97 @@ void C_ClientCommand(edict_t *pEntity)
...
+						{
+							len += ke::SafeSprintf(&buf[len], sizeof(buf)-len, " (%s)", (*a).getVersion());
+						}
+						if(*(*a).getAuthor())
+						{
+							len += ke::SafeSprintf(&buf[len], sizeof(buf)-len, " by %s", (*a).getAuthor());
+						}
+					
+						len += ke::SafeSprintf(&buf[len], sizeof(buf)-len, " %s\n", (*a).getName());
+                        CLIENT_PRINT(pEntity, print_console, buf);
```
WildCard65: I think formatting is not correct for this line.

amxmodx/meta_api.cpp
```diff
@@ -958,29 +958,97 @@ void C_ClientCommand(edict_t *pEntity)
...
+					}
+                    ++a;
```
WildCard65: Ditto.

WildCard65:
Looks like you used spaces instead of tabs in your code edits.

Ni3znajomy:
@WildCard65 You're right, I missed that, now it should be ok.

IgnacioFDM:
I like the idea but only if there's a cvar to disable it.

WildCard65:
@IgnacioFDM This PR is based around the very same feature SM has which SM doesn't have a cvar to disable it.

IgnacioFDM:
So what if SM does it that way? If there is no trivial way to disable it, I find this feature invasive.

Arkshine:
I don't mind the idea but I would agree admin should be able to control it.
But then you have already commands for that:  `amx_plugins` and  `amx_modules, in `admincmd.sma`.
Just a matter to change the commands access level.

Arkshine:
Closing this then. If you need to discuss further about this, feel free to do it.

PR #320, opened 2015-12-13
Fix incorrect translation language for %l format identifier

Arkshine:
Original report: #313
Related to: #229


Ni3znajomy:
It's much better solution than mine. :+1:

Arkshine:
Thanks for reporting by the way!

Nextra:
:christmas_tree:

PR #321, opened 2015-12-13
Fix uncolor print

9iky6:
Add 0 byte default color (yelow) to show color message.

### DIFF
--- a/amxmodx/amxmodx.cpp
+++ b/amxmodx/amxmodx.cpp
@@ -336,9 +336,11 @@ static cell AMX_NATIVE_CALL client_print_color(AMX *amx, cell *params) /* 3 para
 			{
 				g_langMngr.SetDefLang(i);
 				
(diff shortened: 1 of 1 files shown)


Nextra:
Your code overwrites plugin data instead of actually inserting the 0x01, this is a bug.

9iky6:
Unfortunately I don't know how to do differently, but in *.the sma plug-ins necessary to use this option, and nothing terrible happens (no data replacement). (sorry for the translator)

Nextra:
That is because they then have the format function write from msg[1] onwards, keeping the msg[0] intact. We'll think about how to solve this, thanks for the suggestion.

Arkshine:
Thanks. I've reposted in #338.

@9iky6 You should really use a tool like SmartGIT, so you can work locally and amending/merging commits easily.

PR #322, opened 2015-12-15
New client_connectex forward & extend client_authorized

Ni3znajomy:
It's extended version of client_connect (as suffix "ex" says :smile:)

#### **client_connectex**
It allows to reject player's attempt to join to a server.
Forward passes some extra parameters like player's name & ip also gives ability to change reason why player gets rejected.

#### **client_authorized**
It has been extended to also pass client steamid.

### DIFF
--- a/amxmodx/amxmodx.h
+++ b/amxmodx/amxmodx.h
@@ -339,6 +339,7 @@ extern int FF_PluginEnd;
 extern int FF_InconsistentFile;
 extern int FF_ClientAuthorized;
 extern int FF_ChangeLevel;
+extern int FF_ClientConnectEx;
 
 extern bool g_coloredmenus;
 
--- a/amxmodx/meta_api.cpp
+++ b/amxmodx/meta_api.cpp
@@ -145,6 +145,7 @@ int FF_PluginEnd = -1;
 int FF_InconsistentFile = -1;
 int FF_ClientAuthorized = -1;
 int FF_ChangeLevel = -1;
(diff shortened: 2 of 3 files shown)


Arkshine:
You should avoid the use of `FP_STRINGEX`, it's a bad design. Use  `FP_ARRAY` instead like it has been done for `client_disconnected`.

luxxxoor:
why you not pass the steamid too in client_authorizedex ? i mean this is the main use for that forward

Ni3znajomy:
Sorry for this little mess with references but i had some mistakes in previous ones.
@Arkshine done.
@luxxxoor I didn't think about this that time. I agree with you that forward should pass steamid too, so now it passes.

luxxxoor:
it's right to add new forwards only to client _authorized/connect ? 
i mean if client_putinserver/disconnected won't have this params it will look wierd.

Arkshine:
Yes. Having steamid is a given, but name and ip address, I'm not sure.This could be useful but If you put them there, then for consistency they should be in the others forwards too. That they are in _connect makes sense since it's the first user's connection. Ip address is not often used and name is versatile. Well, I don't think you should put them in _authorized.

By the way, about authorized, it should be safe to put new argument there. No need new forward. And it looks like you missed two occurrences in the code.

Ni3znajomy:
I removed client_authorizedex, now client_authorized passes client id and authid.

Arkshine:
Looks good now. Thanks.

PR #323, opened 2015-12-19
Adding flag parameter to user_silentkill.

luxxxoor:
The api says that the user_silentkill() "Kills a user without a message.",  that means it is just user_kill() without sending deathmsg.
But user_silentkill() automatically kills player without affecting the client's score (and that is not mentioned in api), so i'll add a second parameter to user_silnetkill() to let user to decide if he want to kill a player without or not affecting the table.
New parameter has the default value 1, so it won't break the compatibility.


Arkshine:
Fair enough. Documentation can wait Nextra. Thanks.

PR #324, opened 2015-12-22
Update SQLite

ClaudiuHKS:
Update SQLite so it has the same options as SourceMod's.
- It's thread-safe.
- It supports update/ delete limitation. **DELETE FROM `tabName` WHERE `colName` = `Val` LIMIT `%`;**

### DIFF
--- a/modules/sqlite/AMBuilder
+++ b/modules/sqlite/AMBuilder
@@ -10,6 +10,9 @@ binary.compiler.cxxincludes += [
 binary.compiler.defines += [
   'SM_DEFAULT_THREADER',
   'HAVE_STDINT_H',
+  'SQLITE_OMIT_LOAD_EXTENSION',
+  'SQLITE_THREADSAFE',
+  'SQLITE_ENABLE_UPDATE_DELETE_LIMIT',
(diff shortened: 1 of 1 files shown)


GuskiS:
Care for adding/making posgresql module?

IgnacioFDM:
postgres would be amazing indeed.

Arkshine:
I've followed discussion on SM PR, while I agree with psychonic that your reason to update is arguable from your example and about SQL standard too, I do agree with you though that If MySQL supports it already, It would make sense and I don't see any harm in allowing it in SQLite. Unless there is some drawbacks by enable it, I would be okay with this. Did you test it?

Arkshine:
@ClaudiuHKS ?

ClaudiuHKS:
Tested and working properly. Sorry for the answer delay.

Arkshine:
@ClaudiuHKS `SQLITE_THREADSAFE` doesn't seem to be needed because it's on 1 by default, but I guess it's okay to explicitly define it. You would make me happy if you could update sqlite VS project with those defines just for consistency.

PR #328, opened 2016-01-01
SELinux compatibility: memalign -> mmap

WPMGPRoSToTeMa:


modules/hamsandwich/hook.h
```diff
@@ -82,7 +82,9 @@ class Hook
 		ivtable[entry]=(int *)func;
 #if defined(_WIN32)
 		VirtualFree(tramp, 0, MEM_RELEASE);
-#elif defined(__linux__) || defined(__APPLE__)
+#elif defined(__linux__)
+		munmap(tramp, sysconf(_SC_PAGESIZE));
```
Arkshine: I might be wrong, but shouldn't you pass the same size as `m_size` value? You will likely need to do something to get such length.

public/memtools/CDetour/detourhelpers.h
```diff
@@ -89,6 +89,8 @@ inline void FreePageMemory(void *addr)
 {
 #if defined(WIN32)
 	VirtualFree(addr, 0, MEM_RELEASE);
+#elif defined(__linux__)
+	munmap(addr, sysconf(_SC_PAGESIZE));
```
Arkshine: Same here.

modules/hamsandwich/hook.h
```diff
@@ -37,9 +37,10 @@ class Hook
 	int              del;     // 1 if this hook should be destroyed after exec
 	void            *tramp;   // trampoline for this hook
 	char			*ent;     // ent name that's being hooked
+	int              trampSize;
```
Nextra: This should also be `size_t`

Arkshine:
Thanks for the changes. This looks technically correct now.

PR #329, opened 2016-01-03
Fix broken UTF8 support.

IgnacioFDM:
UTF8 detection was broken. Main sma would never get detected as "utf8", and therefore work correctly. Included files would get detected as utf8, and give incorrect results ingame.

A simple demostration is the following (make sure both files are UTF8 without BOM):

test1.sma

``` c

#include "test2.sma"
public plugin_init()
{
    register_clcmd("tryme", "tryme");
}

public tryme(id)
{
    client_print(0, print_chat, "asd ñ á é í asd");
    test();
}
```

test2.sma

``` c
test()
{
    client_print(0, print_chat, "asd ñ á é í asd");
}
```

Ingame, the first message will show correclty, and the second message incorrectly.


WPMGPRoSToTeMa:
The second message is encoded using UTF-32 if I'm not mistaken.

compiler/libpc300/sc.h
```diff
@@ -722,7 +722,6 @@ SC_FUNC int cp_path(const char *root,const char *directory);
 SC_FUNC int cp_set(const char *name);
 SC_FUNC cell cp_translate(const unsigned char *string,const unsigned char **endptr);
 SC_FUNC cell get_utf8_char(const unsigned char *string,const unsigned char **endptr);
```
reuben: Looks like you forgot to remove this declaration.

IgnacioFDM: You're right

IgnacioFDM:
There

rsKliPPy:
Looking at the code, what happens if you define NO_UTF8 in compiler options? Wouldn't just that work?

IgnacioFDM:
To be honest this PR is so old that I don't remember how the code worked (however I've been using the compiler with this commit cherrypicked ever since and fixes the issue and works fine)

I think I removed all that code instead of using that define because there was no point in leaving broken code.

wopox1337:
What about this merge? so useful when develop "not 1-file plugin"

Arkshine:
I tried to look again at this to understand why the compiler converts to UTF-32 nor why it doesn't process the main file.

The Pawn Language Guide provide some answer:

>The pawn language requires that its keywords and symbols names are in
ascii, and it allows non-ascii characters in strings. There are five ways
that a host application could support non-ascii characters in strings and
character literals:
>  1. Support codepages: in this strategy the entire complexity of choosing
the correct glyphs and fonts is delegated to the host application. The
codepage support is based on codepage mapping files with a file format
of the “cross mapping tables” distributed by the Unicode consortium.
> 2. Support Unicode or UCS-4 and let the pawn compiler convert scripts that
were written using a codepage to “wide” characters: for this strategy, you
need to set #pragma codepage or use the equivalent compiler option. The
compiler will only correctly translate characters in unpacked strings.
> 3. Support Unicode or UCS-4 and let the pawn compiler convert scripts
encoded in UTF-8 to “wide” characters: when the source file for the
pawn compiler is in UTF-8 encoding, the compiler expands characters
to Unicode/UCS-4 in unpacked strings.
> 4. Support UTF-8 encoding internally (in the host application) and write
the source file in UTF-8 too: all strings should now be packed strings to
avoid the compiler to convert them.

> The pawn compiler accepts a source file as an UTF-8 encoded text file —see
page 169. When the source file is in UTF-8 encoding, “wide” characters in
an unpacked string are stored as multi-byte Unicode/UCS-4 characters; wide
characters in a packed string remain in UTF-8 encoding. To write source
files in UTF-8 encoding, you need, of course, a (programmer’s) editor that
supports UTF-8. Codepage translation does not apply for files that are in
UTF-8 encoding.

Basically, what does the compiler is expected, but the issue specific to us is 4. You are supposed to use packed strings if your file is in UTF-8. We don't use packed strings in AMXX.

In our context, this is not interesting nor helpful to get strings converted in "wide" characters.
I guess it's fine to disable it and let people to use the UTF-8 encoding in the concerned source files.

PR #330, opened 2016-01-06
Add cs_get_user_weapon_entity() and cs_get_user_weapon() natives

WPMGPRoSToTeMa:


Arkshine:
Getting current weapon is often used. Having CS version is welcomed for the sake of convenience/readability/reliability, I guess. Code looks okay, thanks!

PR #331, opened 2016-01-07
Fix an out of bound buffer index

Arkshine:
Silly typo. Reported by IgnacioFDM. Related to #320.


IgnacioFDM:
After a few days of testing, this seem's okay.

Nextra:
Went through the code and looked at the `atcprintf`uses, and it seems like this should be safe. Hope the testing was enough. :ship:

PR #334, opened 2016-01-11
Enable detours in ServerActivate to avoid sig conflict with third-party AMXX modules

Arkshine:
This gives a chance for third-party modules which use `OnPluginsLoaded` to sigscan game library , before AMXX patches memory (when detour is enabled).

Requested by someone, somewhere.

### DIFF
--- a/modules/cstrike/cstrike/CstrikeMain.cpp
+++ b/modules/cstrike/cstrike/CstrikeMain.cpp
@@ -67,13 +67,19 @@ void OnPluginsLoaded()
 	ForwardInternalCommand = MF_RegisterForward("CS_InternalCommand", ET_STOP, FP_CELL, FP_STRING, FP_DONE);
 	ForwardOnBuy           = MF_RegisterForward("CS_OnBuy"          , ET_STOP, FP_CELL, FP_CELL, FP_DONE);
 	ForwardOnBuyAttempt    = MF_RegisterForward("CS_OnBuyAttempt"   , ET_STOP, FP_CELL, FP_CELL, FP_DONE);
+}
 
-	if (!ClientCommandDetour) // All CS_* forwards requires ClientCommand. Unlikely to fail.
+void OnServerActivate(edict_t *pEdictList, int edictCount, int clientMax)
+{
+	// Used to catch WeaponList message at map change.
+	EnableMessageHooks();
+
+	if (!ClientCommandDetour) // All CS_* forwards requires ClientCommand. Unlikely to fail. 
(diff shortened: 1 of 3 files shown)


xPaw:
So what happens if third party modules patch memory before AMXX does?

modules/cstrike/cstrike/CstrikeMain.cpp
```diff
@@ -67,13 +67,19 @@ void OnPluginsLoaded()
...
 	ForwardOnBuyAttempt    = MF_RegisterForward("CS_OnBuyAttempt"   , ET_STOP, FP_CELL, FP_CELL, FP_DONE);
+}
+
+void OnServerActivate(edict_t *pEdictList, int edictCount, int clientMax)
+{
+	// Used to catch WeaponList message at map change.
+	EnableMessageHooks();
 
-	if (!ClientCommandDetour) // All CS_* forwards requires ClientCommand. Unlikely to fail.
+	if (!ClientCommandDetour) // All CS_* forwards requires ClientCommand. Unlikely to fail.
```
xPaw: Nice space! :scream:

Arkshine: lol, It was a silly change to refresh build on Appveyor since for some reason AMBuild fails randomly.

Arkshine:
AMXX would patch on it without issue (chain is preserved) since addresses are still retrieved one time at server start, but memory patches would actually happen later.

Arkshine:
@Nextra , please.

Nextra:
So just for understanding the new load-chain is as follows:

1. AMXX Startup
2. AMXX Sigscan
3. Module Startup
4. Module Sigscan
5. Module Patch
6. AMXX Patch

For modules that this fix targets: Is this process properly reversed on deactivation so that those modules remove their patches before AMXX does? Does it cause issues that module detours now chain after AMXX detours (as they are patched first)?

Arkshine:
The list is correct.

About deactivation, as long module use AMXX API `OnPluginLoaded/Unloaded`, this should be fine. 

Whatever AMXX/Cstrike, patching occurs in `ServerActivate`, so _after_ `OnPluginLoaded`. The third-party module will patch first.
Unpatching occurs in `ServerDeactivate`, so _before_ `OnPluginUnloaded/ing` is called, AMXX will unpatch first, then third-party module next.

Does it make sense what I'm saying?

Nextra:
Alright so this should improve things. :icecream:

PR #335, opened 2016-01-14
Fix recursion issue with show_menu and newmenus

Arkshine:
Reported by Depresie here: https://forums.alliedmods.net/showthread.php?t=277625.

Side-effect of #140 bugfix. Now that `show_menu` closes previous newmenu, it doesn't consider that calling callback could display another newmenus. This patch applies the same infinite loop protection than `menu_display`.

### DIFF
--- a/amxmodx/amxmodx.cpp
+++ b/amxmodx/amxmodx.cpp
@@ -1251,8 +1251,11 @@ static cell AMX_NATIVE_CALL show_menu(AMX *amx, cell *params) /* 3 param */
 				pPlayer->menu = 0;
 
 				// Fire newmenu callback so closing it can be handled by the plugin
-				if (Menu *pMenu = get_menu_by_id(pPlayer->newmenu))
-					pMenu->Close(pPlayer->index);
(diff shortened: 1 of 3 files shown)


amxmodx/amxmodx.cpp
```diff
@@ -1251,8 +1251,11 @@ static cell AMX_NATIVE_CALL show_menu(AMX *amx, cell *params) /* 3 param */
 				pPlayer->menu = 0;
 
 				// Fire newmenu callback so closing it can be handled by the plugin
-				if (Menu *pMenu = get_menu_by_id(pPlayer->newmenu))
-					pMenu->Close(pPlayer->index);
+				if (!pPlayer->CloseNewMenus())
+				{
+					LogError(amx, AMX_ERR_NATIVE, "Plugin called menu_display when item=MENU_EXIT");
```
Nextra: This is not `menu_display()`

Nextra:
I think the new function use `get_menu_by_id()`, it makes for much cleaner code.

Nextra:
:ship:

PR #336, opened 2016-01-15
Update GameConfig

Arkshine:
* Allows single config file without master. Really for convenience. Master has been introduced for better maintainability overall (AMBuild, Updater, etc), but for simple config this is not really useful. I think it's fine to have this choice depending the associated config(s) to a plugin. To be discussed. 
* Fixes `LoadGameConfigFile` native which will return a valid handle if native is called two times ormore with an invalid path. It's because path is always saved in cache. A possible solution is to close gameconfig right away. Internal design is kind of weird.
* Prints a message to console when there is a gamedata file loaded from the `custom` folder to act as a reminder that original value is overwritten. Should we save in amxx log instead?

### DIFF
--- a/amxmodx/CGameConfigs.cpp
+++ b/amxmodx/CGameConfigs.cpp
@@ -651,7 +651,15 @@ bool CGameConfig::Reparse(char *error, size_t maxlength)
 		if (g_LibSys.PathExists(path))
 		{
 			g_LibSys.PathFormat(path, sizeof(path), "custom/%s.txt", m_File);
-			return EnterFile(path, error, maxlength);
+
+			auto success = EnterFile(path, error, maxlength);
+
+			if (success)
+			{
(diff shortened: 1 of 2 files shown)


Arkshine:
@Nextra , please.

Nextra:
- So this allows games to supply their own gamedata file without a master file? Wouldn't games manually trigger the parse anyway or was this impossible before?
- Seems good
- I don't know about this. If you want to log this it should probably be in the AMXX log. Is it triggered every map change or only on startup?

Arkshine:
- It was impossible before. You had to create a directory, put your files, and adding a master to tell parser to read those files. 
- About message, it should concern only default gamedata which are read one time.

Nextra:
- Good
- Okay, if you want to log it put it in the AMXX logs where the other gamedata info goes.

:gem:

PR #337, opened 2016-01-15
Move Cvar_DirectSet to gamedata

Ni3znajomy:
Binding/Hooking cvars don't work on ReHLDS server without modification AMXX code. I think the best option is to move Cvar_DirectSet to gamedata files for easier configuration. I'm not sure about windows signature but when I was testing it, it worked as it should. I'll appreciate if anyone checks it just for sanity. Tested on Linux & Windows.

### DIFF
--- a/amxmodx/CvarManager.cpp
+++ b/amxmodx/CvarManager.cpp
@@ -116,41 +116,16 @@ void CvarManager::CreateCvarHook(void)
 	//   	Cvar_DirectSet(var, value); // <- We want to hook this.
 	// }
 
-	byte *baseAddress = (byte *)g_engfuncs.pfnCvar_DirectSet;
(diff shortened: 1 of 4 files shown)


Arkshine:
Since we have gamedata now, I don't mind removing this kind of code in favor of signatures. Will check windows signature later.

Arkshine:
I would use `\x55\x8B\x2A\x81\x2A\x2A\x2A\x2A\x2A\x56\x8B\x2A\x2A\x57\x8B\x2A\x2A\x85` ; I don't know much assembly but I prefer to take only the first byte of each instructions, just to be safe.

Ni3znajomy:
Done.

Arkshine:
Should be okay. Thanks.

PR #338, opened 2016-01-16
Improve client_print_color by inserting default color code at the start of string

Arkshine:
Originally reported by @9iky6 in #321.

To display colored message, `SayText` requires that at first index a color code is present.
Native will now automatically insert the default one if not present. For the sake of convenience.


Nextra:
:gem:

PR #339, opened 2016-01-17
Normalize line endings and whitespace

IgnacioFDM:
Since .gitattribute * text=auto, git will automatically convert these 34 or so files right after cloning. This PR only commits the auto modified by git files, so when you clone amxmodx you don't automatically get a lot of "modofied" files right off the bat.


Arkshine:
Just curious, how do you clone, in what environnement? Because I can't seem to reproduce this behavior on cloning. Maybe I've git configured.

IgnacioFDM:
Ok, so just to be sure, I retried this in a fresh VM with Debian 8. I simply installed git (apt-get install git) and git clone'd without touching anything else. Immediately after git cloning, if I git status, I get all these files I fixed in this commit as "modified"

This thing happens to me in Windows too, so it's not just some linux thingie.

Arkshine:
Alright. Thanks!

PR #340, opened 2016-01-22
Fix CS_OnBuy* not triggered from alias containing uppercase letters

Arkshine:
Reported by GoRiLliAz (a.k.a the Bug Finder).

It's actually my fault, code was removed in #183 because function was not checking character validity, and I thought actually this was needed, thus removing all. But it's required as input from console is unchanged and uppercase letters are allowed by game.

Arbitrary buffer max length of 32 which should be largely enough to contain an alias name.

### DIFF
--- a/modules/cstrike/cstrike/CstrikeUtils.cpp
+++ b/modules/cstrike/cstrike/CstrikeUtils.cpp
@@ -12,6 +12,7 @@
 //
 
 #include "amxxmodule.h"
+#include <amtl/am-algorithm.h>
 
 extern int MessageIdTextMsg;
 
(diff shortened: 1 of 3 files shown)


Arkshine:
@Nextra , please.

Nextra:
why no in-place lowering like before? copying seems like an unnecessary step

Arkshine:
Because `menuselect` is fired from client and even the game server-side does a case-sensitive check, so I figured out it was not needed. 

About the unnecessary copy, what would be the right thing to do?

modules/cstrike/cstrike/CstrikeUtils.cpp
```diff
@@ -47,3 +48,22 @@ bool UTIL_CheckForPublic(const char *publicname)
 
 	return false;
 }
+
+void UTIL_StringToLower(const char *str, char *buffer, size_t maxlength)
```
xPaw: Is this really the first place where "string to lower" function was needed and used? There's probably more places in amxx codebase.

Arkshine: Well, others occurrences use a simplified version directly integrated to code. We don't have yet some manager with all those util functions so it can be used in modules. Could be done later.

Nextra:
I was just wondering if you had a good reason not to do it this time around. If you have, :shipit:

Arkshine:
Because argv() returns a `const char` and I don't want really modify original stringg?

PR #342, opened 2016-01-29
Add cs_get_item_alias(), improve cs_get_item_id() & cleanup alias datas

Arkshine:
- Added `cs_get_item_alias `native to lookup an alias (and if it exists its alternative name) associated with an item id.
- Cleaned up alias infos in gamedata file. There is now just one list and to avoid duplicating data, the alternative alias name has been moved into the associated alias. The `WeaponAlias` list has been removed because this is not used by game.
- Internally there is two aliases lists now: the default and the alternatives ones. The later is separated to keep the default list with unique item id.
- `cs_get_item_id` has been improved to make sure that a name is well tested with/out `item_` and `weapon_` prefixes.

### DIFF
--- a/modules/cstrike/cstrike/CstrikeItemsInfos.cpp
+++ b/modules/cstrike/cstrike/CstrikeItemsInfos.cpp
@@ -24,7 +24,6 @@ char WeaponNameList[MAX_WEAPONS][64];
 CsItemInfo::CsItemInfo()
 	:
 	m_ParseState(PSTATE_ALIASES_TYPE),
-	m_List(nullptr),
 	m_ListsRetrievedFromConfig(false)
 {}
 
@@ -35,9 +34,8 @@ CsItemInfo::~CsItemInfo()
 
 void CsItemInfo::Clear()
 {
-	m_CommonAliasesList.clear();
-	m_WeaponAliasesList.clear();
 	m_BuyAliasesList.clear();
+	m_BuyAliasesAltList.clear();
 }
 
 bool CsItemInfo::HasConfigError()
@@ -51,33 +49,15 @@ SMCResult CsItemInfo::ReadSMC_NewSection(const SMCStates *states, const char *na
 	{
 		case PSTATE_ALIASES_TYPE:
 		{
-			m_List = nullptr;
-
-			if (!strcmp(name, "CommonAlias"))
-			{
-				m_List = &m_CommonAliasesList;
-			}
-			else if (!strcmp(name, "WeaponAlias"))
-			{
-				m_List = &m_WeaponAliasesList;
-			}
-			else if (!strcmp(name, "BuyAlias") || !strcmp(name, "BuyEquipAlias") || !strcmp(name, "BuyAmmoAlias"))
-			{
-				m_List = &m_BuyAliasesList;
-			}
-
-			if (m_List)
+			if (!strcmp(name, "BuyAliases"))
 			{
 				m_ParseState = PSTATE_ALIASES_ALIAS;
 			}
-
 			break;
 		}
 		case PSTATE_ALIASES_ALIAS:
 		{
-			m_AliasInfo.clear();
 			m_Alias = name;
-
 			m_ParseState = PSTATE_ALIASES_ALIAS_DEFS;
 			break;
 		}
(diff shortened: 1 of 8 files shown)


Arkshine:
@Nextra

modules/cstrike/cstrike/CstrikeItemsInfos.cpp
```diff
@@ -190,47 +185,47 @@ bool CsItemInfo::GetAliasInfosFromBuy(const char *alias, AliasInfo *info)
 
 bool CsItemInfo::GetAliasInfosFromName(const char *name, AliasInfo *info)
 {
-	static const char prefix_weapon[] = "weapon_";
-	static const char prefix_item[] = "item_";
+	static const char prefix_weapon[] = "weapon";
+	static const char prefix_item[] = "item";
 
-	const char *alias = name;
+	if (name[6] == '_' && !strncmp(name, prefix_weapon, sizeof(prefix_weapon) - 1))
```
Nextra: Either hardcode the values all the time or use sizeof consistently. Maybe define another const for easier reading?

modules/cstrike/cstrike/CstrikeNatives.cpp
```diff
@@ -1809,6 +1809,40 @@ static cell AMX_NATIVE_CALL cs_get_item_id(AMX* amx, cell* params)
...
+		return 0;
+	}
+
+	auto itemid = params[1];
+
+	if (itemid == CSI_SHIELDGUN)
+	{
+		itemid = CSI_SHIELD;
+	}
+	else if (itemid == CSI_GLOCK)
```
Nextra: ?

modules/cstrike/cstrike/CstrikeNatives.cpp
```diff
@@ -1809,6 +1809,40 @@ static cell AMX_NATIVE_CALL cs_get_item_id(AMX* amx, cell* params)
...
+	{
+		itemid = CSI_GLOCK18;
```
Nextra: ?

modules/cstrike/cstrike/CstrikeItemsInfos.cpp
```diff
@@ -190,47 +185,47 @@ bool CsItemInfo::GetAliasInfosFromBuy(const char *alias, AliasInfo *info)
...
-		alias = name + sizeof(prefix_item) - 1;
+			return true;
+		}
 	}
 
-	if (GetAliasInfos(alias, info))
+	for (auto iter = m_BuyAliasesList.iter(); !iter.empty(); iter.next())
 	{
-		return true;
+		if (iter->value.classname.length() && !strcmp(iter->value.classname.chars() + sizeof(prefix_item), name))
```
Nextra: The strcmp looks erroneous.

Arkshine: Why? Are you saying I should make sure "item_" exists in classname?

modules/cstrike/cstrike/CstrikeNatives.cpp
Arkshine: _GLOCK has index of 2, which is not used by game for some reason (likely the original HL glock) but you have some internal arrays used for alias which redirects _GLOCK data  to _GLOCK18 and "???" alias is associated to _GLOCK. I guess it doesn't hurt to add some fallback. I should probably add the constant with some comment.

modules/cstrike/cstrike/CstrikeItemsInfos.cpp
```diff
@@ -190,47 +185,50 @@ bool CsItemInfo::GetAliasInfosFromBuy(const char *alias, AliasInfo *info)
 
 bool CsItemInfo::GetAliasInfosFromName(const char *name, AliasInfo *info)
 {
-	static const char prefix_weapon[] = "weapon_";
-	static const char prefix_item[] = "item_";
+	static const char prefix_weapon[] = "weapon";
+	static const char prefix_item[] = "item";
 
-	const char *alias = name;
+	static const size_t prefix_weapon_Length = sizeof(prefix_weapon) - 1;
```
Nextra: casing

modules/cstrike/cstrike/CstrikeNatives.cpp
Nextra: Okay, weird quirk. Should be fine then.

PR #344, opened 2016-02-01
Remove inline comment support in multilingual INI parser (bug 6476)

Arkshine:
Multilingual system uses INI parser since #260. Originally, only comment on a new line was supported.

If you do for example `KEY = ; VALUE`, the current parser will throw an Invalid multi-lingual line error.
It's possible to use `KEY = "; VALUE"`, but changing an old behavior for existing plugin is not welcomed.

For backward compatibility patch removes inline comment support for ML files only. Comment on new line is still working.

### DIFF
--- a/amxmodx/CLang.cpp
+++ b/amxmodx/CLang.cpp
@@ -490,7 +490,7 @@ int CLangMngr::MergeDefinitionFile(const char *file)
 	Data.currentFile = file;
 
 	unsigned int line, col;
-	bool result = textparsers->ParseFile_INI(file, static_cast<ITextListener_INI*>(this), &line, &col);
+	bool result = textparsers->ParseFile_INI(file, static_cast<ITextListener_INI*>(this), &line, &col, false);
 
 	if (!result)
 	{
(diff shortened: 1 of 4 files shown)


Nextra:
Meh I don't like this. But with the generic parser design I don't have a nice solution right now.

Arkshine:
I don't see any solution, '; ' was not designed to be an inline comment in the previous ML parser.

Actually, like we talked about it, If we change ML format like SM, we would not need this patch I guess.

Nextra:
I'd really want to keep inline comments in the parser but anything that preserves that would probably include an ugly hack. Not sure here.

Arkshine:
Then let's wait for a little whether new ML format is going to happen.

Nextra:
Well that won't solve the problem since the old parser still needs to be around

Arkshine:
You're right, it should keep the same behavior, even if we have a new ML format, backward compatibility is a must. Forgot this. 

I still don't know what to do, apart removing inline comment support. Maybe removing only for ML file and not for others files which textparser API. RIP consistency though.

Arkshine:
I've modified code to disallow inline comment support only for ML files.

Nextra:
Very much so. :shipit:

PR #345, opened 2016-02-03
Engine: Fix more bcompat for CHECK_ENTITY

Arkshine:
#250 + #259

`find_ent*` natives should not throw error. It's expected to loop over all found entities and letting coder checking entity validity before doing stuff.


Nextra:
We should just convert all of engine to TypeConversion, this INDEXENT stuff melts my brain.

Arkshine:
Will do that in the next PR.

Arkshine:
Just for reference:

`ENTINDEX`: edict -> index ; will sys error if calculated index is invalid
`INDEXENT`: index -> edict ; just returns null if passed index or if edict is invalid

Nextra:
:+1:

PR #348, opened 2016-02-03
Refactor and cleanup the AMBuildScript a bit

Arkshine:
Similar to https://github.com/alliedmodders/sourcemod/pull/427


Arkshine:
@dvander should be fine really, but checking fastly I did not miss something won't hurt.

dvander:
Looks fine - `cxx.vendor == 'msvc'` is better as `cxx.like('msvc')` but unless you're planning to switch to the 2.1 API it doesn't really matter.

PR #350, opened 2016-02-07
Add fmt() native to format and return a string inline 

Arkshine:
Motivation is we have a bunch of natives, like new menu API, which is not really multilingual friendly, Instead of creating alternative native or new whole API with formatting syntax, we introduce `fmt()` native which can format and return a string inline. 

That would encourage the use of multilingual where it's usually things are not translated and would improve slightly code readability/usability. 

This is not in any ways a replacement for `format/ex` and should be used only in native for fast and unique formatting. The buffer is set to 256 for sanity, and it should be largely enough for most of the situations.

### DIFF
--- a/plugins/include/string.inc
+++ b/plugins/include/string.inc
@@ -24,7 +24,12 @@
  * 			in the length.  This means that this is valid: 
  * 			copy(string, charsmax(string), ...)
  */
- 
+
+/**
+ * Buffer size used by fmt().
+ */
+#define MAX_FMT_LENGTH 256
+
 /**
  * Calculates the length of a string.
  *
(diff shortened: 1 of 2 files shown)


plugins/include/string.inc
```diff
@@ -168,6 +173,23 @@ native format(output[], len, const format[], any:...);
 native formatex(output[], len, const format[], any:...);
 
 /**
+ * Formats and returns a string according to the AMX Mod X format rules
+ * (see documentation).
```
xPaw: What documentation?

plugins/include/string.inc
```diff
@@ -168,6 +173,23 @@ native format(output[], len, const format[], any:...);
...
+ *
+ * @note Example: menu_additem(menu, fmt("My first %s", "item")).
+ * @note This should be used only for simple and unique inline formatting
+ *       like above example.  Do not use this function to store formated
+ *       string into a variable, use other formating functions.
```
xPaw: But why? Differences should be explained.

plugins/include/string.inc
Arkshine: A wiki page like SM has but for AMXX. It has yet to be created. Feel free to do it!

plugins/include/string.inc
Arkshine: There is an example, not sure what you would want to write more. Feel free to propose.

Arkshine:
Requires #351

plugins/include/string.inc
Nextra: If we want to comment on the overhead involved something like this might work?

@note This should only be used for simple inline formatting like in the above example. Avoid using this function to store strings into variables as an additional copying step is required.

plugins/include/string.inc
Nextra: This has probably existed somewhere at some point in time!

plugins/include/string.inc
Nextra: Also there are other performance implications on this native (namely the return buffer is temporary for each call) but I'd really not want to go that deep on the documentation for this. Our docs are detailed for minute performance penalties, sometimes to their detriment.

Nextra:
That `num_params + 1` feels odd but that's how it is. Code looks fine, doc needs a bit of polish. Otherwise :+1:

PR #351, opened 2016-02-08
Fix issue where native functions returning an array would not compile

Arkshine:
Fix from Pawn upstream where it fails to compile in some situations: https://github.com/compuphase/pawn/commit/f134bda57493cbbb45200e8e5711e1b221946829

Slightly changed by adding `root=origRoot` so `delete_symbol` knows that we're passing `&glbtab` since `RemoveFromHashTable` is called as well there.


Arkshine:
@dvander Please care of your silly :baby: as well.

dvander:
:ship:

PR #352, opened 2016-02-08
Fix linux/mac GiveNamedItem and GiveDefaultItems symbols 

Ni3znajomy:
Fix linux & mac signatures, windows signatures seem right.


Arkshine:
Damnit, how did I do this mistake. Thanks.

PR #353, opened 2016-02-13
Add extra parameters to cs_get_user_weapon native

Ni3znajomy:
Related to #330.

```get_user_weapon``` native has optional 2 parameters (clip & ammo), so I think ```cs_get_user_weapon``` native should have them too.

### DIFF
--- a/modules/cstrike/cstrike/CstrikeNatives.cpp
+++ b/modules/cstrike/cstrike/CstrikeNatives.cpp
@@ -1946,11 +1946,14 @@ static cell AMX_NATIVE_CALL cs_get_user_weapon_entity(AMX *amx, cell *params)
 	return (weaponEntIndex != -1) ? weaponEntIndex : 0;
 }
 
-// native cs_get_user_weapon(playerIndex);
+// native cs_get_user_weapon(playerIndex, &clip = 0, &ammo = 0);
 static cell AMX_NATIVE_CALL cs_get_user_weapon(AMX *amx, cell *params)
 {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: modules/cstrike/cstrike/CstrikeNatives.cpp [hunk 1]
```diff
@@ -1961,7 +1964,15 @@ static cell AMX_NATIVE_CALL cs_get_user_weapon(AMX *amx, cell *params)
...
 	if (!FNullEnt(pWeapon))
 	{
-		return get_pdata<int>(pWeapon, m_iId);
+		int iId = get_pdata<int>(pWeapon, m_iId);
+
+		cell *cpTemp = MF_GetAmxAddr(amx, params[2]);
+		*cpTemp = get_pdata<int>(pWeapon, m_iClip);
+
+		cpTemp = MF_GetAmxAddr(amx, params[3]);
+		*cpTemp = (iId != CSW_KNIFE) ? get_pdata<int>(pPlayer, m_rgAmmo, get_pdata<int>(pWeapon, m_iPrimaryAmmoType)) : 0;
```
Arkshine: Instead checking id, I would check if `m_iPrimaryAmmoType` <= 0, because ammo index starts to 1 and you have -1 when the index is not found from ammo name (typically like the knife which doesn't have ammo name) and that would make the code more generic. What do you think?

Ni3znajomy: Sounds logical to me.

Arkshine:
Looks good. For consistency with `get_user_weapon`, it makes sense to add those parameters.

PR #346, opened 2016-02-03
Engine: Replace ENTINDEX with TypeConversion

Arkshine:
- Some `find_ent_` natives were using `is_valid_ent` which checks for players if they are in game. Considering such index is just to define a start index to search from, and a player's edict is always valid internally, it should be okay to skip it.
- Added some extra check just in case, because `edict_to_id` could return -1 if edict is null, and with array this doesn't  work well. This is really unlikely to have invalid edict in such forwards, but extra check doesn't kill.

PR #355, opened 2016-02-24
Fix typo in find_ent_in_sphere()

Arkshine:
Reported by Sylvert0. Related to #346.

Not sure why compiler ignored this.


Nextra:
Probably because the function argument is `void*` and that eats just about anything. I shouldn't have missed this one. :ship:

PR #356, opened 2016-02-25
Fix issues with some cstrike natives

Arkshine:
- Fixed bcompat with `cs_s/get_user_backpackammo`. 
 - Reported by Depresie [here](https://forums.alliedmods.net/showthread.php?t=279551). 
 - Related to #255.
 - Currently, to set `m_rgAmmo`, it relies on `m_rgpPlayerItems` to find the associated item in order to retrieve the ammo index. But it supposes that item is still present in the player's inventory. If you delete your item and you call these natives, this won't work since the item has been already unlinked in `m_rgpPlayerItems`. It was a bad idea to use that. It was working before because it was hardcoded with offset. 
 - This patch retrieves ammo index from WeaponList, and actually simplify the native code a lot. Sanity check has been modified to be more generic.
- Fixed a silly typo in cs_get_translated_item_alias()
- Fixed incomplete sanity check in cs_get_weapon_info().

### DIFF
--- a/modules/cstrike/cstrike/CstrikeItemsInfos.cpp
+++ b/modules/cstrike/cstrike/CstrikeItemsInfos.cpp
@@ -15,7 +15,7 @@
 #include "CstrikeHacks.h"
 
 CsItemInfo ItemsManager;
-char WeaponNameList[MAX_WEAPONS][64];
+ItemInfo WeaponsList[MAX_WEAPONS];
 
 #define PSTATE_ALIASES_TYPE       0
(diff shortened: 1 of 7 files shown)


Arkshine:
@Nextra please.

Javivi:
I'm not Nextra, but the changes look good to me. Good job with the bpammo change

Arkshine:
Thanks for reviewing. You will be responsible for any damage patch would cause! :P

PR #357, opened 2016-02-28
Filter evil characters in name and message (bug 5912, bug 6301)

Arkshine:
This fixes a known exploit around localized string and var arg, like `#Spec_Help_Text`, `+forward` or `%s` , which can cause bugs with multiple occurrences, at worst crashing a client.

While it should not crash on the client (typically an overflow with localized strings), unfortunately, this is happening because AMXX sends a message as it is directly and people are enough cunning to use such string in their name as well.

The concerned messages are: `TextMsg`, `SayText` and `ShowMenu`.

This means that any plugin which displays a menu with player's name (like slap menu) or rewrites the chat output (like adminchat with `say_team @`), at best you get an unwanted translated message, at worst it crashes the client who receives the message.

Ideally, the name should be filtered in the engine, and anything typed in `say/_team` in the mod, but since AMXX is equally responsible, it makes sense we do this filtering.

-- 

This patch filters the following characters in those specific events:
 - When at player changes name
  - By filtering name, this should cover most of the situations since name is often inserted in message
 - When a player says something via `say` and `say_team`
  - By filtering when a command is issued, it should cover `read_args` which is used by the plugin to retrieve the whole input buffer from chat before being passed in natives using `TextMsg`, `SayText`  such as  `client_print/_color` or manually via `write_string`.

Character           | Description                                     | Message
----------------------|-----------------------------------------------|-----------------------------|
 `+` `<key>`        | Key binds. Translated to `[key] `.  |  `ShowMenu`
 `&` `<value>`     | Hotkeys in VGUI. Unsure.             | `ShowMenu` ?
 `#` `<value>`     | Localized string                             | `ShowMenu`, `SayText`, `TextMsg`
 `%` `s`                | Format specifier                            | `SayText`, ``TextMsg`

Some notes ^:
 - I'm not totally sure about `&`. It seems engine already filters it by replacing by spaces, but for safety I keep it
 - `%` should be filtered by the mod if you opt for the beta, but for safety I keep it.
 - Only `#` and `%` are concerned in messages

-- 

I would like some feedbacks.
 - Characters are replaced with similar Unicode (code written by @WPMGPRoSToTeMa ). What do you think about it? Is a config should be added allow admin to choose whether they want to replace evil characters with a specific one instead of Unicode?
 - Any usage I could have forgotten?

### DIFF
--- a/amxmodx/amxmodx.h
+++ b/amxmodx/amxmodx.h
@@ -142,6 +142,7 @@ unsigned int UTIL_ReplaceAll(char *subject, size_t maxlength, const char *search
 char *UTIL_ReplaceEx(char *subject, size_t maxLen, const char *search, size_t searchLen, const char *replace, size_t replaceLen, bool caseSensitive);
 void UTIL_TrimLeft(char *buffer);
 void UTIL_TrimRight(char *buffer);
+size_t UTIL_FilterEvilCharacters(const char *in, char *out, size_t out_maxlen, bool name_only);
 
 #define GET_PLAYER_POINTER(e)   (&g_players[ENTINDEX(e)])
 //#define GET_PLAYER_POINTER(e)   (&g_players[(((int)e-g_edict_point)/sizeof(edict_t))])
--- a/amxmodx/meta_api.cpp
+++ b/amxmodx/meta_api.cpp
@@ -149,6 +149,7 @@ int FF_ClientConnectEx = -1;
 
 IFileSystem* g_FileSystem;
 HLTypeConversion TypeConversion;
+server_static_t *ServerStatic;
 
 bool ColoredMenus(const char *ModName)
 {
@@ -929,6 +930,27 @@ void C_ClientPutInServer_Post(edict_t *pEntity)
 	RETURN_META(MRES_IGNORED);
 }
 
+void C_ClientUserInfoChanged_Pre(edict_t *pEntity, char *infobuffer)
+{
+	if (ServerStatic)
+	{
+		auto clientIndex = TypeConversion.edict_to_id(pEntity);
+
+		const char *oldName = ServerStatic->clients[clientIndex - 1].name;
+		const char *newName = INFOKEY_VALUE(infobuffer, "name");
+
+		if (oldName[0] == '\0' || strcmp(newName, oldName) != 0)
+		{
+			char safeNewName[MAX_NAME];
+			UTIL_FilterEvilCharacters(newName, safeNewName, sizeof(safeNewName), true);
+
+			SET_CLIENT_KEYVALUE(clientIndex, infobuffer, "name", safeNewName);
+		}
+	}
+
+	RETURN_META(MRES_IGNORED);
+}
+
 void C_ClientUserInfoChanged_Post(edict_t *pEntity, char *infobuffer)
 {
 	CPlayer *pPlayer = GET_PLAYER_POINTER(pEntity);
@@ -1006,6 +1028,19 @@ void C_ClientCommand(edict_t *pEntity)
 #endif
 			CLIENT_PRINT(pEntity, print_console, buf);
 			RETURN_META(MRES_SUPERCEDE);
+		}
+	}
+
+	if (!strcmp(cmd, "say") || !strcmp(cmd, "say_team"))
+	{
+		auto args = CMD_ARGS();
+
+		if (args && *args)
+		{
+			char safeMessage[128];
+			auto length = UTIL_FilterEvilCharacters(args, safeMessage, sizeof(safeMessage), false);
+
+			memcpy(const_cast<char*>(args), safeMessage, length + 1);
 		}
 	}
 
@@ -1586,15 +1621,37 @@ C_DLLEXPORT	int	Meta_Attach(PLUG_LOADTIME now, META_FUNCTIONS *pFunctionTable, m
 
 	g_CvarManager.CreateCvarHook();
 
-	void *address = nullptr;
+	if (CommonConfig)
+	{
+		void *address = nullptr;
 
-	if (CommonConfig && CommonConfig->GetMemSig("SV_DropClient", &address) && address)
-	{
-		DropClientDetour = DETOUR_CREATE_STATIC_FIXED(SV_DropClient, address);
-	}
-	else
-	{
-		AMXXLOG_Log("client_disconnected forward has been disabled - check your gamedata files.");
+#if defined(KE_WINDOWS)
+		TypeDescription typeDesc;
+
+		if (CommonConfig->GetOffset("svs", &typeDesc))
+		{
+			auto base = *reinterpret_cast<uintptr_t*>(reinterpret_cast<byte*>(g_engfuncs.pfnGetCurrentPlayer) + typeDesc.fieldOffset);
+			ServerStatic = reinterpret_cast<decltype(ServerStatic)>(base - 4);
+		}
+#else
+		if (CommonConfig->GetMemSig("svs", &address))
+		{
+			ServerStatic = reinterpret_cast<decltype(ServerStatic)>(address);
+		}
+#endif
+		else
+		{
+			AMXXLOG_Log("svs global variable is not available - check your gamedata files.");
+		}
+
+		if (CommonConfig->GetMemSig("SV_DropClient", &address) && address)
+		{
+			DropClientDetour = DETOUR_CREATE_STATIC_FIXED(SV_DropClient, address);
+		}
+		else
+		{
+			AMXXLOG_Log("client_disconnected forward has been disabled - check your gamedata files.");
+		}
 	}
 
 	GET_IFACE<IFileSystem>("filesystem_stdio", g_FileSystem, FILESYSTEM_INTERFACE_VERSION);
@@ -1739,6 +1796,7 @@ C_DLLEXPORT	int	GetEntityAPI2(DLL_FUNCTIONS *pFunctionTable, int *interfaceVersi
 	gFunctionTable.pfnInconsistentFile = C_InconsistentFile;
 	gFunctionTable.pfnServerActivate = C_ServerActivate;
 	gFunctionTable.pfnClientConnect = C_ClientConnect;
+	gFunctionTable.pfnClientUserInfoChanged = C_ClientUserInfoChanged_Pre;
 
 	memcpy(pFunctionTable, &gFunctionTable, sizeof(DLL_FUNCTIONS));
 
--- a/amxmodx/util.cpp
+++ b/amxmodx/util.cpp
@@ -718,3 +718,41 @@ void UTIL_TrimRight(char *buffer)
 		}
 	}
 }
+
+size_t UTIL_FilterEvilCharacters(const char *in, char *out, size_t out_maxlen, bool name_only)
+{
+	auto remainChars = out_maxlen - 1;
+	auto n = 0;
+
+	for (auto s = in; *s != '\0' && remainChars; ++s)
+	{
+		if (*s == '#' ||
+			*s == '%' ||
+			(name_only &&
+				*s == '&' ||
+				(n && out[n - 1] == '+' && static_cast<signed char>(*s) > 0 && isalnum(*s))
+			))
+		{
+			if (remainChars < 3)
+			{
+				break;
+			}
+
+			// http://unicode-table.com/blocks/halfwidth-and-fullwidth-forms/
+			out[n++] = 0xEF;
+			out[n++] = 0xBC | (((*s - 0x20) & 0x40) >> 6);
+			out[n++] = 0x80 + ((*s - 0x20) & 0x3F);
+
+			remainChars -= 3;
+		}
+		else
+		{
+			out[n++] = *s;
+			remainChars--;
+		}
+	}
+
+	out[n] = '\0';
+
+	return n;
+}


LeninChan:
ahahahah

Phantom-J:
It's hapenned

Arkshine:
What do you mean? I know it has been around since some time, and you have some metamod module  (like localizebugfix or SafeChatAndMessage) for that, even reHLDS partially fixed, but it should be anyway available by default. 

I would prefer more helpful answer really.

IgnacioFDM:
I'm not sure if I understand correctly. Will the character + be replaced unconditionally every time you show a menu?

Arkshine:
Yes, by an unicode character. http://unicode-table.com/blocks/halfwidth-and-fullwidth-forms/

IgnacioFDM:
And they look pixel identical ingame I would guess?

Arkshine:
I would not say identical, but it's very close. 

I.e.: image

IgnacioFDM:
Close enough. On HUD/Menus fonts I'm sure it will look flawless (since it's font much better with unicode). A cvar to disable this feature entirely would be nice.

Arkshine:
Why would you want to disable a protection?

IgnacioFDM:
Because this protection isn't seamless (it replaces user names, and strings in a way that is invisible to plugins) and you may want to handle this at a plugin level or whatever. Options are always a good idea when you have new, different behavior

luxxxoor:
with this feature, do we still need client_printex() stock ? 
https://www.amxmodx.org/api/message_stocks/client_printex

Arkshine:
@luxxxoor We have no natives to send predefined text with arguments like does the game. You can do that manually via `message_begin` and this is what the stock is for.

@IgnacioFDM 
Well, the whole point of using Unicode is because the user is more important than coder here. Change is done before everything else, it should not make much difference, the plugin still receives an original input, right? 
Actually name is already partially filtered by the engine before `pfnClientUserInfoChanged` is called. So for name it's not going to be a surprise.

But I think you're saying that plugin should get an unmodified input and filtering should happen only in case message is going to be sent, after plugin did their sutff on it. Purpose would be then to keep current engine/mod/amxx behavior but filtering at the very end for displaying. For a fix inside AMXX, this would be actually more welcomed I guess. Is it what you mean?

About filtering a plugin, it's probably a bad idea. You can filter properly unless you're using and fixing an exploit should be anyway in core.
About the option, It could make sense but not for the reason you said. The only valid I see is if you're using reHLDS/GameDLL and because they have a similar protection, that would be redundant.

IgnacioFDM:
Filtering happening after sending the message (so plugins get raw input) is a much better way to do it. Imagine you have a plugin where you have (similar to admin say) hook"say" clcmd, and you print text in a special way when the message begins with + (or #) (similarly to how @ is treated for admin say). Now not only you break compatibility with older plugins without support, but you also require plugins to detect some silly longer than 1 byte unicode character instead of simply ```text[pos] == '+'```, when users are simply pressing the + button (this is really unintuitive to some developer who hasn't read this PR).

Besides, the cvar for future regamedll/hlds changes as you said is very important. Furthermore, implementing a cvar is simply less effort and would have taken less time than this discussion :laughing: 

In general, I've been frustrated many times (in other games) for having "protections" (such as anti speed hack, anti DoS or whatever) that end up backfiring, so having an option to toggle is always a good idea unless we are talking of something 100% innocuous such as maybe fixing some buffer overflow or whatever.

By the way, is the adminsay.sma (or whatever it's called) exploit fixed, where you could ```say_team @%s0 hehe``` resulting in admin's game crashing? There are many exploits possible (and working on approved plugins) not related to client side parsing of %s (and similar % keywords or whatever they are called), but actually related to parsing them with formatex

Arkshine:
I agree to filter `say/_team`after plugins are done is a more backward compatible solution, so I guess I will commit later. But the change in name is probably fine as it is (because engine already filters things, and because it will be tricky to filter `ShowMenu`).

About cvar, if the patch doesn't mess with the original input, you have no reason to have it disabled, but because of reHLDS/GameDLL and maybe other tools/contexts, it's okay to add one I guess.

About `%s`, it's filtered in the client if you opt for the beta If I remember. It will replace `s` by space.
But I'm not aware of any others exploits with `%`, so if you know something, it's time to say it.

Thanks for your feedback.

IgnacioFDM:
Did you try 
```say_team @%s```
and
```say_team @%s0```

(admins' client should crash)

I'm not home and can't test now, but that's a confirmed exploit on amx <= 1.8.2

**Comment by IgnacioFDM** [comment]
(I mean without this commit, with this commit I'm sure it won't work)

**Comment by Arkshine** [comment]
```
] say_team @%s
(ADMIN) Arkshine :   s
] say_team @%s0
(ADMIN) Arkshine :   s0
```

When I said beta, I mean the client CS. Right now, I have opted for the beta and it doesn't crash.
It probably crashes if non-beta.

**Comment by Arkshine** [comment]
Yes, you're right about client_t, but probability struct is updated again is near zero. Could be fixed through a gamedata file. It's anyway not a good example since it can be done differently. 

I know about svencoop, but they do what they can to not break compatibility.Though they are trying to get rid of AMXX with their AngelScript, so sooner or later, things might be complicated. Considering the low number of servers under AMXX and because I'm kind of alone, dealing with svencoop specificities is not really a priority. But this should be fine for some time, this is not in their interest.

About your last question, functionalities relying on gamedata are optional. Only associated stuff are concerned.

**Comment by In-line** [comment]
@Arkshine Any news ?

**Comment by IgnacioFDM** [comment]
Some feedback after having used something similar to this over a year:

I've been using the UTIL_FilterEvilChars reimplemented in pawn with great success (I needed having access to the "raw data" and a bit more fine-grained control than this PR). 
To merge this PR I'd say we need to
- Check interactions with ReHLDS since it does this with names by default
- Native/stock is definitely needed, so developers can filter their stuff (think menus, data coming externally (databases, userinfo, etc) and so on)
- I really don't like the "feeding fake data to read_args". I get the PHP Magic Quotes vibe from it. Maybe a better one size fits all approach would be filtering client_print? I don't know. I haven't really thought much about other solutions that can "backport" this fix to older plugins. Not too fussed about this as long as there is a cvar to disable it, since it can unleash chaos in some scenarios (it did to me, think duplicate SQL entries for names or other user input with the both the older normal characters and the weird new ones, or admin commands needing correct characters to work such as checking for hardcoded strings like "#TTs")
- As mentioned above, cvars to disable this

That's my personal opinion of course.

**Comment by WPMGPRoSToTeMa** [comment]
Chat can be fixed in another way for CS. `#Spec_PlayerItem` expands into `%s1`, [it is used in my ChatPrint](https://github.com/WPMGPRoSToTeMa/ChatPrint/commit/79750c229315e1d5840535cf16ce39725ba6e56d).

**Comment by WPMGPRoSToTeMa** [comment]
Chat doesn't need fixes since last update.

**Comment by Arkshine** [comment]
Totally forgot about this. I'm kind of lost about this.

@WPMGPRoSToTeMa, do you know exactly what it still needs to be filtered in HLDS and what to skip with ReHLDS?

**Comment by WPMGPRoSToTeMa** [comment]
@Arkshine nothing with ReHLDS and only nicknames with HLDS. `&`, `%` are replaced by default in HLDS, `#` is replaced too, but only at the beginning of name. We can also add gamedata files for `+` commands and localizable strings to avoid unnecessary replacements and save compatibility with safe nicknames. And also we can add functions that can print `%` and `#` symbols in chat messages (`#Spec_PlayerItem` that is used in [my ChatPrint](https://github.com/WPMGPRoSToTeMa/ChatPrint/commit/79750c229315e1d5840535cf16ce39725ba6e56d) can help).

**Comment by Arkshine** [comment]
Everything I'm about to say can be discussed, right.

The API is kind of cumbersome. You should keep it simple and let coder extend them if necessary. In case they need to use a handle across plugins, they can use a fake native (like you used in your plugin) or xvar for example.

Having something like that would be enough:
```PAWN
enum Logger
{
    Invalid_Logger = 0,
};

enum LoggerFormat
{
    LogFormat_Name,
    LogFormat_Message,
    LogFormat_Date,
    LogFormat_Time,
    LogFormat_Path,
    LogFormat_Trace
};

enum LoggerSeverity
{
    LogSeverity_None  = 0,
    LogSeverity_Debug = 1,
    LogSeverity_Info  = 101,
    LogSeverity_Warn  = 201,
    LogSeverity_Error = 301,
};

native Logger:LoggerCreate([...]);
native LoggerDestroy(&Logger:handle);

native LoggerLog(const Logger:handle, const LoggerSeverity:severity, const format[], any:...);

native LoggerLogDebug(const Logger:handle, const format[], any:...);
native LoggerLogInfo(const Logger:handle, const format[], any:...);
native LoggerLogWarn(const Logger:handle, const format[], any:...);
native LoggerLogError(const Logger:handle, const format[], any:...);

native LoggerSeverity:LoggerGetVerbosity(const Logger:handle);
native LoggerSeverity:LoggerSetVerbosity(const Logger:handle, const LoggerSeverity:verbosity);
native LoggerSeverity:LoggerSetDefaultVerbosity(const LoggerSeverity:verbosity);

native LoggerFormat:LoggerGetFormat(const Logger:handle, LoggerFormat:type, format[], const maxLength);
native LoggerFormat:LoggerSetFormat(const Logger:handle, LoggerFormat:type, const format[]);
native LoggerFormat:LoggerSetDefaultFormat(LoggerFormat:type, const format[]);
```
You get rid of "ThisLogger" and "AllLogger" handle which is confusing and doesn't apply to all natives.
You get rid of duplicated natives. If you are using an handle you should only use that.
You separate per-handle and all handlers configuration.
You factorize format configuration because they have all the same header.
This is more straight & sane I believe.

That's said, this doesn't feel right.
Having a create/free handle system when you are going to use always them globally is a bit silly.
And I don't see any usage having several loggers per plugin.

Now we removed explicit handles, we need a way to extend a plugin's log config to another plugin.
Since we are working per plugin, we should probably have a native which passes a plugin's id that you wish to copy its config from.
Similar to handle, you could pass an id either with a fake-native or xvar.

At the end I would do, maybe something like:
```PAWN
enum LoggerFormat
{
    LogName,
    LogMessage,
    LogDate,
    LogTime,
    LogPath,
    LogTrace
};

enum LoggerVerbosity
{
    UnsetLevel = 0,
    DebugLevel = 1,
    InfoLevel  = 101,
    WarnLevel  = 201,
    ErrorLevel = 301,
};

native Logger(const LoggerVerbosity:verbosity, const format[], any:...);

native LoadLogger(const pluginId);

native LoggerVerbosity:GetLoggerVerbosity();
native LoggerVerbosity:SetLoggerVerbosity(const LoggerVerbosity:verbosity);

native LoggerFormat:GetLoggerFormat(LoggerFormat:type, format[], const maxLength);
native LoggerFormat:SetLoggerFormat(LoggerFormat:type, const format[]);
```
As you can see it\'s much easier and cleaner.
* I removed the create/free handle natives.
* I removed `*Default*` natives because they are not really useful in a normal plugin. Changing the default configuration should be handled by an admin using the cvar provided by the logger plugin, and you can loop over all the plugin's id there.
* I tried to rename constants name and their length to be more comfortable and less redundant.
* Since verbosity constant names have been reduced, I think having multiple `LoggerLog*` is now redundant.
* I added `LoadLogger` native to essentially copy an existing plugin logger configuration.
* Note sure about `Logger` alone and 'LoadLogger' if enough descriptive.

As a side-note, about the format specifiers, is it really needed to have precision/width here?
I would go with named specifier, like `{plugin}`. Can use `%` or whatever.

Single plugin:
```PAWN
public plugin_init()
{
    register_plugin("Plugin", "1.0", "Github");
}

foo()
{
    Logger(WarnLevel, "This is warning");
}
```

Multiple plugin:
```PAWN
#include <amxmodx>

new MainPluginId;

public plugin_init()
{
    MainPluginId = register_plugin("Plugin A", "1.0", "Github");

    // Changing default configuration
    SetLoggerVerbosity(DebugLevel);
    SetLoggerFormat(LogName   , "{plugin} | {datestamp}");
    SetLoggerFormat(LogMessage, "[{verbosity}] [{timestamp}] {script}::{function} - {message}");
    SetLoggerFormat(LogTrace  , "    at {script}::{function} : {line}");
}

public plugin_natives()
{
    register_native("GetBasePluginId", "@Native_GetBasePluginId");
}

@Native_GetBasePluginId()
{
    return MainPluginId;
}

foo()
{
    Logger(InfoLevel, "Hello world!");
    Logger(ErrorLevel, "This is an error.");
}
```
```PAWN
#include <amxmodx>

native GetBasePluginId();

public plugin_init()
{
    register_plugin("Plugin B", "1.0", "Github");

    LoadLogger(GetBasePluginId());
}

foo()
{
    Logger(InfoLevel, "Hello world!");
}
```

Overall, I think it's not that bad.

Any thoughts?

Arkshine:
Thanks, will see what I can do.

Arkshine:
@WPMGPRoSToTeMa Looks like ReHLDS doesn't do that. I'm not sure if it's a good idea that both don't behave the same. Any thoughts on it? Is it really worth? Actually, what are the compatibility issues, I see tag checking I guess, are there others?

About localized strings, if I remember, only the keys in /resources/$mod_english.txt are concerned, right?

WPMGPRoSToTeMa:
> Looks like ReHLDS doesn't do that.

I didn't think about that when I commited in ReHLDS. There is still an open issue https://github.com/dreamstalker/rehlds/issues/198, so I will resolve it in ReHLDS in future. Some people also said that these symbols aren't displayed correctly for them (empty squares).

> About localized strings, if I remember, only the keys in /resources/$mod_english.txt are concerned, right?

In `ShowMenu` only `titles.txt` is considered. This file is present on the server-side, so looks like we can easily use it without any additional gamedata file.
Hm, `config.cfg` is also present on the server-side. We can retrieve a + commands list from it. But looks like it doesn't contain a full list. Manual collecting of + commands isn't so hard, we can just use `cmdlist log` command on the client and leave strings starting with + from the cmdlist.txt file. Basically if the gamedata file isn't present we can use `config.cfg`, then if it is also not present we can disable this option or behave as ReHLDS. (same for localized strings)

afwn90cj93201nixr2e1re:
and what bout client_command cmd's?

Nord1cWarr1or:
any news?

djearthquake:
Country on Name with somebody from North Africa got my attention making a lag storm about a decade ago. The newer AMXX at least we can detect evil symbol and rename client on the spot with this. If admin does not have a list of names to change client to this will generate one. Originally made for bots and has CVARS to keep it on the bots or humans too to prevent this lame exploit.
```c
/*
 * This program is free software; you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation; either version 2 of the License, or
 * (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with this program; if not, write to the Free Software
 * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
 * MA 02110-1301, USA.
 *
 * When clients connect with the duplicate name or non-utf8 it renames them.
 *
    *
    Credit: OciXCroM https://forums.alliedmods.net/showthread.php?t=336394. Helped with rename and array.
    Install instructions: Recommended quick start:
    Symbolic link BotNames.ini with existing 'alternative' bot names file.
    *
 *
*/

#include <amxmodx>
#include <amxmisc>

#define MAX_PLAYERS                32
#define MAX_NAME_LENGTH            32
#define MAX_AUTHID_LENGTH          64
#define MAX_CMD_LENGTH             128
#define MAX_USER_INFO_LENGTH       256

#define PLUGIN "!Client ReNaMeR"
#define VERSION "1.0"

new const SzBotFileName[]="/BotNames.ini"

new Array:g_aBotNames, g_iTotalNames;

new ClientAuth[MAX_PLAYERS+1][MAX_AUTHID_LENGTH];
new ClientName[MAX_PLAYERS+1][MAX_NAME_LENGTH];
new XBots_only, XCyborgFilter, XUTF8_Strafe;

static const SzInitFakeName[] = "gamer"

public client_infochanged(id)
{
    if(is_user_connected(id))
    {
        if(!is_user_bot(id) && get_pcvar_num(XBots_only))
            return

        get_user_name(id,ClientName[id],charsmax(ClientName[]))

        new Szcheck[2]
        copy(Szcheck, charsmax(Szcheck), ClientName[id])

        if(containi(ClientName[id],"(1)") > -1 || equal(ClientName[id], "") || get_pcvar_num(XUTF8_Strafe) && get_char_bytes(Szcheck) != 1)
            @player_rename(id)
    }
}

@player_rename(id)
{
    if(g_iTotalNames)
    {
        if(is_user_connected(id))
        {
            if(is_user_bot(id) && get_pcvar_num(XCyborgFilter))
            {
                get_user_authid(id,ClientAuth[id],charsmax(ClientAuth[]))
                if(!equali(ClientAuth[id], "BOT"))
                    return
            }
            if(g_iTotalNames--)
            {
                new szName[MAX_NAME_LENGTH], i = random(g_iTotalNames)
                ArrayGetString(g_aBotNames, i, szName, charsmax(szName))
                ArrayDeleteItem(g_aBotNames, i)

                static const szNameField[] = "name"
                server_print "%s renamed to %s", ClientName[id], szName
                set_user_info(id, szNameField, szName)
            }
            else
            {
                server_print "Try growing your name list!"
                @init_fake_file() //will make new name based on time
            }
        }

    }
    else
    {
        log_amx "Try growing your array list!"
        pause ("c")
    }
}

public plugin_init()
{
    register_plugin(PLUGIN, VERSION, "SPiNX")

    XCyborgFilter   = register_cvar("sv_rename_intregrity","0") //If humans complain when sv_rename_humans 1 seldomly fails.
    XBots_only      = register_cvar("sv_rename_humans","0") //This was made for bots but applied to humans also.
    XUTF8_Strafe    = register_cvar("sv_rename_utf","0") //Ever watched the flood from renaming non-utf8 CountryOnName? Not pretty.

    g_aBotNames = ArrayCreate(MAX_NAME_LENGTH)
    ReadFile()
}

public plugin_end()
    ArrayDestroy(g_aBotNames)

ReadFile()
{
    new szFilename[MAX_USER_INFO_LENGTH]
    get_configsdir(szFilename, charsmax(szFilename))
    add(szFilename, charsmax(szFilename), SzBotFileName)
    new iFilePointer = fopen(szFilename, "rt")
    new szData[MAX_NAME_LENGTH]

    if(!iFilePointer)
    {
        @init_fake_file()
        return
    }

    else if(iFilePointer)
    {
        while(!feof(iFilePointer))
        {
            fgets(iFilePointer, szData, charsmax(szData))
            trim(szData)

            switch(szData[0])
            {
                case EOS, '#', ';': continue
                default:
                {
                    g_iTotalNames++
                    ArrayPushString(g_aBotNames, szData)
                }

            }

        }
        fclose(iFilePointer)
    }

}

@init_fake_file()
{
    new rSzName[MAX_NAME_LENGTH]
    new mod_name[MAX_NAME_LENGTH]

    new SzBuffer[MAX_NAME_LENGTH]

    //make name off what is in script
    copy(SzBuffer, charsmax(SzBuffer), SzInitFakeName)
    @file_data(SzBuffer)

    //make name off what is in script 1st LTR upper
    copy(SzBuffer, charsmax(SzBuffer),SzInitFakeName)
    mb_ucfirst(SzBuffer, charsmax(SzBuffer))
    @file_data(SzBuffer)

    //add time in epoch to default alias
    formatex(rSzName, charsmax(rSzName), "%s:%i", SzInitFakeName, get_systime())
    copy(SzBuffer, charsmax(SzBuffer), rSzName)
    @file_data(SzBuffer)

    //add time in epoch to default alias Swapped
    formatex(rSzName, charsmax(rSzName), "%s:%i", SzInitFakeName, swapchars(get_systime()))
    copy(SzBuffer, charsmax(SzBuffer), rSzName)
    @file_data(SzBuffer)

    //make name off mod name
    get_modname(mod_name, charsmax(mod_name))
    copy(SzBuffer, charsmax(SzBuffer), mod_name)
    @file_data(SzBuffer)

    //make name off what is in script all upper
    mb_strtoupper(SzBuffer, charsmax(SzBuffer))
    @file_data(SzBuffer)

    //make name off mod name 1sr cap plus time
    mb_ucfirst(mod_name, charsmax(mod_name))
    formatex(SzBuffer, charsmax(SzBuffer), "%s:%i", mod_name, get_systime())
    //mb_strtolower(SzBuffer, charsmax(SzBuffer))
    @file_data(SzBuffer)

    //go back to reading
    ReadFile()
}

@file_data(SzBuffer[MAX_NAME_LENGTH])
{
    server_print "%s|trying save %s", PLUGIN, SzBuffer
    new szFilePath[ MAX_USER_INFO_LENGTH ]
    get_configsdir( szFilePath, charsmax( szFilePath ) )
    add( szFilePath, charsmax( szFilePath ), SzBotFileName )

    write_file(szFilePath, SzBuffer)
}
```

PR #358, opened 2016-03-02
Allow null invoker to be passed in EngFunc_PlayBackEvent (bug 6467)

Arkshine:
This allows passing `-1` (you can use `FM_NULLENT`) in `EngFunc_PlayBackEvent` for invoker to tell the module to pass a `NULL` invoker edict. This is allowed by the engine to use a specific origin rather than the invoker's origin.

This fix the forward as well to return -1 instead of 0 if invoker edict is null. Should be safe change.


Nextra:
:+1: and hope this doesn't go nuclear :P

PR #359, opened 2016-03-16
Improve UTF-8 support (bug 6475)

Arkshine:
A first attempt to extend UTF-8 support for some natives.
It relies on [UTF-8 Rewind](https://bitbucket.org/knight666/utf8rewind), which is small, cross-platform, easy-to-use, well-tested C library. And enough for our needs.

Compatibility should be fine as utf-8 was never really supported before.

Current modified natives:
- Improving default behavior:
  - `strtolower` and `strtoupper`
  - `strcmp`, `strncmp` and `strfind` when case-insensitive is enabled only
  - `replace_stringex` and `replace_string` when case-insensitive is enabled only
  - `equali` and `containi`
  - `ucfirst`
- With an additional argument:
  - `strlen`, a new parameter to return the number of full characters (codepoints).
- New natives:
  - `is_string_category` to check if the input string conforms to the category specified by flags (you can check the whole string or just one character, it's meant to to be a replacement for `is*` string natives and  more unicode specificities.
  - `strtotitle` while we are on it and since the library offers a function for that.


IgnacioFDM:
Although I think trim should by default remove unicode whitespace because it's the msot logical (imo), since it doesn't now it's actually best (necessary in fact) to not remove unicode whitespace by default since that's the current behavior and it would break compatibility changing it. 

>What about the others natives which are using internally isspace, isalpa, etc. Should we have a global config for them?

What do you mean with global config for them? As in a server cvar or something like that to determine if it's unicode aware? That would be a pretty bad idea in my opinion. Having some external global state that modifies the behavior of simple functions is a bad idea.

Having a "strtotitle" would be nice, since it's just wrapping ``utf8totitle``

About adding parameters to strlen (and others), I don't remember exactly how pawn VM works, so I have to ask, can you add a parameter to an already existing native without breaking compatibility with already compiled plugins? I would asume you need to push all arguments to the stack if you don't want to read garbage, but I could be wrong since I've never looked at the pawn VM in detail

Great work

rsKliPPy:
@IgnacioFDM You can easily add new parameters to a native, the native just has to check how many arguments were passed to it and take action based on that.
[Here](https://github.com/alliedmodders/amxmodx/blob/2eeeb2997d4f5acd0e6364ab62a1e3a117f69dff/amxmodx/datastructs.cpp#L353) is an example of a native that introduced a third parameter in 1.8.3.

Arkshine:
Well, a global config doesn't sound a good idea. Actually, there are not many natives which need attention, so having an extra param (yes, we can add param without breaking bcomp) should be enough.

Natives:
- ~~`strtok`~~ (_deprecated_)
- `strtok2`
- `argparse`
-  ~~`strbreak`~~ (_deprecated_)
- `format_args`
- `parse`: you can't sanely add a param for that one because it can take a variable number of params, you would need probably a new natives.

Stocks:
- `argbreak`
- `is_str_num`

Arkshine:
A little update. 

For this PR, updating essentially natives related to the case.

I was wondering if it was necessary to support Unicode whitespaces for natives relying on `isspace`, but this doesn't seem really useful for us. If someone really needs that, we could always support them later. 

The first post has been updated. I will run more tests later. If you see something missing or want to give your thoughts, feel free to say it.

Arkshine:
I'm redoing the PR. Not happy with it as it, messy and this needs extra attention.

PR #364, opened 2016-03-21
Remove Non-GPL matches

Ni3znajomy:
This PR removes all code related to nongpl_matches.

##### Reasons
1. I believe this feature was introduced due to response Live's GunGame AMXX plugin. Since the code has been reproduced by sawce (https://forums.alliedmods.net/showthread.php?t=53249) and Live's work is most just copy&paste, I'm sure no one uses his GunGame today anymore.
2. This feature has not been updated with any new closed source plugins since 2007.
3. It's almost impossible to add **every** closed source plugin to block list. 
4. People should be self-aware about installing closed source plugins, which can harm their servers.


dvander:
I don't object to removing this code, but #3 and #4 aren't good reasons. SourceMod has a similar mechanism that can be updated via a downloaded database, and this has been used in the past to disable malware out of band with SourceMod releases. It's our only tool to combat malicious behavior (other than the DMCA), and entries are not added lightly.

So - removing this because it's unused is one thing, but I would support AMX Mod X reviving it to have the same abilities as SourceMod.

Arkshine:
^ I agree too. I don't see any valid reasons to remove such tool _now_, especially when AMXX is not going to die soon and a new version is coming. Also, since we introduced gamedata, it's actually planned to use a similar mechanism as SourceMod, with an auto-updater as well. Insertion will be easier and it will be definitively proved to be handy in the future.

Ni3znajomy:
It would be great if AMXX had same/similar mechanism as SM (database). If you want to keep this code for now I think this PR can be closed.

PR #366, opened 2016-03-28
Extend "read_argv" native

Ni3znajomy:
~~This PR introduces a new native ```read_argv2```.~~
This PR extends ```read_argv``` native. It works like ```read_data``` native. We can get integer/float value directly instead writing it to an array then executing ```str_to_num``` or ```str_to_float``` natives.

### DIFF
--- a/amxmodx/amxmodx.cpp
+++ b/amxmodx/amxmodx.cpp
@@ -2435,7 +2435,19 @@ static cell AMX_NATIVE_CALL read_argv(AMX *amx, cell *params) /* 3 param */
 	int argc = params[1];
 
 	const char *value = g_fakecmd.notify ? (argc >= 0 && argc < 3 ? g_fakecmd.argv[argc] : "") : CMD_ARGV(argc);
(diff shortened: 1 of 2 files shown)


Nextra:
If we want this, is it even necessary to introduce a new native for this? Feels like `read_argv` can just be extended in a backwars-compatible way (or alternatively add two type-safe variants for int/float).

Ni3znajomy:
If ```read_argv``` can be extended that's ok. I just wasn't sure about it. New commit arrived.

WPMGPRoSToTeMa:
Why do you return float as additional param? You can just return float as return value and use `any` return tag for `read_argv`.

Nextra:
Then it will assign the float value to an int without the compiler making you aware of a type mismatch. It will not do what you want it to do, and thus you'd still need an extra parameter.

Arkshine:
It will be for sure more convenient in some situations. Code looks good and backward compatibility should be kept. Thanks!

PR #367, opened 2016-03-31
Fix typos in entity_intersects() and get_global_edict2() natives

Arkshine:
* `entity_intersects`: Return value is reversed. Two years later no one reported the issue \o/. Related  to #70.
* `get_global_edict2`: By using `FNullEnt` this ignores worldspawn edict, which should be allowed here. `edict_to_id` returns  -1 if pointer is null. Related  to #250.


Nextra:
Urgh on the first one.
The second one seems like it has been "wrong" all along?

Arkshine:
Likely we confused about what does really `FNullEnt` at some point.

IgnacioFDM:
>Two years later no one reported the issue \o/

Lol, tbh I didn't know such native existed and I'm using the same function implemented in pawn

PR #368, opened 2016-04-04
cmd_targetex

luxxxoor:
This is a feature stolen from SM, and it will introduce multi-targets commands.

There is a list of all new targeting types : @all        : target all players from server. 
@alive      : target all alive player from server.
@aim        : target the player that the admin is currently looking at.
@bots       : target only bots.
@dead       : target only dead players. 
@humans     : target only human players.
@me         : target the admin that write the command. 
@!me        : target everyone in the server except the admin that use the command. 
@ct         : target anyone from CT team.
@t          : target anyone from TERRORIST team.
@spec       : target anyone from SPECTATOR team.
@spectating : target the player that the admin is currently spectating.

AMXX is getting very old and we still don't have a multi targetting option... so there is my suggestion.

### DIFF
--- a/plugins/admincmd.sma
+++ b/plugins/admincmd.sma
@@ -185,38 +185,70 @@ public cmdKick(id, level, cid)
 	if (!cmd_access(id, level, cid, 2))
 		return PLUGIN_HANDLED
 
-	new arg[32]
+	new arg[32], players[MAX_PLAYERS], lang[32];
 	read_argv(1, arg, charsmax(arg))
-	new player = cmd_target(id, arg, CMDTARGET_OBEY_IMMUNITY | CMDTARGET_ALLOW_SELF)
+	new plnum = cmd_targetex(id, arg, CMDTARGET_OBEY_IMMUNITY | CMDTARGET_ALLOW_SELF | CMDTARGET_SAFETY, players, lang, charsmax(lang))
 	
-	if (!player)
+	if (!plnum)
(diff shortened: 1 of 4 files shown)


AngeIII:
You have some commented ```//server_print("cmd_targetex : ```

luxxxoor:
i know that there are some bugs.. Artick(HamletEagle) startet the stock.. i tried to finish it.. but my time is not my friend... so here was our "project" we don't have time to finish it.. so ark please take it just as suggestion

Javivi:
@luxxxoor What are the bugs and what is missing. If you don't say anything it is harder for someone to take over and finish this.

luxxxoor:
at the moment it has only one known bug: if you'll use a command (ex: amx_slay) @all and cmd_targetex has CMDTARGET_NO_SPEC flag, it will work as CMDTARGET_ONLY_ALIVE, because i cannot find a proper way to implement that feature

Javivi:
And what would be the correct behavior ? Targeting both alive and dead players, but not spectators ?

luxxxoor:
yes

i think the name of flags are sufficiently suggestive, or they are not ?

Javivi:
But in certain cases, like with amx_slay, it would not make any sense targeting dead players. 

What's sourcemod behaviour ?

luxxxoor:
it should not be designed only for base commands from admincmd

Arkshine:
Closed in favor of #515.

PR #369, opened 2016-04-05
Values by reference in forwards

Ni3znajomy:
This PR allow us to pass a value to forwards by reference. (In AMXX and its modules we pass pointer)

We have new param types:
For AMXX Core and its modules: ```FP_CELL_BYREF``` and ```FP_FLOAT_BYREF```
For AMXX Plugins: ```FP_VAL_BYREF```

```FP_VAL_BYREF``` is the synonym of ```FP_CELL_BYREF``` because int and floats values are handled in the same way in plugins. It may be confusing for people when they notice there's only ```FP_CELL_BYREF``` and there isn't ```FP_FLOAT_BYREF```, so I came with some generic name to refer both int and floats.

I've also done 2 test plugins to test this functionality. Some feedback would be nice. :smiley:

### DIFF
--- a/amxmodx/CForward.cpp
+++ b/amxmodx/CForward.cpp
@@ -59,7 +59,7 @@ cell CForward::execute(cell *params, ForwardPreparedArray *preparedArrays)
 			if (pDebugger)
 				pDebugger->BeginExec();
 			
-			// handle strings & arrays
+			// handle strings & arrays & values by reference
 			int i;
 			
 			for (i = 0; i < m_NumParams; ++i)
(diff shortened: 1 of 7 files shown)


WPMGPRoSToTeMa:
Why you have a difference between `cell` and `real` references?

Ni3znajomy:
We pass REAL* to ```cell executeForwards(int id, ...)``` and then we casting it to cell. If I didn't check param's type in ```[CForward/CSPForward]::execute``` every param "by reference" would be casted to cell*. Maybe this isn't necessary but I did it for consistency.

Arkshine:
You could achieve the same with FP_ARRAY, but it's definitively more convenient having new FP_.
I'm okay with FP_CELL_BYREF and FP_FLOAT_BYREF for consistency.
It looks okay, I guess you did test it anyway since you provided a test plugin.
Thanks!

PR #370, opened 2016-04-07
Fix a crash after cs_set_user_model is used with model_index parameter set.

Arkshine:
Related to #255.
Reported by PartialCloning on the [forum](https://forums.alliedmods.net/showthread.php?t=281270).

Allocating the new model in the engine is required.


Javivi:
Looks good to me

PR #372, opened 2016-05-04
Fix compilation error for GCC 6

Ni3znajomy:
Fix compilation error due ```-Wmisleading-indentation``` warning. It was added to ```-Wall``` in GCC 6.


Arkshine:
Thanks!

PR #373, opened 2016-05-11
Update mapsmenu.sma

luxxxoor:
Essentially:

- Old Menus were replaced with New Menus.
- Most of old natives like 'format' were replaced with more specific natives like 'formatex'/'add'.
- Added 'Search' feature for both menus but it will reduce the number of items per page to 6 (mapmenu) or 5 (votemapmenu).
- Added 'Unselect' feature for votemapmenu, item will be now colored in red, not disabled like before (that was needed, it was very annoying when you would select a map, but you selected another because you could not unselect and you were forced to reopen the menu)

Hope you like it, this plugin is a core amxx plugin and because of inexistent feature of 'unselect' for votemapmenus is needing a rework.
Plugin was tested, in my opinion is fully functionally.
Just test it and enjoy the new features.

### DIFF
--- a/plugins/mapsmenu.sma
+++ b/plugins/mapsmenu.sma
@@ -14,499 +14,612 @@
 #include <amxmodx>
 #include <amxmisc>
 
-new Array:g_mapName;
-new g_mapNums
-new g_menuPosition[MAX_PLAYERS + 1]
+new Array:g_mapName; 
+new g_mapNums;
 
-new g_voteCount[5]
+new g_voteCount[5];
 
-new g_voteSelected[MAX_PLAYERS + 1][4]
-new g_voteSelectedNum[MAX_PLAYERS + 1]
+new g_voteSelected[MAX_PLAYERS + 1][4];
+new g_voteSelectedNum[MAX_PLAYERS + 1];
 
-new g_coloredMenus
-
-new g_choosed
+new g_choosed;
 
 public plugin_init()
 {
-	register_plugin("Maps Menu", AMXX_VERSION_STR, "AMXX Dev Team")
-	register_dictionary("mapsmenu.txt")
-	register_dictionary("common.txt")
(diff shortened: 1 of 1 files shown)


GoRiLliAz:
Hello

I tried no problem.
Added a cvar to choose the number of map in the votemap would be good.
Put a different fichier.ini for the votemap and the mapmenu, personally I have a vip and I give access only votemap and they have not any map.

luxxxoor:
You can edit the plugin for your preferences, the termen of "vip player" means nothing for amxx core plugins, so i can not help you with that.
About the your ideas.. you can request that after the accepting that pull request if it will be accepted

GoRiLliAz:
If you can avoid the need of recompiling the plugin, it should be done, unless there is a good reason to hardcode.

The VIP was just an example. You have only a single file for the mapchooser and the mapmenu. Having some config for that is more than welcomed.

In my opinion, it's best to do all the changes at once.

luxxxoor:
idk if those sugestions are productive... 
let's just wait for another opinions, right ?

plugins/mapsmenu.sma
```diff
@@ -14,499 +14,612 @@
...
-	new len = format(menuBody, charsmax(menuBody), g_coloredMenus ? "\y%L\R%d/%d^n\w^n" : "%L %d/%d^n^n", id, "VOTEMAP_MENU", pos + 1, (g_mapNums / 7 + ((g_mapNums % 7) ? 1 : 0)))
-	new end = start + 7, keys = MENU_KEY_0
-
-	if (end > g_mapNums)
-		end = g_mapNums
-
-	new tempMap[32];
-	for (new a = start; a < end; ++a)
+	new menuBody[64], tempMap[32];
+	formatex(menuBody, charsmax(menuBody), "\y%l", "CHANGLE_MENU");
```
all85100, line 203: formatex(menuBody, charsmax(menuBody), "\y%L", LANG_PLAYER, "VOTEMAP_MENU");

The menu name is displayed incorrectly

PR #374, opened 2016-06-06
Fix UTIL_ReplaceAll not properly tracking length.

Arkshine:
Same as https://github.com/alliedmodders/sourcemod/pull/518

> UTIL_ReplaceEx returns a pointer to the character after the newly-replaced stuff. While maxlength is the size of the original buffer, the remaining space left needs to get reduced by how far forward we jumped after a replace happens.

Unlike SM, it doesn't impact the native (`replace_string`) which uses this function due output is copied again with proper max length, but still fixing this function is welcomed.

### DIFF
--- a/amxmodx/util.cpp
+++ b/amxmodx/util.cpp
@@ -459,11 +459,14 @@ unsigned int UTIL_ReplaceAll(char *subject, size_t maxlength, const char *search
 	size_t searchLen = strlen(search);
 	size_t replaceLen = strlen(replace);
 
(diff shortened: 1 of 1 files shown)


Ni3znajomy:
Looks good. 👍

Nextra:
If I'm not mistaken this bug would cause the function to basically ignore the maxlength in certain situations, so `replace_string` should be affected. Still fine to :ship: as it's an obvious bug (regression even?)

PR #375, opened 2016-06-14
Fixed some natives descriptions

voed:


Arkshine:
Thanks!

Nextra:
The copy pasta is strong with this one. Good spot :)

voed:
Also, documentation needs update: http://www.amxmodx.org/api/amxmodx/is_linux_server
pawn-docgen does not removing exist data from database, as i know

Nextra:
Parser needs to be run manually I think.

PR #377, opened 2016-06-26
Extend amxx list command

rsKliPPy:
This commit extends `amxx list` or `amxx plugins` command to let users not just list currently loaded plugins but also display only those ones that match given search criteria.
It searches if a given string is found within the plugin title, file name or author string.


Arkshine:
Small QoL for servers having a lot of plugins. Searching on title, file name, and finally author seems to be logical. It looks good. Thanks!

PR #378, opened 2016-06-27
Add interface API

rsKliPPy:
This PR adds an interface API for modules that's similar to SourceMod's.

Module communication was never really encouraged, and module writers almost never provided an interface to their module. The only exceptions I saw are nVault and SQL.
Previous interface exposure was done through `MF_RegisterFunction` and `MF_RequestFunction`. There was no standard way of exposing interfaces and modules could pass through `MF_RegisterFunction` whatever they wanted, so you would have to know specifics for each module's interface. For example, nVault exposes a function that returns an nvault manager, while SQL returns a structure through `MF_RequestFunction`, which is quite ugly and confusing.

New functions are `MF_AddInterface` and `MF_RequestInterface`.

`MF_AddInterface` accepts an `AMXXInterface` pointer and makes AMXX aware of the given interface, so other modules can request it.
`MF_RequestInterface` requests an interface from AMXX, but also checks for version and compatibility between versions. Compatibility check can be overriden for custom versioning, just like in SM.

The goal is to move towards better cross-module communication, so not as much code will have to be reimplemented. For example, if someone wants their module to make use of sockets, they have to make their own cross-platform implementation, while it would be easier if the Sockets module exported an interface for it.

All suggestions are welcome, please discuss.

Also, two test modules that showcase the functionality are attached below. Sorry for the messy VS project hierarchy. :blush:

[AMXXInterfaceTest.zip](https://github.com/alliedmodders/amxmodx/files/335577/AMXXInterfaceTest.zip)


Arkshine:
I don't have much to say since we talked about it together. I think it's a good first step for a better communication and will definitely ease further upcoming change.

Having a message on failure could be useful so the admin knows what's happening. I'm not sure the ideal place, inside RequestInterface seems okay.

If @Nextra could share some of his wisdom on this PR, that would be appreciated.

dvander:
For interfaces like this, you should make sure AMX Mod X has two-stage shutdown. Like, a "pre-shutdown" callback so interfaces can negotiate releasing shared resources, then the actual shutdown. Otherwise you get into a scenario where module A shuts down, then module B shuts down, but module B tries to free something in module A.

rsKliPPy:
I would guess that [OnPluginsUnloaded](https://github.com/rsKliPPy/amxmodx/blob/module-interfaces/amxmodx/meta_api.cpp#L1634) is good enough for that, as it's called right before modules are detached, and by the way, OnPluginsLoaded seem to be the best place to request interfaces.

What do you think?

rsKliPPy:
Updated this PR.

Added a new function to API: `MF_RequestLibrary` which requests the specified module (library) to load, similar to what plugins do.
This function isn't intended for "wide" use, but is actually used behind the scenes in `AMXX_Attach`, unless someone wants an optional dependency. The reason behind it is because `OnAmxxAttach` was never made in a way that a coder can return an `AMXX_*` status value from it, so adding a return value would possibly break older modules (or it may not, I need someone to clarify).
The workaround consists of a new array named `MODULE_DEPENDENCIES` in `moduleconfig.h`, where one can list libraries that are required for this module to be run.

Two test modules in OP updated to showcase new changes.

PR #382, opened 2016-08-14
Allow execute forwards without need to create variable for returned value

Ni3znajomy:
Just little enhancement that allows us to execute forwards without need to create a variable for returned result, it's convenient in some situations when we don't care about it (```ET_IGNORE```).

So we can do:
```c
new fwd_handle = CreateMultiForward("test_forward", ET_IGNORE);
ExecuteForward(fwd_handle);
```
instead of:
```c
new fwd_handle = CreateMultiForward("test_forward", ET_IGNORE);
new dump;
ExecuteForward(fwd_handle, dump);
```

### DIFF
--- a/plugins/include/amxmodx.inc
+++ b/plugins/include/amxmodx.inc
@@ -3139,7 +3139,7 @@ native PrepareArray(const array[], size, copyback = 0);
  * @note Passing arrays requires them to be prepared using PrepareArray()
  *
  * @param forward_handle    Forward handle
- * @param ret               Variable to store return value in
+ * @param ret               Optional variable to store return value in
  * @param ...               Variable number of parameters to pass through
  *
  * @return                  1 on success, 0 if forward can't be executed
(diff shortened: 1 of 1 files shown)


Arkshine:
Kippy was lazy to say it, but If you set a default value to a reference, you would want to check in the code this value for 0 in order to avoid bad thing (like overwriting data): https://github.com/alliedmodders/amxmodx/blob/master/amxmodx/amxmodx.cpp#L4204

WPMGPRoSToTeMa:
> Kippy was lazy to say it, but If you set a default value to a reference, you would want to check in the code this value for 0 in order to avoid bad thing (like overwriting data): https://github.com/alliedmodders/amxmodx/blob/master/amxmodx/amxmodx.cpp#L4204

@Ni3znajomy for example see `parse_loguser` https://github.com/alliedmodders/amxmodx/blob/732a05dde0bbcce5446e0cc0b8297a2c6eba33f1/amxmodx/amxmodx.cpp#L3128

rsKliPPy:
It's all good, I failed, what I said to AssShine wasn't correct. I just analysed the assembly once again to make sure.
At first I thought that PAWN VM pushed 0 to the stack as an address, but it actually pushed the address of the newly allocated 0 value. So there is no danger of data being overwritten, it will just change value of a temporarily allocated variable on the heap.

Arkshine:
Alright, thanks for the small QoL change.

PR #383, opened 2016-08-14
Fix huge read_argv regression

IgnacioFDM:
Introduced in https://github.com/alliedmodders/amxmodx/pull/366

This reverts commit aaa2934.

This broke binary compatibility with either older, already compiled
plugins, or newly compiled plugins if you forgot to update .inc headers.

This happened because read_argv used to receive maxlen by value, and
after this commit it receives it by reference. This causes read_argv either
to fail, or worse, to buffer overflow, resulting in a security vulnerability.

Newly introduced functionality for read_argv should be added with a new,
separate native.

### DIFF
--- a/amxmodx/amxmodx.cpp
+++ b/amxmodx/amxmodx.cpp
@@ -2460,19 +2460,7 @@ static cell AMX_NATIVE_CALL read_argv(AMX *amx, cell *params) /* 3 param */
 	int argc = params[1];
 
 	const char *value = g_fakecmd.notify ? (argc >= 0 && argc < 3 ? g_fakecmd.argv[argc] : "") : CMD_ARGV(argc);
-
-	switch (*params / sizeof(cell))
-	{
-		case 1:
-			return atoi(value);
-		case 3:
-			return set_amxstring_utf8(amx, params[2], value, strlen(value), *get_amxaddr(amx, params[3]));
-		default:
-			cell *fCell = get_amxaddr(amx, params[2]);
-			REAL fparam = static_cast<REAL>(atof(value));
-			*fCell = amx_ftoc(fparam);
-			return static_cast<int>(fparam);
-	}
+	return set_amxstring_utf8(amx, params[2], value, strlen(value), params[3]);
 }
 
 static cell AMX_NATIVE_CALL read_args(AMX *amx, cell *params) /* 2 param */
--- a/amxmodx/amxmodx.cpp
+++ b/amxmodx/amxmodx.cpp
@@ -2463,6 +2463,35 @@ static cell AMX_NATIVE_CALL read_argv(AMX *amx, cell *params) /* 3 param */
 	return set_amxstring_utf8(amx, params[2], value, strlen(value), params[3]);
 }
 
+static cell AMX_NATIVE_CALL read_argv_int(AMX *amx, cell *params) /* 1 param */
+{
+	int argc = params[1];
+
+	if (argc <= 0)
+	{
+		return 0;
+	}
+
+	const char *value = g_fakecmd.notify ? (argc >= 1 && argc < 3 ? g_fakecmd.argv[argc] : "") : CMD_ARGV(argc);
+
+	return atoi(value);
+}
+
+static cell AMX_NATIVE_CALL read_argv_float(AMX *amx, cell *params) /* 1 param */
+{
+	int argc = params[1];
+
+	if (argc <= 0)
+	{
+		return 0;
+	}
+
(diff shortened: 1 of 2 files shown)


IgnacioFDM:
To be honest this kind of stuff even makes you wonder if the PR was malicious considering this "bug" allows you to buffer overflow.

Arkshine:
Unless I'm drunk, it's correct. Parameter is "..." so values are passed by reference and therefore you need to use `get_amxaddr()`.

IgnacioFDM:
@Arkshine 
Ok, 50% my bad then, didn't know that about "...". 

What happens is this PR 
- Breaks binary compatibility with all already compiled plugins, because they will not send the maxlen by ref
- Breaks binary compatibility with newly compiled plugins if you forget to update the .inc as I did

Anyway this results during runtime in either
- ```maxlen``` will be a huge, erroneous number, therefore maxlen check is bypassed and you can buffer overflow
- ```maxlen``` is so big it because negative and read_argv outright returns a wrong result. This is how I noticed, I updated amxmodx without recompiling any plugin and some clcmd that used read_argv broke.

Sorry for causing some excess panic but this is still really bad and should be reverted because anybody running a new amxmodx with old already compiled plugins (or who didn't update .inc) will either have broken plugins or worse, be vulnerable.

IgnacioFDM:
By they way (obviously) don't accept this PR since it will break the "improved" read_argv. I think the old PR should be reverted, and some new native added to have this functionality.

Nextra:
And this is why you stay technical and to-the-point in your PRs. Thanks for finding this issue but we can do without the attitude. Furthermore with the knowledge that @Arkshine gave you it should become obvious that your own PR is just as broken as #366, making it so *newer* plugins will not run properly with this code. The old PR needs to be re-done.

IgnacioFDM:
Yes, sorry, I was too cocky in the OP. Anyway, as I mentioned above, this PR breaks the new read_argv

IgnacioFDM:
I added the revert commit in case you want to go that route.

Ni3znajomy:
I'll never ever include any bug intentionally, I always try to propose changes that will be useful for others but there aren't always successful as we see. Some things have been explained already (about ```...```), so I think I'm not longer a "bad guy" 😈. Maybe it's time for ```read_argv2```?

IgnacioFDM:
Yes I'm very sorry for being so paranoid. My sincere apologies.

Ni3znajomy:
Since code looks ok, just needs to be as another native (as I understand), is it something wrong if this could be done in this PR?

IgnacioFDM:
Yes, sure. If you want open another PR and I'll close this one, or PR my repo or whatever. I don't feel like deciding the name for the new native and whatnot, I leave that to you guys :stuck_out_tongue_closed_eyes:

Ni3znajomy:
@IgnacioFDM merged my commit to his branch, currently this PR restore old behavior of ```read_argv``` and add ```read_argv2``` native. Furthermore I marked ```read_argv``` as deprecated in favor of ```read_argv2```. Give your thoughts about that deprecation. If you don't mind it I'll push a commit that updates plugins to use it.

Arkshine:
I'm not sure it should be deprecated because `read_argv` is not broken by any means. `read_argv2` would be just an extended version, this doesn't mean `read_argv` can't be used.

Nextra:
Or a dedicated variant each for float and int, similar to how our newer APIs value type-safe natives.

IgnacioFDM:
I like the dedicated variant more

Arkshine:
That would be ideal.

IgnacioFDM:
There we go, I merged and rebased @Ni3znajomy new natives into this PR

Arkshine:
For consistency, we could have `read_argv_string` but making an alias of `read_argv` is kind of redundant, or deprecating it would not make much sense. 

Code looks good. Unless someone has something to say, let's go with that.

IgnacioFDM:
:shipit:

Ni3znajomy:
I consider usage of them to get command i.e. ```read_argv_int(0)``` (same for float). I'm not sure that functionality is needed since commands are usually string. I believe these natives should be used to get command's arguments rather than the command itself but if such feature can be useful in some cases then 🚢.

Arkshine:
@Ni3znajomy Documentation says index starts to 1. If it happens someone uses 0, nothing bad would happen. But checking for 0 makes sense for those natives and there is no reason to avoid sanity check anyway, there is no harm to add them.

IgnacioFDM:
Merged @Ni3znajomy changes

PR #384, opened 2016-08-19
Update hashers

Ni3znajomy:
PR updates [Portable C++ Hashing Library](http://create.stephan-brumme.com/hash-library) to version 7 (from 5?).
Current version of library has a bug that causes returning incomplete hash for SHA3 & Keccak 224 bits.
This update fixes it.


Arkshine:
Thanks! 

Related to #206.

PR #385, opened 2016-08-27
use exponentiation by squaring instead of n-times multiplications

alldroll:
Replace linear algorithm with exponentiation by squaring 

**Computational complexity**: O(log n) instead of O(n)


Arkshine:
This seems to be a micro-optimization for a pawn usage, but there is no harm and is welcomed to improve code if it's faster.

Code looks good, thanks for your contribution!

PR #386, opened 2016-09-08
add levenshtein distance

alldroll:
The games are place where developer has deal with natural language. It would be nice to have tools for processing it. In this PR was implemented algorithm `Levenshtein distance`
 with memory usage `O(min(n, m)`) and time complexy `O(mn)`. 
There is list of tools which i am going to implement (If natural language processing tools are welcomed):
* `Soundex`
* `Metaphone`
* Improve Levenshtein distance (`Damerau–Levenshtein distance`)
* `Prefix tree (trie)`

### DIFF
--- a/amxmodx/string.cpp
+++ b/amxmodx/string.cpp
@@ -1424,6 +1424,52 @@ static cell AMX_NATIVE_CALL fmt(AMX *amx, cell *params)
 	return 1;
 };
 
+/**
+ *
+ */
+static cell AMX_NATIVE_CALL levenshtein(AMX *amx, cell *params)
+{
+	int a_len, b_len, i, j, prev, tmp;
+	const char *a = get_amxstring(amx, params[1], 0, a_len);
+	const char *b = get_amxstring(amx, params[2], 1, b_len);
+
+	if (a_len == 0) {
+		return b_len;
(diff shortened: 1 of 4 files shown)


rsKliPPy:
The usage of these algorithms is very specific and, in my opinion, they are well out of AMXX's scope.

What you could do is implement that in Pawn and post it in Code Snippets/Tutorials on the forum, or if you really prefer to do it in C++, possibly using some libraries for these algorithms to make it easier for you, implement it in a module and post it on the forums. I'm sure someone will make use of it.

IgnacioFDM:
I think this could be a nice feature

Imagine the following

```amx_kick arksihne```

```Did you mean "amx_kick Arkshine"? ```
```Type "amx_kick %" to confirm```

```amx_kick %```
```Arkshine was kicked from the server```

rsKliPPy:
Maybe if we had a better target filter (see #368 ) so it works on all commands, but just as a native on its own I don't think it currently has place in AMXX. If a PR extends all AMXX commands to use this new feature, then probably yes.

alldroll:
We can use this tools for detecting swear words. But @rsKliPPy is right, this should be a module or plugin.

PR #387, opened 2016-09-24
Revert access check for say command

voed:
Server owners reported that admins without ADMIN_PASSWORD flag can use "say @" command. I proposed them to revert access check back to 1.8.2 code. Works fine, but its not a solution of problem. We need to fix flag manager.

### DIFF
--- a/plugins/adminchat.sma
+++ b/plugins/adminchat.sma
@@ -69,9 +69,9 @@ public plugin_cfg()
 	}
 }
 
-public cmdSayChat(id, level)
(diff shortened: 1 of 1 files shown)


Arkshine:
Can you elaborate more please?

ClaudiuHKS:
I don't see anything wrong.
It's obvious that admins without the **Admin_Password** flag can use **say @**.

voed:
I already forgot about this lol
I will test later, but definitly something was wrong

PR #389, opened 2016-10-29
Add wrappers for natives with string flags

rsKliPPy:
This PR adds wrappers for some of the existing natives that have flags parameters as strings, so now they have properly named constants as flags. This increases code readability and maintainability, but you also won't ever have to take a look at the documentation once you decide to use get_players. :yum:

I would like some feedback though:
- ~What should I do with pause() and unpause() natives? They are kind of... ugh.~
- The documentation of new functions - is it alright or do I need to add anything else? I am not entirely happy with it right now.
- Is the *_ex suffix alright or should new functions be named different?
- Are there any other natives that need such wrappers?
- Should set_task_ex() and get_players_ex() be moved into amxmodx.inc, or even be created into natives like the other two (which I had to because of varargs)? I'm not sure if amxmisc.inc is the best place for them.

### DIFF
--- a/amxmodx/amxmodx.cpp
+++ b/amxmodx/amxmodx.cpp
@@ -8,6 +8,7 @@
 //     https://alliedmods.net/amxmodx-license
 
 #include <time.h>
+#include <amtl/am-utility.h>
 #include "amxmodx.h"
 #include "CMenu.h"
 #include "newmenus.h"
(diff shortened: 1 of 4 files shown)


WPMGPRoSToTeMa:
> Is the *_ex suffix alright or should new functions be named different?

`GetPlayers` and `SetTask` :open_mouth:

Arkshine:
It looks okay now. Let's go with that.

PR #391, opened 2016-11-24
Add missing format param

Ni3znajomy:
Quick fix that adds missing format parameter in ```next_hudchannel``` native.


Arkshine:
Thanks!

PR #392, opened 2016-12-09
Fixed is_visible native

WPMGPRoSToTeMa:


Arkshine:
Reported by Depresie here https://forums.alliedmods.net/showthread.php?t=290962&highlight=is_visible.
Issue is about native not working on player because traceline hits the target's head before the end point.
Proposed fix is to disable entities collision check (which makes sense for this kind of native).

@WPMGPRoSToTeMa Does it happen only on player? I tried to dig the engine code source but could not really figure out something.

WPMGPRoSToTeMa:
@Arkshine no, not only on player. It happens on all solid entities (except triggers), because we can't ignore both entities via ignoreEnt TraceLine parameter, we can ignore only one entity looker or target. To ignore the remain entity we can make it non-solid temporarily.
If we ignore only looker entity then trace will hit bbox or hitboxes of target entity before it reach the center of target entity (which is always inside target's bbox and almost always inside some hitbox).

rsKliPPy:
What if instead of checking the fraction we check if the desired entity is hit? This could fail if origin + view_ofs isn't within some hitbox, but not sure how probable that is.
Maybe even a new optional parameter to switch between multiple methods of checking with the possibility of adding more methods in the future?

WPMGPRoSToTeMa:
@rsKliPPy not instead but both should be good:
```if (tr.flFraction == 1.0 || tr.pHit == pTarget)```

Arkshine:
If we can avoid the hacky way, that would be welcomed.

WPMGPRoSToTeMa:
But this hack is better for TraceLine performance, because target's hitboxes won't be calculated (bones positions and rotations) and checked (there is additional minitrace for each hitbox).

Arkshine:
Well, can't deny that. For this kind of native, this would make the most sense, but at the same time the less safe backward-compatibility speaking. 

Do we have plugin hooking pfnTraceLine directly and checking solid state or whatever there? Probably not, especially when this situation requires special modules and native was partially broken, so I guess the hacky way would be safe to be used. Anyone see some potential issue with it?

IgnacioFDM:
Kind of related to this: Does anybody know about why ```get_user_aiming``` sometimes returns "phantom results", returning as a result a player who isn't actually in front of you (but somewhere else in the map).

PR #397, opened 2017-01-05
Fix the modules/mysqlx/AMBuilder not building on Visual Studio 2015

evandrocoan:
Applied the same fix as in:

Fix building of MySQL DBI extension on VS 2015+ by psychonic:
https://github.com/alliedmodders/sourcemod/commit/94383d995da8f7c208b7060eaccdb3ae73b93163

Originally reported in:
https://forums.alliedmods.net/showthread.php?t=292365


evandrocoan:
You are welcome.

PR #400, opened 2017-01-10
Fix missing CS_NORESET constant

Arkshine:
Added in #287, but has disappeared somewhere...


WildCard65:
I found where it went missing: https://github.com/alliedmodders/amxmodx/commit/d3e5957215b7f3cb75aaffd92fd12529a4776c42

PR #401, opened 2017-01-17
Fix UTF-8 character parsing due to improper casting

souvikdas95:
Upcasting of -ve signed characters (char) to signed integers (cell) results in -ve signed integers which don't represent valid printable characters.
eg. UTF-8 Character '中' (0xE4 0xB8 0xAD) when casted results in 0xFFFFFFE4 0xFFFFFFB8 0xFFFFFFAD which are not valid printable characters.
Simple Characters like '4' (0x34) don't get affected because they are +ve signed characters.


Arkshine:
Similar issue as #23.

Second commit is out of scope of this PR, you should make another PR with it.

PR #403, opened 2017-01-17
Add detail to 'VOTE_SUCCESS' & 'VOTE_FAILED' in mapsmenu plugin

souvikdas95:
1. Details for Vote Success & Failed is useful for transparency and fairness in voting.
2. Patch Language file "mapsmenu.txt" according to parameters specified.


plugins/lang/mapsmenu.txt
```diff
@@ -217,8 +217,8 @@ CHANGLE_MENU = Verander Map Menu
 [es]
 RESULT_REF = Resultado rechazado
 RESULT_ACC = Resultado aceptado
-VOTE_SUCCESS = Votacion finalizada. Se va a cambiar el mapa a
-VOTE_FAILED = Votacion sin exito
+VOTE_SUCCESS = Votacion finalizada (Tiene: %d) (Necesario: %d) Se va a cambiar el mapa a '%s'.
```
Javivi: Did you pull the translations from google translate ?

(%d/%d) would be better than adding the wrong translation

rsKliPPy: Agreed. Change that and it should be ready to go.

Arkshine: French translation is also weird. I think you should just add only the translations you know for sure. The missing keys can be dealt later in a global language update or something.

Freeman-AM:
@souvikdas95, any news about the state of that PR concerning requested changes ?

PR #404, opened 2017-01-21
Fix client_death forward called on kill command sent by a spectator

WPMGPRoSToTeMa:
This bug can be tested with [Ultimate Gore plugin](https://forums.alliedmods.net/showthread.php?p=19346).

### DIFF
--- a/modules/cstrike/csx/meta_api.cpp
+++ b/modules/cstrike/csx/meta_api.cpp
@@ -125,17 +125,27 @@ const char* get_localinfo( const char* name , const char* def = 0 )
 	return b;
 }
 
+static bool ClientKill_wasAlive;
+
+void ClientKill(edict_t *pEntity)
+{
+	CPlayer *pPlayer = GET_PLAYER_POINTER(pEntity);
+	ClientKill_wasAlive = pPlayer->IsAlive();
+
+	RETURN_META(MRES_IGNORED);
+}
+
 void ClientKill_Post(edict_t *pEntity)
 {
 	CPlayer *pPlayer = GET_PLAYER_POINTER(pEntity);
-	if ( pPlayer->IsAlive())
-		RETURN_META(MRES_IGNORED);
-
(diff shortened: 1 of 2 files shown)


modules/cstrike/csx/meta_api.cpp
```diff
@@ -125,17 +125,25 @@ const char* get_localinfo( const char* name , const char* def = 0 )
 	return b;
 }
 
-void ClientKill_Post(edict_t *pEntity)
+static bool ClientKill_wasAlive;
```
souvikdas95: Why use static? Isn't it global already?

WPMGPRoSToTeMa: For file locality

Arkshine:
If I understand well, the bug is because of the silly engine which doesn't check `deadflag` in `pfnClientKill` and because a spectator is given 1 HP by the game, `pfnClientKill` is unnecessarily fired, then since csx module doesn't check for spectator, `client_death` is fired as well.

Proposed fix looks fine. What about having something similar to the game `IsObserver` and checking for `pev->iuser1`, wouldn't it be enough?

WPMGPRoSToTeMa:
@Arkshine gamedll sets iuser1 for killed player immediately after death.
https://github.com/s1lentq/ReGameDLL_CS/blob/5044d4ed131d683cd38a3728941723aa352bcbdb/regamedll/dlls/player.cpp#L2160

Arkshine:
Looking at the code, this is not guaranteed to be set right away, so I guess the current fix is fine then. Thanks.

Arkshine:
@WPMGPRoSToTeMa are you alive?

WPMGPRoSToTeMa:
@Arkshine yep.

Arkshine:
Don't you see the review above? It misses a `RETURN_META(MRES_IGNORED);`. If you don't provide it metamod will throw an unnecessary warning, if I'm not wrong.

WPMGPRoSToTeMa:
@Arkshine thanks! What review do you mean? I see only this review.
default

Arkshine:
It looks like I did not understand it was necessary to 'submit the review' once you finished to comment. :P

PR #405, opened 2017-01-25
Fixed round stats resetting

WPMGPRoSToTeMa:
Now round stats are resetted only on round start (as mentioned in documentation), not on player spawn. Also round stats now can't be resetted via `fullupdate` command.


souvikdas95:
Really? mazdan? -_-

modules/cstrike/csx/meta_api.cpp
```diff
@@ -81,18 +81,18 @@ struct sUserMsg
...
+	{"Damage",     &gmsgDamageEnd,  Client_Damage_End, true},
+	{"WeaponList", &gmsgWeaponList, Client_WeaponList, false},
+	{"HLTV",       &gmsgHLTV,       Client_HLTV,       false},
+	{"AmmoX",      &gmsgAmmoX,      Client_AmmoX,      false},
+	{"ScoreInfo",  &gmsgScoreInfo,  Client_ScoreInfo,  false},
+	{"AmmoPickup", &gmsgAmmoPickup, Client_AmmoPickup, false},
+	{"SendAudio",  &gmsgSendAudio,  Client_SendAudio,  false},
+	{"TextMsg",    &gmsgTextMsg,    Client_TextMsg,    false},
+	{"BarTime",    &gmsgBarTime,    Client_BarTime,    false},
+	{"DeathMsg",   &gmsgDeathMsg,   Client_DeathMsg,   false},
```
Nextra: Without judging the rest of this patch, you should update your editor settings so stuff like this doesn't happen. These are unnecessary changes.

WPMGPRoSToTeMa:
This PR is very questionable because DM servers already use reset on spawn behaviour. This would break it. We can also add cvar for this and/or autodetect CSDM, but we need normal Spawn hook, because `ResetHUD` hook is not good.

souvikdas95:
yeah

PR #407, opened 2017-02-11
Improve UTF-8 support in some natives (bug 6475)

Arkshine:
This extends UTF-8 support for some natives. It relies on [UTF-8 Rewind](https://bitbucket.org/knight666/utf8rewind), which is small, cross-platform, easy-to-use, well-tested C library. And enough for our needs.

UTF-8 is improved in the following natives:
- `containi`
- `strfind` _(with ignorecase set)_
- `strcmp` _(with ignorecase set)_
- `strncmp` _(with ignorecase set)_
- `equali`
- `replace_string` _(with ignorecase set)_
- `replace_stringex` _(with ignorecase set)_
- `get_players` _(with name and ignorecase flags set)_
- `find_player` _(with name and ignorecase flags set)_

For safety and/or compatibility, the following natives are added:

- `mb_strotolower`
- `mb_strtoupper`
- `mb_ucfirst`

Finally, while we are on it and because the library has already the corresponding internal functions:

- `mb_strtotitle` _self-explanatory_
- `is_string_category` _allows to check if the input string conforms to the category specified by flags, this is meant to replace `is*` natives_.

Worth to be noted:
- Improving utf-8 in natives should be fine as it was never really supported before.
- Stocks and constants from `string.inc` have been moved to their own includes.
- All natives have been tested.


Arkshine:
@Nextra @rsKliPPy if you could take a look, thanks.

Arkshine:
@rsKliPPy @IgnacioFDM  @Ni3znajomy @Nextra Can someone do some fast review please?

IgnacioFDM:
Looks OK to me. By the way reviewing this I noticed in ```UTIL_TrimRight``` the ```for``` loop is nonsensical

rsKliPPy:
Looks good now. :ship:

PR #408, opened 2017-02-19
Fix a potential crash in ArrayClone for too large arrays

Arkshine:
If there is not enough memory to clone an array, throw an error instead of potentially crashing the server.

Related to f78fda6d9c995eced7d6a28f8f2096f21b2fd49a.


rsKliPPy:
:ship:

PR #412, opened 2017-02-20
Add RequestFrame() native

rsKliPPy:
This makes it easy to offset execution to the next frame without manually hooking StartFrame or using other workarounds to do so.
```
/**
 * Creates a single use hook for the next frame.
 *
 * @param callback      Function to be executed on the next frame.
 * @param data          Optional data to be passed to the callback function.
 *
 * @note                Callback function prototype:
 *                          function(data)
 *
 * @noreturn
 */
native RequestFrame(const callback[], any:data = 0);
```

Note: It may be better to hook `void _Host_Frame(float time)` and execute frame hooks in there. Also possibly provide an easy way for other modules to hook that too in the future. I would like feedback on this matter.

Test plugin provided in *plugins/testsuite/request_frame_test.sma*.

### DIFF
--- a/amxmodx/CFrameAction.h
+++ b/amxmodx/CFrameAction.h
@@ -3,6 +3,7 @@
 
 #include "amxmodx.h"
 #include <amtl/am-deque.h>
+#include <amtl/am-autoptr.h>
 
 class CFrameActionMngr
 {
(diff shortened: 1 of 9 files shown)


WPMGPRoSToTeMa:
Can you provide use-cases for this?

rsKliPPy:
SourceMod had RequestFrame() native for a long time, so I thought why wouldn't we have it too.

Currently no useful examples on my mind.

Arkshine:
I know plugins and I've also encountered such situations where you need to postpone at least the next frame to do stuffs. Especially true when you alter the game. I know some plugins which use the trick with `FM_StartFrame` (registering and unregistering in the hook) and also using `set_task`.

If we can have a readable, clean and precise way to do that, it's welcomed I guess.

IgnacioFDM:
I've seen many plugins indeed using set_task for this purpose. I'd argue using set_task this way, FM_StartFrame or this new function is never the correct way, but sometimes it's "useful" when you are too lazy to investigate what the correct hook is.

WPMGPRoSToTeMa:
I've remembered one situation where it can be used. You can use it for getting total money change after frame, like killing two players at once with awp.

PR #414, opened 2017-02-22
Add client_remove() forward and fix a bug with client not internally disconnected

Arkshine:
Related to #264.

- Added `client_disconnected_post` forward to be the post version of  `client_disconnected`, called when a player is really disconnected internally. Having a post forward is useful if you use natives counting players like `get_playersnum` or `get_players` relying on the internal state in the core or simply when you want to do stuff on disconnected players.
- Fixed a bug introduced in the related PR where a client would not be properly disconnected internally in AMXX.

About the first commits, long short story, I wanted to follow proper semantic, that's it, a pre-forward called `client_disconnecting` (client is still valid), and a post-forward `client_disconnected` (client is no more valid) ; but considering the original PR has been merged quite a while, that would cause likely issues on current plugins and that it's impossible to measure how much plugins are using `client_disconnected`, I changed my mind at the end. It's kind of silly, but I should have reacted sooner.

Feel free to propose another name  instead of `client_disconnected_post`.

### DIFF
--- a/amxmodx/meta_api.cpp
+++ b/amxmodx/meta_api.cpp
@@ -135,6 +135,7 @@ int FF_ClientCommand = -1;
 int FF_ClientConnect = -1;
 int FF_ClientDisconnect = -1;
 int FF_ClientDisconnecting = -1;
+int FF_ClientDisconnected = -1;
 int FF_ClientInfoChanged = -1;
 int FF_ClientPutInServer = -1;
 int FF_PluginInit = -1;
@@ -497,6 +498,7 @@ int	C_Spawn(edict_t *pent)
 	FF_ClientConnect = registerForward("client_connect", ET_IGNORE, FP_CELL, FP_DONE);
 	FF_ClientDisconnect = registerForward("client_disconnect", ET_IGNORE, FP_CELL, FP_DONE);
 	FF_ClientDisconnecting = registerForward("client_disconnecting", ET_IGNORE, FP_CELL, FP_CELL, FP_ARRAY, FP_CELL, FP_DONE);
+	FF_ClientDisconnected = registerForward("client_disconnected", ET_IGNORE, FP_CELL, FP_CELL, FP_ARRAY, FP_CELL, FP_DONE);
 	FF_ClientInfoChanged = registerForward("client_infochanged", ET_IGNORE, FP_CELL, FP_DONE);
 	FF_ClientPutInServer = registerForward("client_putinserver", ET_IGNORE, FP_CELL, FP_DONE);
 	FF_PluginCfg = registerForward("plugin_cfg", ET_IGNORE, FP_DONE);
@@ -687,8 +689,15 @@ void C_ServerDeactivate()
 
 		if (pPlayer->ingame)
 		{
+			auto wasDisconnecting = pPlayer->disconnecting;
+
 			pPlayer->Disconnect();
 			--g_players_num;
+
+			if (DropClientDetour && !wasDisconnecting)
+			{
+				executeForwards(FF_ClientDisconnected, static_cast<cell>(pPlayer->index), FALSE, prepareCharArray(const_cast<char*>(""), 0), 0);
+			}
 		}
 	}
 
@@ -880,7 +889,14 @@ void C_ClientDisconnect(edict_t *pEntity)
 		--g_players_num;
 	}
 
+	auto wasDisconnecting = pPlayer->disconnecting;
+
 	pPlayer->Disconnect();
+
+	if (!wasDisconnecting)
+	{
+		executeForwards(FF_ClientDisconnected, static_cast<cell>(pPlayer->index), FALSE, prepareCharArray(const_cast<char*>(""), 0), 0);
+	}
 
 	RETURN_META(MRES_IGNORED);
 }
@@ -895,7 +911,7 @@ DETOUR_DECL_STATIC3_VAR(SV_DropClient, void, client_t*, cl, qboolean, crash, con
 	ke::SafeVsprintf(buffer, sizeof(buffer) - 1, format, ap);
 	va_end(ap);
 
-	CPlayer *pPlayer;
+	CPlayer *pPlayer {};
 
 	if (cl->edict)
 	{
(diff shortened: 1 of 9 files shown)


IgnacioFDM:
I think ```client_disconnected_post``` is OK (although in general I quite dislike the idea of forwards that register automatically with their hardcoded name, if we went back in time I'd like ```plugin_*``` to be the only forwards that automatically register)

In-line:
`client_disconnect, client_disconnected, client_disconnected_post, client_disconnected_post_ex`
Hm..

Arkshine:
Not totally sure what you're trying to say.

`client_disconnect` is deprecated.

With this PR, we will have:
`client_disconnected`
`client_disconnected_post`

Not idea why you're referring `client_disconnected_post_ex`.

WPMGPRoSToTeMa:
`player_disconnecting`
`player_disconnected`
as aliases, or SourceMod style (CamelCase)
`ClientDisconnecting`
`ClientDisconnected`

rsKliPPy:
`ClientDisconnecting` and `ClientDisconnected` seems fine to me.

IgnacioFDM:
I think it'd be best to follow the current "convention" of client_*. Even though there already are huge inconsistencies in naming, there's no need to further diverge. I'd love new naming conventions, but only as long as they were applied throughout all functions. I much prefer consistency over style.

rsKliPPy:
The CamelCase convention was being used for years now, influenced by SM probably. Every API that was added to 1.8.3 is named using CamelCase.
I know this is kind of different since we are building on and expanding something older, and I know there is a concern that it might be confusing for users, but I think we should just use CamelCase for everything we implement in the future.
I wouldn't even be against aliasing plugin_init to PluginInit, but that might be going too far.

Nextra:
This should definitely stay with `client_` as a tag, we'll have to live with the inconsistencies. In fact I'd rather alias the new-ish APIs to snake_case than changing the entire core to use CamelCase, but we shouldn't do either.

The fact that we already "shipped" `client_disconnected` when it might be better to call it `client_disconnecting` is a bit inconvenient, I agree. As I'm not particularly fond of the `_post` I offered the alternative of naming it `client_remove` as a paring to `client_putinserver` or maybe `client_drop`. In the end it doesn't matter too much I guess.

Arkshine:
I agree we should keep `client_* naming`. I'm fine with `client_remove` which seems better suited.

WPMGPRoSToTeMa:
@Arkshine so, `client_removing` and `client_removed`?

Arkshine:
Nope, what is done is done. AMXX is full of inconsistencies and trying to fix it for everything will only get worse (legacy oblige). It's shame I did not react sooner, but it's too late, so let's focus on only the post forward.

If no one complain about `client_remove`, I will settle with that.

FrictionSource:
It's expected that things will change in the beta state. You shouldn't be using the beta otherwise.

Go for client_disconnecting and client_disconnected (for post), otherwise you're unnecessarily complicating the names by overthinking the consequences.

According to this: http://www.amxmodx.org/newstats.php?mod_id=0&addon_id=1 only a small percentage would be effected.

Nextra:
1.8.2 is over 4 years old. The relevant commit here alone is nearing 2 years old. You don't break backward compatibility after this amount of time just because of slightly unfortunate naming. That we have to argue this basic stuff in so many PR discussions seriously boggles the mind. Pointless bikeshedding all around. The fact that you signed up just to post this crap seriously amazes me.

Arkshine:
@FrictionSource , What you say is true,  but that would have been totally fine if changes were relatively recent, but this one is here for more than 1,5 year. You can be sure people are using it. What is worse is that would change subtly the behavior of the forward, I'm not sure it's worth to put all such trouble on the admins now.

In-line:
@FrictionSource I totally agree with you.

Arkshine:
Guys, even if it's the dev version, because of AMXX is a special state where development is slow and people have started to use it since a long time ago, you can't just change something which breaks the compatibility/behavior if you did something there is almost two years ago. AMXX is stable because backward compatibility is always kept. Breaking this now would be irresponsible, to be honest.

IgnacioFDM:
I agree with Arkshine. Lots of people use the "dev" version as the stable version because there have been many critical fixes since 1.8.2, and 1.8.2 is severely outdated. The dev version really is the much more stable version.

WPMGPRoSToTeMa:
Maybe we should investigate how much plugins are using pre-behaviour of client_disconnected?

Nextra:
🤦

Arkshine:
I really appreciate the feedbacks but this is going to be difficult to measure the impact nor it seems worth to waste our time on that matter. We are talking about just naming, at this point, it's starting to be silly. 

We are not going to argue forever about this. Again, what you guys said is true and it would have really fine if that change was not already that old. Even if dev and if there are not many people using it, this is still wrong to suddenly introduce a breakage in a change made years ago. 

It's was a mistake, we will need to live with it. It's okay. Next time we will have to be more careful.
Let's move on, alright?

WPMGPRoSToTeMa:
@Arkshine Ok.

luxxxoor:
this name mistake is very confusing... it's like naming get_user_name with get_user_naming
why can't we just can't find something that is not so confusing ? 
a name of a native should tell the essence, where is the essence when you call it disconnected when in purpose is called when disconnecting

CliendDisconnecting/ed would be the best... but if not we should search for something else, look my ideea:
cliect_game_removing
client_game_removed

WPMGPRoSToTeMa:
Some offtopic: can someone explain what is the difference between client, player and user in AMXModX/GoldSrc context?

rsKliPPy:
I believe there's no difference, only different naming schemes, unfortunately. Generally it looks like `client` is more used for public forwards and `user` for natives. It does suck.

Nextra:
As far as I am concerned:

Client: The game client running on the PC
Player: The representation of the client on the server
User: The person sitting in front of the PC

It isn't consistently used in AMXX at all, so you can pretty much use them interchangably.

FrictionSource:
If changing it is out of the question then I would have to agree that client_remove or client_removed makes the most sense. You just have to think of it as the last forward for a client, when they're completely removed from the game rather than just a post disconnect hook. I think if you called it client_remove from the beginning no one would have even complained as they wouldn't be thinking of it as a post version of client_disconnected.

rsKliPPy:
I proposed `client_invalidated` but it seems kind of strange for AMXX. Thoughts?

PR #417, opened 2017-03-03
Add basic ReHLDS and ReGameDLL support

Arkshine:
This PR is about adding a basic support for ReHLDS and ReGameDLL by using the provided API in order to get a direct access to the engine/mod functions/global variables that we're using in the core and counter-strike module, that's it without relying on signatures/symbols/hack from game data files. 

This PR is motivated by the following reasons:
- ReHLDS and ReGameDLL are solid projects meant to be used in production, providing more stable binaries than the original, with many fixes, security fixes, improvements and an API for mods/plugins. Performance optimization as well. The projects exist since a few years now, and I believe they are fairly mature, enough to get some attention and to consider a support in AMXX.
- It's a fact, there are more and more admins switching to those projects, and we can see everytime people asking how to update the game data and what to replace, which is not an easy task, especially when some function can't be replaced. If we can allow AMXX to work as it is in others contexts, that would be welcomed and actually, is a really wanted feature.
- Last, probably the most important and decisive point which has motivated me to do something. Someone created a fork of AMXX (based on old dev), named ReAmxModx by adding Re* support and supposedly adding improvements and fixes, but it's _closed source_. Quite absurd, isn't?
I tried to talk as maintainer to the author to see if we could work together and he refused. For some reasons he doesn't want to share the source code, even the bug found. This makes no sense, AMXX being a public and open-source project, and the end this is to help the community. 
This can't be helped to find such behavior quite suspicious (cf the latest news, right), even though it's probably not the case here, but still, we have no way to verify what's happening inside. 
This is even more worrying because of admins being redirecting to this fork (which doesn't seem to be updated often too) due to the lack of support in the original version.  Just this point would be an enough reason to provide a support for Re* API to avoid people being stuck to use some shady fork.

Note: Changes are straight. This doesn't use special wrapper class or similar as we don't have much stuff yet but it happens we use more stuff from engine/mod, that would be likely welcomed. Cstrike module is bit ugly but I guess it's okay for now. A refactoring can be made another time.


IgnacioFDM:
Amazing stuff. I've been waiting for this for a long time. About the closed source ReAmxModX, it should also be mentioned that it's a violation of the GPL.

modules/fakemeta/fakemeta_amxx.cpp
```diff
@@ -13,12 +13,16 @@
 
 #include "fakemeta_amxx.h"
 #include "sh_stack.h"
+#include <reapi/mod_regamedll_api.h>
```
theAsmodai: Maybe will be better name it 'resdk/mod_regamedll_api.h' to avoid confusion with https://github.com/s1lentq/reapi?

Arkshine: I had this thought too but forgot about it. I'm okay with `resdk`, will change it soon.

Adidasman1:
@IgnacioFDM, source code Re AMX Mod X, was provided personally Arkshine.

IgnacioFDM:
@Adidasman1 It's still a violation of the GPL. You have to make the source code available to **every** user of ReAMXModX (that is, anybody who downloads the binary). That's the point of the GNU GPL license.

Anyway, let's not derail this PR.

modules/fakemeta/fakemeta_amxx.cpp
```diff
@@ -61,19 +65,26 @@ void OnAmxxAttach()
...
 	}
 
-	void *address = nullptr;
-
-	if (!CommonConfig->GetAddress("g_pGameRules", &address) || !address)
+	if ((HasRegameDll = RegamedllApi_Init()))
 	{
-		MF_Log("get/set_gamerules_* natives have been disabled because g_pGameRules address could not be found. ");
-		return;
+		GameRulesRH = ReGameApi->GetGameRules();
```
s1lentq: At this stage, GetGameRules will return nullptr, but we can register a hook [InstallGameRules](https://github.com/s1lentq/reapi/blob/adb34bb0ed46c1ccd21e0ba5ab8bfcaae9db8116/reapi/src/main.cpp#L93) to get the g_pGameRules pointer.

Arkshine: Unless I'm wrong, this should be fine. `AMXX_Attach` is called at world spawn, and `g_pGameRules` is initialized at world precache, so it should not be null. I think I've tested and I did want to add more code than necessary.

S0m3Th1nG-AwFul:
I'm currently testing AMXX build based on this PR since 4.03.2017 on latest ReHLDS+ReGameDLL and it seems to be stable enough for me :)

modules/fakemeta/fakemeta_amxx.cpp
Arkshine: Looks like you were right! Not sure why I thought it was working.

rsKliPPy:
> A refactoring can be made another time.

:ship:

Arkshine:
Alright. It should be good now. I had not the chance to test again without Re* enabled with all the latter changes. Fixed few typos.

Let's go with it.

PR #419, opened 2017-03-04
Introduce a datapack position tag

Arkshine:
Same as https://github.com/alliedmodders/sourcemod/pull/363, full description there.
It happened today on the forum and I just remembered SM did such modification.


rsKliPPy:
🚢

PR #421, opened 2017-03-10
Update Fun module

Arkshine:
This is essentially a code clean up following with some welcomed changes:
- Updated documentation (made by @rsKliPPy)
- Toggled internal `PlayerPreThink` (used with `get/set_user_footsteps`) and `TraceLine` (used with `get/set_user_hitzones)` forwards on-demand (in others words, those forwards won't be hooked if no plugins are using such natives or if no players are actually set by them)
- Added `HITZONE_*` constants for `get/set_user_hitzones`.
- Made `TraceLine` a post forward.  Reason: as it is it breaks plugins hooking `TraceLine` because of the original game call is being superceded and other modules can't catch it. It looks like it's this way from the very start fun module has been introduced 13 years ago. Fakemeta module comes a little later. Change should be safe 1) it's unlikely having both plugins, one using `set_user_hitzones` and another hooking `TraceLine` as the latter would not work properly 2) This native is rarely used. 

To do:
- [x] Testing more

### DIFF
--- a/modules/fun/fun.cpp
+++ b/modules/fun/fun.cpp
@@ -40,28 +40,9 @@
 	}
 */
 
-char g_bodyhits[33][33];	// where can the guy in the first dimension hit the people in the 2nd dimension? :-)
-bool g_silent[33];			// used for set_user_footsteps()
-
 HLTypeConversion TypeConversion;
+CPlayer Players[kMaxClients + 1];
 
-// ######## Utils:
-void FUNUTIL_ResetPlayer(int index)
-{
-	//MF_PrintSrvConsole("Resetting player index %d! maxclients: %d\n", index, gpGlobals->maxClients);
-	for (int i = 1; i <= gpGlobals->maxClients; i++) {
-		g_bodyhits[index][i] = (char)((1<<HITGROUP_GENERIC) | 
-						(1<<HITGROUP_HEAD) | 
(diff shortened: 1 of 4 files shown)


modules/fun/fun.cpp
```diff
@@ -11,562 +11,442 @@
...
-/******************************************************************************************/
+
 void PlayerPreThink(edict_t *pEntity)
 {
-	if (g_silent[ENTINDEX(pEntity)]) {
+	auto index = TypeConversion.edict_to_id(pEntity);
+
+	if (Players[index].HasSilentFootsteps())
+	{
 		pEntity->v.flTimeStepSound = 999;
```
In-line: I think it's more correct to hook `PM_PlayStepSound`

Arkshine: Yes. I could be wrong, but so far I remember, the best method to silent _totally_ a player's footsteps is to stop `PM_PlayStepSound` execution (a good method which doesn't require gamedata is to hook `PM_Move` and set `runfuncs` to 0, I believe), then to stop the sound on your client, you need to send an updated `SVC_NEWMOVEVARS` message (this requires to hook `SV_WriteMovevarsToClient` as well).

Is it worth? I'm not sure. It could be if we had a native which can exploit this method to set options that we can't usually have per player (airaccelerate, edgefriction, roll, sky?, etc.).

PR #422, opened 2017-03-11
Rework build pathname functions

Ni3znajomy:
This PR makes ```build_pathname*``` functions use AMTL's ```ke::path::Format``` and ```ke::path::FormatVa```. Code is shorter and cleaner inside those functions now.

Also editor removed white spaces in edited files, hope you don't mind that 😇, but if you do, i'll create commit without those unrelated changes.

### DIFF
--- a/amxmodx/amxmodx.h
+++ b/amxmodx/amxmodx.h
@@ -105,10 +105,8 @@ extern AMX_NATIVE_INFO g_GameConfigNatives[];
 
 #if defined(_WIN32)
 	#define PATH_SEP_CHAR		'\\'
-	#define ALT_SEP_CHAR		'/'
 #else
 	#define PATH_SEP_CHAR		'/'
-	#define ALT_SEP_CHAR		'\\'
 #endif
 
 #ifndef GETPLAYERAUTHID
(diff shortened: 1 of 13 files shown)


amxmodx/amxmodx.h
```diff
@@ -105,10 +105,8 @@ extern AMX_NATIVE_INFO g_GameConfigNatives[];
 
 #if defined(_WIN32)
```
Arkshine: You can probably remove those defines now, we have a `PLATFORM_` version.

Ni3znajomy:
Commit divided into smaller ones. Now it should be easier to review.

amxmodx/amxmodx.h
```diff
@@ -96,19 +96,17 @@ extern AMX_NATIVE_INFO g_GameConfigNatives[];
...
-#if defined(_WIN32)
+#if defined PLATFORM_WINDOWS
 	typedef HINSTANCE DLHANDLE;
 #else
 	typedef void* DLHANDLE;
 	#define INFINITE 0xFFFFFFFF
 #endif
 
-#if defined(_WIN32)
+#if defined PLATFORM_WINDOWS
```
Arkshine: As said in the previous review, you can remove those constants, they are redundant with `PLATFORM_` version

Ni3znajomy: They're gone now

Arkshine:
Alright, it should be okay now, thanks!

PR #427, opened 2017-03-26
Port code to AMTL

Ni3znajomy:
First bunch of patches which port code to use AMTL.
The point of this PR is to drop some old cruft and replace it with more modern AMTL. It's mainly focused on removing CList dependency. Also removes some duplicated and unused files.

### Summary
- Removed CQueue class (unused)
- Removed sh_string (unused)
- Removed files sh_list.h & sh_stack.h in amxmodx/ (both of them exist in public/)
- Removed sm_queue (replaced with ke::Deque in NVault module)
- Removed CList class, got replaced with ke::Vector and ke::LinkedList
- Moved sh_tinyhash.h from amxmodx/ to public/


IgnacioFDM:
Vectors in most cases are much much faster than LinkedLists, I haven't specifically checked in which situations you are using LLs though.

https://www.youtube.com/watch?v=YQs6IC-vgmo

Ni3znajomy:
I've done some testing and works ok.

Vector has been used where elements are just added, i.e. CTask, ClEvent, ForceObject and for func_s. 

While LinkedList has been used where elements are added but also removed, I'm aware of that we can also remove from Vector but it shifts all elements while LinkedList does not. But if it doesn't affect performance much or less than LinkedList itself executing malloc/free per node, then I think LinkedList could be replaced with Vector entirely.

Arkshine:
LinkedList has been made for keeping compatibility. Ideally when possible using InlineList which doesn't malloc(), or Vector. But it's fine to use LinkedList ultimately if the context of allocation performance is not a factor or because you don't have the choice.

Not a requirement, but while you're on it, we can always try to improve things.

WPMGPRoSToTeMa:
```C++
template <typename T>
using LinkedList = Vector<T>;
```
😄 But they have different interface...

Ni3znajomy:
AMTL has been updated, update fixes pre increment InlineList's iterator.
LinkedList has been replaced using InlineList and Vector.

IgnacioFDM:
In practice Linked Lists are still slower than vectors for removal operations in 99% of cases (basically for any non big structure), even if they are "faster" in terms of big O. A simple benchmark will show this. And that's even considering the fact that AMTL's remove is horribly slow (compared to STL) if you have to remove multiple elements because it doesn't support the erase remove idiom (or equivalent, at least last time I checked).

Anyway, it doesn't really matter since those code sections aren't performance critical. In my opinion the biggest amxmodx bottleneck is executeforward, which I've been meaning to optimize for a while, and even then, you can only "feel it" when using things like pfnAddToFullPack with lots of ents and such. These things are just a drop in the bucket.

That said, changes look OK (though haven't tested in a server).

Arkshine:
@IgnacioFDM Please do it. Small cumulative improvements here and there, can only lead globally to a better product.

amxmodx/CEvent.cpp
```diff
@@ -292,16 +292,16 @@ void EventsMngr::parseValue(int iValue)
 
 	// loop through the registered funcs, and decide whether they have to be called or not
 	// if they shouldnt, their m_Done is set to true
-	for (ClEventVecIter iter = m_ParseFun->begin(); iter; ++iter)
+	for (auto &event : *m_ParseFun)
```
In-line: Reallocation = crash

amxmodx/CEvent.cpp
```diff
@@ -386,16 +386,16 @@ void EventsMngr::parseValue(const char *sz)
...
```
In-line: Reallocation = crash

amxmodx/CEvent.cpp
```diff
@@ -455,22 +455,22 @@ void EventsMngr::executeEvents()
 	}
 
 	// Reset this here so we don't trigger re-entrancy for unregistered messages
-	ClEventVec *parseFun = m_ParseFun;
-	m_ParseFun = NULL;
+	auto parseFun = m_ParseFun;
+	m_ParseFun = nullptr;
 
-	for (ClEventVecIter iter = parseFun->begin(); iter; ++iter)
+	for (auto &event : *parseFun)
```
In-line: Reallocation = crash

PR #428, opened 2017-04-03
Fix signed/unsigned mismatch warning

IgnacioFDM:


Ni3znajomy:
These kind of warnings also occur in:
- https://github.com/alliedmodders/amxmodx/blob/master/modules/cstrike/cstrike/CstrikeItemsInfos.cpp#L106
- https://github.com/alliedmodders/amxmodx/blob/master/modules/cstrike/cstrike/CstrikeItemsInfos.cpp#L176
- https://github.com/alliedmodders/amxmodx/blob/master/amxmodx/file.cpp#L604
- https://github.com/alliedmodders/amxmodx/blob/master/amxmodx/CLang.cpp#L592

It'd be great if you could fix them all at once.

IgnacioFDM:
Weird, is that with GCC? Because I don't get them with VC++

Ni3znajomy:
Yes, GCC version 6.3.1

IgnacioFDM:
Haven't compiled with GCC and I'm not sure if these fixes would be "correct", so please review the changes

IgnacioFDM:
(forgot the file.cpp one)

Ni3znajomy:
Compiles without warnings 👍

PR #429, opened 2017-04-06
Fix inc/dec operators not returning an iEXPRESSION

Arkshine:
Similar to https://github.com/alliedmodders/sourcepawn/pull/108.

### Code example
```PAWN
function()
{
    new someInt[MAX_PLAYERS + 1] = {123, ...};
    new client;

    server_print("%d", ++someInt[client]);
    server_print("%d", someInt[client]++);
    server_print("%d", someInt[client]--);
    server_print("%d", --someInt[client]);
}
```

### Actual result
```
82
82
1694498816
20992
```

### Expected result
```
124
124
125
123
```
___
Asking you @dvander because even though it works, code is not totally the same as SM, so just in case.

---
This change is [Reviewable](https://reviewable.io/reviews/alliedmodders/amxmodx/429)

### DIFF
--- a/compiler/libpc300/sc1.c
+++ b/compiler/libpc300/sc1.c
@@ -5334,7 +5334,7 @@ static void doreturn(void)
     /* "return <value>" */
     if ((rettype & uRETNONE)!=0)
       error(78);                        /* mix "return;" and "return value;" */
-    ident=doexpr(TRUE,FALSE,TRUE,TRUE,&tag,&sym,TRUE);
+    ident=doexpr(TRUE,FALSE,TRUE,FALSE,&tag,&sym,TRUE);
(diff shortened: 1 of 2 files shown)


Arkshine:
It looks like inc/dec in `return` doesn't seem to work if compiled with optimization (from -d2), i.e.:
```PAWN
{
    new var;
    return ++var;
}
```
Essentially if the expression is after a `return`, the compiler omits to load the previously saved value from `load.s.pri`. 
Not totally sure where it should be fixed. If you have some idea david.

Arkshine:
Actually, the issue in "return" statement has been reported by you and has been fixed in 3.2.3664.

dvander:
:ship:

PR #431, opened 2017-04-24
Allow value be empty in lang files

In-line:
Prior 1.8.3 amxmodx was able to parse lang line with empty value like this
```
LANG_NAME = 
```
I have a few plugins that don't work correctly because amxmodx 1.8.3 consider  lang lines with empty value invalid. This commit should fix backward compatibility.


Arkshine:
A bit silly to allow empty value, but okay, compatibility is necessary. 
Thanks, it should be okay.

PR #433, opened 2017-04-29
Add or adjust a bunch of safety checks

Arkshine:
Probably not finished yet. Totally forgot about it I had that locally, so pushing here so I can remember and try to complete what I wanted to change later.

Feel free to point out any known and missing changes.


amxmodx/meta_api.cpp
```diff
@@ -1407,8 +1407,15 @@ int	C_Cmd_Argc(void)
 // Only	here we	may	find out who is	an owner.
 void C_SetModel(edict_t *e, const char *m)
 {
-	if (e->v.owner && m[7]=='w' && m[8]=='_' && m[9]=='h')
+	if (m[9] != 'h' || m[8] != '_' || m[7] != 'w')
```
WPMGPRoSToTeMa: There should be check for string length or better to check string fully.

amxmodx/meta_api.cpp
```diff
@@ -1407,8 +1407,15 @@ int	C_Cmd_Argc(void)
...
+	{
+		RETURN_META(MRES_IGNORED);
+	}
+
+	if (e && e->v.owner)
```
WPMGPRoSToTeMa: I think it is required that `e` is not `nullptr`, so if you want to check it, then better to make this as an assertion.

modules/cstrike/cstrike/CstrikeNatives.cpp
```diff
@@ -1957,7 +1957,7 @@ static cell AMX_NATIVE_CALL cs_get_user_weapon(AMX *amx, cell *params)
 
 	edict_t *pWeapon = TypeConversion.cbase_to_edict(get_pdata<void *>(pPlayer, m_pActiveItem));
 
-	if (!FNullEnt(pWeapon))
+	if (!FNullEnt(pWeapon) && pWeapon->pvPrivateData)
```
WPMGPRoSToTeMa: I think it is required in CS that all weapon entities have private data.

modules/cstrike/cstrike/CstrikePlayer.cpp
```diff
@@ -28,7 +28,7 @@ void ClientDisconnect(edict_t *pEntity)
 
 void ClientUserInfoChanged(edict_t *pEntity, char *infobuffer)
 {
-	if (pEntity->pvPrivateData)
+	if (pEntity && pEntity->pvPrivateData)
```
WPMGPRoSToTeMa: The same as for SetModel hook.

modules/cstrike/csx/meta_api.cpp
```diff
@@ -178,14 +178,18 @@ void PlayerPreThink_Post( edict_t *pEntity ) {
...
-		if (pPlayer->rank) pPlayer->Disconnect();
+
+	for( i = 1; i <= gpGlobals->maxClients; ++i)
+	{
+		GET_PLAYER_POINTER_I(i)->Disconnect();
 	}
-	if ( (g_rank.getRankNum() >= (int)csstats_maxsize->value) || ((int)csstats_reset->value == 1 ) ) {
-		CVAR_SET_FLOAT("csstats_reset",0.0);
+
+	if (int(csstats_maxsize->value) <= 0 || g_rank.getRankNum() >= int(csstats_maxsize->value) || int(csstats_reset->value) != 0)
```
IgnacioFDM: It'd probably be nicer to decide between (int)x, int(x) and C++ style casts and use throughout the code the same style of cast. It doesn't matter too much of course.

modules/fakemeta/pdata.cpp
```diff
@@ -192,30 +192,35 @@ static cell AMX_NATIVE_CALL set_pdata_string(AMX *amx, cell *params)
...
 		if (IsBadWritePtr(szData, 1))
 			return 0;
-		strcpy(szData, data);
 	} else {
 		szData = get_pdata<char*>(pEdict, iOffset);
 		if (IsBadWritePtr(szData, 1))
 			return 0;
-		if (params[4] == 1)
+
+		if (len > static_cast<int>(strlen(szData)))
```
IgnacioFDM: Such as here

modules/engine/engine.cpp
```diff
@@ -455,7 +455,12 @@ static cell AMX_NATIVE_CALL set_view(AMX *amx, cell *params) {
 			g_pFunctionTable_Post->pfnPlayerPostThink=PlayerPostThink_Post;
 
 			plinfo[iIndex].iViewType = CAMERA_3RDPERSON;
-			pNewCamera = CREATE_NAMED_ENTITY(MAKE_STRING("info_target"));
+
+			if (!(pNewCamera = CREATE_NAMED_ENTITY(MAKE_STRING("info_target"))))
```
WPMGPRoSToTeMa: Assertion should be better, it is not normal that entity is `nullptr`.

IgnacioFDM: I really like these kind of assignments since they save some unnecessary lines of code, but since it's too easy to confuse visually assignment and comparison I'm generally against using them and just assume every conditional with an assignment is a bug. I think some compilers or static analyzers even emit warnings unconditionally about them.

amxmodx/meta_api.cpp
WPMGPRoSToTeMa: There should be check for string length or better to check full string.

amxmodx/meta_api.cpp
WPMGPRoSToTeMa: I think it is required that `e` is not `nullptr`, so if you want to check it, then better to make this as an assertion.

modules/cstrike/cstrike/CstrikeNatives.cpp
WPMGPRoSToTeMa: I think it is required in CS that all weapon entities have private data.

modules/engine/engine.cpp
WPMGPRoSToTeMa: Assertion should be better, it is not normal that entity is `nullptr`. Or native error.

modules/cstrike/cstrike/CstrikePlayer.cpp
WPMGPRoSToTeMa: The same as for `SetModel` hook.

modules/engine/engine.cpp
WPMGPRoSToTeMa: Assertion should be better, it is not normal that entity is `nullptr`. Or native error.

amxmodx/meta_api.cpp
WPMGPRoSToTeMa: There should be check for string length or better to check full string.

amxmodx/meta_api.cpp
WPMGPRoSToTeMa: I think it is required that `e` is not `nullptr`, so if you want to check it, then better to make this as an assertion.

modules/cstrike/cstrike/CstrikeNatives.cpp
WPMGPRoSToTeMa: I think it is required in CS that all weapon entities have private data.

modules/cstrike/cstrike/CstrikePlayer.cpp
WPMGPRoSToTeMa: The same as for `SetModel` hook.

modules/engine/engine.cpp
WPMGPRoSToTeMa: Assertion should be better, it is not normal that entity is `nullptr`. Or native error.

amxmodx/util.cpp
```diff
@@ -272,7 +272,7 @@ void UTIL_ClientPrint(edict_t *pEntity, int msg_dest, char *msg)
 	if (pEntity)
 		MESSAGE_BEGIN(MSG_ONE, gmsgTextMsg, NULL, pEntity);
 	else
-		MESSAGE_BEGIN(MSG_BROADCAST, gmsgTextMsg);
+		MESSAGE_BEGIN(MSG_ALL, gmsgTextMsg);
```
In-line: Mb another commit ?

gtteamamxx: IS MSG_ALL safer than broadcast?

In-line: @gtteamamxx It uses reliable channel.

modules/cstrike/csx/meta_api.cpp
Arkshine: AFAIK, `int(x)` is not a C-style cast and should be the same as `static_cast<int>(x)`. I think both are fine, sometimes it feels appropriate to use `static_cast`, sometimes not. Well, I guess for consistency, we should probably always use `*_cast`.

amxmodx/meta_api.cpp
Arkshine: Yep, I agree.

amxmodx/meta_api.cpp
Arkshine: But `SetModel` is not necessary called from the mod, a plugin (very unlikely) or a module could call directly this function and passing null pointer for some reasons. This seems more proper to let the engine crashes?

WPMGPRoSToTeMa: We can use this check, but logically it should be an assertion.

modules/cstrike/cstrike/CstrikeUtils.h
```diff
@@ -45,7 +45,7 @@ void UTIL_StringToLower(const char *str, char *buffer, size_t maxlength);
 		return 0; \
 	} else { \
 		if (x <= gpGlobals->maxClients) { \
-			if (!MF_IsPlayerIngame(x)) { \
+			if (!MF_IsPlayerIngame(x) || !MF_GetPlayerEdict(x)->pvPrivateData) { \
```
In-line: Maybe separate error message when entity doesn't have private data?

Arkshine:
Rebase + Fixed based on comments. It should be enough for this PR.
Is everything look good?

amxmodx/amxmodx.cpp
```diff
@@ -792,6 +792,11 @@ static cell AMX_NATIVE_CALL is_user_alive(AMX *amx, cell *params) /* 1 param */
 
 	CPlayer* pPlayer = GET_PLAYER_POINTER_I(index);
 
+	if (!pPlayer->initialized)
```
justgo97: can't we just replace this with **pPlayer->ingame** check and remove the other one ?

modules/fakemeta/pdata.cpp
```diff
@@ -32,18 +32,18 @@
 //implement these with setjmp later.
 bool IsBadReadPtr(void *l, size_t size)
 {
-	return false;
+	return l ? false : true;
```
WPMGPRoSToTeMa: `l == nullptr`?

modules/fakemeta/pdata.cpp
```diff
@@ -192,30 +192,35 @@ static cell AMX_NATIVE_CALL set_pdata_string(AMX *amx, cell *params)
...
-			delete [] szData;
-			szData = new char[len + 1];
+			if (params[4] == 1)
+			{
+				free(szData);
+				szData = (char *)malloc(len + 1);
+			}
+			else if (params[4] == 2) {
+				delete[] szData;
+				szData = new char[len + 1];
```
WPMGPRoSToTeMa: Not related to this PR, but who will `free`/`delete[]` this data? Should we have some storage for allocated strings? (and free them on map change?)

Arkshine: I think the library should care for it automatically, no?
I don't mind a storage, but not sure if it makes actually a difference.

WPMGPRoSToTeMa: @Arkshine do you mean gamedll? I'm worry a bit about of compatibility between gamedll's `delete[]` and amxmodx's `delete[]`, but I hope they use the same since there is the `libstdc++` located in the server's root directory.

PR #349, opened 2016-02-04
Add a param to register_*cmd() and get_*cmd() to indiquate info is a multilingual key + modify plugin

Arkshine:
Small QoL. Motivation is to encourage the translation of plugin commands and to help admin/user to read such infos in their defined language. There is no reason to not translate them to the end-users.

Changes have been reflected on `cmd_access()` function and `amx_help` command. Not sure if I've forgotten others usage?

PR #434, opened 2017-05-01
Fixed bug in cmd_access with usage printing

WPMGPRoSToTeMa:
Bug appeared after #349.

How to reproduce:
```C++
register_clcmd("needtocheck", "NeedToCheck", -1, "- help me pls");

public NeedToCheck(player, level, cmdid) {
	if (!cmd_access(player, level, cmdid, 2))
		return PLUGIN_HANDLED;
	
	return PLUGIN_HANDLED;
}
```
Would just print:
`Usage:  `

`register_concmd` and `register_srvcmd` is ok since console cmds are both client and server cmds (`who = CMD_ServerCommand` for 0).


Arkshine:
Oops. Thanks.

PR #436, opened 2017-05-10
Fix #427, AMTL Code porting

In-line:
# TODO:

- [x] Fix crash in `CTaskMngr::startFrame()`
- [x] Fix crash in `EventsMngr::executeEvents()`
- [x] ~Check other cases?~ Checked, I didn't find something suspicious

### Explanation:
`CTaskMngr::startFrame()` iterates through `m_Tasks` vector and executes forwards if necessary, but forwards can also create new tasks, same goes for `EventsMngr::executeEvents()`. Everything will be fine before reallocation occurs, but if address changed then we will have crash (or more precisely undefined behavior). 
[Comment](https://github.com/alliedmodders/amtl/blob/c91e8560fb00984465a1a916172123b80a76dd04/amtl/am-vector.h#L202-#L203)  inside am-vector.h

### DIFF
--- a/amxmodx/CTask.cpp
+++ b/amxmodx/CTask.cpp
@@ -270,7 +270,7 @@ bool CTaskMngr::taskExists(int iId, AMX *pAmx)
 void CTaskMngr::startFrame()
 {
 	auto lastSize = m_Tasks.length();
-	for(auto i = 0; i < lastSize; i++)
+	for(auto i = 0u; i < lastSize; i++)
 	{
 		auto &task = m_Tasks[i];
 
--- a/amxmodx/CTask.cpp
+++ b/amxmodx/CTask.cpp
@@ -269,7 +269,8 @@ bool CTaskMngr::taskExists(int iId, AMX *pAmx)
 
 void CTaskMngr::startFrame()
 {
-	for(auto i = m_Tasks.length() - 1; i != 0; i--)
+	auto lastSize = m_Tasks.length();
+	for(auto i = 0; i < lastSize; i++)
 	{
 		auto &task = m_Tasks[i];
 
--- a/amxmodx/CTask.cpp
+++ b/amxmodx/CTask.cpp
@@ -269,8 +269,10 @@ bool CTaskMngr::taskExists(int iId, AMX *pAmx)
 
 void CTaskMngr::startFrame()
 {
-	for (auto &task : m_Tasks)
+	for(auto i = m_Tasks.length() - 1; i != 0; i--)
 	{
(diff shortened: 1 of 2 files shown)


amxmodx/CTask.cpp
```diff
@@ -269,8 +269,10 @@ bool CTaskMngr::taskExists(int iId, AMX *pAmx)
 
 void CTaskMngr::startFrame()
 {
-	for (auto &task : m_Tasks)
+	for(auto i = m_Tasks.length() - 1; i != 0; i--)
```
Arkshine: Actually, we want to keep the order that the tasks have been created. You have plugins creating multiple tasks at once, this will break this behavior. 

So, we need more looping from 0 to saved size:
```PAWN
auto size = static_cast<int>(m_Tasks.length());
for (auto i = 0; i < size; ++i) 
{
`

Comment (In-line) [inline amxmodx/CTask.cpp]: @Arkshine Do you think that they are plugins that rely on the order ? AMX Mod X task system is considered not precise. Also if new tasks inserted we will check them too (but time for task is minimum 0.1).

Comment (rsKliPPy) [inline amxmodx/CTask.cpp]: Your loop `i != 0` also skips the first task I believe.
It won't, it caches the length first, so newly appended tasks won't make it into the iteration. However, newly added tasks that occupy an old index will get processed but never executed, but your code (and the old one) have the same flaw so it's not a problem.

Comment (Arkshine) [inline amxmodx/CTask.cpp]: Yes, there are plugins relying on that order, we should always keep old behavior anyway.

Comment (In-line) [inline amxmodx/CTask.cpp]: 4a22686 fine ?

File: amxmodx/CTask.cpp [hunk 2]
```diff
@@ -269,8 +269,11 @@ bool CTaskMngr::taskExists(int iId, AMX *pAmx)
 
 void CTaskMngr::startFrame()
 {
-	for (auto &task : m_Tasks)
+	auto lastSize = m_Tasks.length();
+	for(auto i = 0; i < lastSize; i++)
```
Arkshine: `0u` was not a typo. `i` should be marked as 'unsigned', because `lastSize` has `size_t` type. You can cast `lastSize` to `int` as well but `u` is shorter.

In-line: @Arkshine Okey. `lastSize` eliminates the situation when we can have recursive calls. Is it okey ? Like doing `register_event` in `register_event` handler :)

Arkshine:
It should be good now and I believe it was the only places where vector could mutate. 

Thanks!

PR #438, opened 2017-05-14
Fix a engclient_cmd and amxclient_cmd buffer issue

Arkshine:
Related to 9c191949d80eb029431cdacefac59e70f7f80fe7.
Reported PartialCloning here: https://forums.alliedmods.net/showthread.php?t=296510.

This fixes a common static buffer issue where the original content is overwritten. It happens when a native relying on the static buffers calls a plugin callback, and inside that callback others natives relying on those static buffers are used as well.

Proposed fix is to copy the retrieved strings in order to not rely on the static buffers.
Since this is not a native called often, the use of AString should be fine.

.

### DIFF
--- a/amxmodx/amxmodx.cpp
+++ b/amxmodx/amxmodx.cpp
@@ -2626,32 +2626,40 @@ static cell AMX_NATIVE_CALL server_exec(AMX *amx, cell *params)
 	return 1;
 }
 
-int sendFakeCommand(AMX *amx, cell *params, bool fwd = false)
+int sendFakeCommand(AMX *amx, cell *params, bool send_forward = false)
 {
-	int ilen;
-	const char* szCmd = get_amxstring(amx, params[2], 0, ilen);
-	const char* sArg1 = get_amxstring(amx, params[3], 1, ilen);
+	enum args { arg_count, arg_index, arg_command, arg_argument1, arg_argument2 };
 
-	if (ilen == 0)
-		sArg1 = 0;
+	char command[128 * 2];
+	auto command_length = strncopy(command, get_amxaddr(amx, params[arg_command]), sizeof(command));
 
-	const char* sArg2 = get_amxstring(amx, params[4], 2, ilen);
+	if (!command_length)
+	{
+		return 0;
+	}
 
-	if (ilen == 0)
-		sArg2 = 0;
+	char argument1[128];
+	char argument2[128];
+	auto argument1_length = strncopy(argument1, get_amxaddr(amx, params[arg_argument1]), sizeof(argument1));
+	auto argument2_length = strncopy(argument2, get_amxaddr(amx, params[arg_argument2]), sizeof(argument2));
 
-	if (params[1] == 0)
+	const char *pArgument1 = argument1_length ? argument1 : nullptr;
+	const char *pArgument2 = argument2_length ? argument2 : nullptr;
+
+	int index = params[arg_index];
+
+	if (index == 0)
 	{
 		for (int i = 1; i <= gpGlobals->maxClients; ++i)
 		{
 			CPlayer* pPlayer = GET_PLAYER_POINTER_I(i);
 
 			if (pPlayer->ingame /*&& pPlayer->initialized */)
-				UTIL_FakeClientCommand(pPlayer->pEdict, szCmd, sArg1, sArg2, fwd);
+				UTIL_FakeClientCommand(pPlayer->pEdict, command, pArgument1, pArgument2, send_forward);
 		}
-	} else {
-		int index = params[1];
-
+	} 
+	else 
+	{
 		if (index < 1 || index > gpGlobals->maxClients)
 		{
 			LogError(amx, AMX_ERR_NATIVE, "Invalid player id %d", index);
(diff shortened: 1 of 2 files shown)


WPMGPRoSToTeMa:
Maybe better to modificate the `argv` field in `fakecmd_t`?

WPMGPRoSToTeMa:
I found a way how to get a bug with `engclient_cmd` 😄
```Pawn
#include <amxmodx>
#include <fakemeta>

public plugin_init() {
	register_clcmd("say /test", "Test");
	
	register_forward(FM_CVarGetFloat, "CVarGettingFloat");
}

public CVarGettingFloat(const name[]) {
	if (equal(name, "mp_forcecamera")) {
		// strcmp uses get_amxstring with id=1 for second arg
		// So, we change command
		// from "follow \"(1)[WPMG]PRoSToTeM@\""
		// to "follow \"[WPMG]PRoSToTeM@\""
		strcmp("follow", "[WPMG]PRoSToTeM@");
		//strcmp("", "[WPMG]PRoSToTeM@"); // also works, because cmd name is already checked
	}
}

public Test(player) {
	engclient_cmd(player, "follow", "(1)[WPMG]PRoSToTeM@");
}
```

[`follow` command](https://github.com/s1lentq/ReGameDLL_CS/blob/90d08ee3a043034fc69da86a56fd5a0d5491924c/regamedll/dlls/client.cpp#L2924-L2930)
[`GetForceCamera` call in `Observer_FindNextPlayer`](https://github.com/s1lentq/ReGameDLL_CS/blob/7c9d59101b67525a35b0b3a31e17159ab5d42fbd/regamedll/dlls/observer.cpp#L148)
[`CVAR_GET_FLOAT("mp_forcecamera")` in `GetForceCamera`](https://github.com/s1lentq/ReGameDLL_CS/blob/7c9d59101b67525a35b0b3a31e17159ab5d42fbd/regamedll/dlls/observer.cpp#L14)
[Checking name (from argv) after `GetForceCamera` call](https://github.com/s1lentq/ReGameDLL_CS/blob/7c9d59101b67525a35b0b3a31e17159ab5d42fbd/regamedll/dlls/observer.cpp#L169)

Arkshine:
Damn you!

Alright, updated a little. Looking a the engine, it looks like `argv` has actually a fixed size of 80, did the same.

Arkshine:
@WPMGPRoSToTeMa Does the patch looks good for you?

WPMGPRoSToTeMa:
@Arkshine it is ok, but buffers can be smaller, [because the engine uses only 127 bytes of received command](https://github.com/dreamstalker/rehlds/blob/f63bd849fa5b9e49e5ca836d8b65e2299fcf0a66/rehlds/engine/sv_user.cpp#L1417-L1424).

Arkshine:
Alright, done.

amxmodx/amxmodx.cpp
```diff
@@ -2626,32 +2626,40 @@ static cell AMX_NATIVE_CALL server_exec(AMX *amx, cell *params)
...
+int sendFakeCommand(AMX *amx, cell *params, bool send_forward = false)
 {
-	int ilen;
-	const char* szCmd = get_amxstring(amx, params[2], 0, ilen);
-	const char* sArg1 = get_amxstring(amx, params[3], 1, ilen);
+	enum args { arg_count, arg_index, arg_command, arg_argument1, arg_argument2 };
 
-	if (ilen == 0)
-		sArg1 = 0;
+	char command[128 * 2];
```
WPMGPRoSToTeMa: Why it is doubled?

WPMGPRoSToTeMa: Why it is doubled?

amxmodx/amxmodx.cpp
```diff
@@ -2626,32 +2626,40 @@ static cell AMX_NATIVE_CALL server_exec(AMX *amx, cell *params)
...
+
+	if (index == 0)
 	{
 		for (int i = 1; i <= gpGlobals->maxClients; ++i)
 		{
 			CPlayer* pPlayer = GET_PLAYER_POINTER_I(i);
 
 			if (pPlayer->ingame /*&& pPlayer->initialized */)
-				UTIL_FakeClientCommand(pPlayer->pEdict, szCmd, sArg1, sArg2, fwd);
+				UTIL_FakeClientCommand(pPlayer->pEdict, command, pArgument1, pArgument2, send_forward);
```
WPMGPRoSToTeMa: Hm, `UTIL_FakeClientCommand` has single `g_fakecmd` variable, so it can be a problem for recursive `engclient_cmd`. We can save a copy of old `g_fakecmd` on the stack.

amxmodx/amxmodx.cpp
Arkshine: Not totally sure why I did that, but yeah if the limit is 128, I will fix it.

amxmodx/amxmodx.cpp
Arkshine: Could you elaborate a little more? I don't really the see the issue here. It's a bit confusing. For sure, if you plan to use the `read_arg*` natives after the first `engclient_cmd` you might get the second `engclient_cmd` data, but that's not useful to do that.

PR #439, opened 2017-05-19
Fix UTIL_ReadFlags function

amezoure:
Converting string to sum of bits is **wrong**.

```sourcepawn
>>> read_flags("!@#$!@!")
-2147483635 // it's wrong, 'cause not in ['a', 'b', ..., 'z']; must be null
```

### DIFF
--- a/amxmodx/amxmodx.h
+++ b/amxmodx/amxmodx.h
@@ -114,7 +114,7 @@ extern AMX_NATIVE_INFO g_GameConfigNatives[];
 #define SETCLIENTMAXSPEED   (*g_engfuncs.pfnSetClientMaxspeed)
 
 char* UTIL_SplitHudMessage(register const char *src);
-int UTIL_ReadFlags(const char* c);
+int UTIL_ReadFlags(const char *c, bool safe = false);
 
 void UTIL_ClientPrint(edict_t *pEntity, int msg_dest, char *msg);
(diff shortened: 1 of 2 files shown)


amezoure:
Gratz! It should be added in amxmodx too :sparkles:

rsKliPPy:
Do we want to allow characters up to '~'? I mean '{', '|', '}' and '~' are valid flags. I noticed that you could use flags other than 'a'-'z' years ago, but I just thought it was normal.

Arkshine:
I guess for safety it should be allowed, It's hard to evaluate whether plugins are using them and documentation doesn't say it must be a-z.

In-line:
@Arkshine We can add one more optional (and by default disabled) parameter to the natives that use `UTIL_ReadFlags` (something like `safe`, `range_checked`, etc..).

amezoure:
I agree with @In-line, and add this feature as soon as possible.

amezoure:
IMHO, we should add information in documentation about input characters (which must be in range `['a', 'b', ..., '{', '|', '}', '~']`), and change calling the `UTIL_ReadFlags` everywhere.

amezoure:
Is this pull request alive? :anguished:

Arkshine:
I thought you wanted to update the documentation?

amezoure:
Closed. I'm not goin' to do anything.

Arkshine:
Alright, is there a reason for that?

PR #440, opened 2017-05-24
Fix temporary ban flags, add lang keys

rsKliPPy:
There was a logic error the flag check expression. Instead of requiring an admin to have either BAN or RCON access, it required them to have both flags.
Also added a new lang key so it prints what actually is disallowing them to ban a player.

Related to: https://bugs.alliedmods.net/show_bug.cgi?id=3218

### DIFF
--- a/plugins/admincmd.sma
+++ b/plugins/admincmd.sma
@@ -232,12 +232,12 @@ public cmdUnban(id, level, cid)
 
 	get_user_authid(id, authid, charsmax(authid))
 
-	if( ~get_user_flags(id) & ( ADMIN_BAN | ADMIN_RCON ) )
+	if( !(get_user_flags(id) & ( ADMIN_BAN | ADMIN_RCON )) )
 	{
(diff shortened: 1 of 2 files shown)


luxxxoor:
@Arkshine it was reported just after a month after their introduction from as i remember, but from as i remember you said  :it's all ok, it must be something else"

Arkshine:
Really? I don't remember it, maybe the ~ with multiple flags got me at this point, but if there is a bug, usually I try to reproduce it, so it's quite weird. Do you have a link for reference? I'm quite curious.

IgnacioFDM:
Even though I run a very recent version of amxmodx, I still have the base plugins from 1.81 or something like that. Probably many others don't bother either to update the plugins and that's why it went unnoticed.

luxxxoor:
i tried to find the link but i don't know how to search all my posts from the forum
anyway i don't think ots relevant anymore

PR #441, opened 2017-05-27
Fix amx_addban and amx_unban, make them safe

rsKliPPy:
Related to: https://github.com/alliedmodders/amxmodx/commit/7589c6c5785d417a8b12b021e42c1b79d03601cf

That update broke `amx_addban` and `amx_unban` by trying to make them safe. However, for whatever reason, Valve made it so it's **required** that one does not put quotation marks around a Steam ID in `banid` and `removeid` commands.
In these commands, Steam IDs are treated as 5 arguments instead of 1 (`STEAM_X:Y:Z` = `STEAM_X`, `:`, `Y`, `:`, `Z`). We take a different approach by checking ourselves if the argument is safe, i.e: it doesn't contain `';'` and `'\n'` characters, because these are command delimiters.

Right now `CL_NOT_FOUND` is printed, but it may not be the most suitable message for this case. Suggestions are welcome.

### DIFF
--- a/plugins/admincmd.sma
+++ b/plugins/admincmd.sma
@@ -221,6 +221,16 @@ public cmdKick(id, level, cid)
 	return PLUGIN_HANDLED
 }
 
+/**
+ * ';' and '\n' are command delimiters. If a command arg contains these 2
+ * it is not safe to be passed to server_cmd() as it may be trying to execute
+ * a command.
+ */
+isCommandArgSafe(const arg[])
+{
+	return contain(arg, ";") == -1 && contain(arg, "\n") == -1;
+}
+
 public cmdUnban(id, level, cid)
 {
 	if (!cmd_access(id, level, cid, 2))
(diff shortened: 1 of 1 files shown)


Arkshine:
@psychonic loves to break things! :P

psychonic:
Welp.

GoRiLliAz:
Here is a screenshot with amx_addban without "" around the steam_id

http://hpics.li/e23bdd1

It would be nice to add a different message that "added to ban list"

It is normal that in banned.cfg there is ecrie
Banid 0.0 STEAM_ID_PENDING
When we do amx_addban without "" around the steam_id.
It should just not write in banned.cfg no?

WPMGPRoSToTeMa:
@GoRiLliAz we can make custom arg parser for amx_addban that won't treat `:` as a delimiter.

GoRiLliAz:
I do not know anything about pawn, so I can not help.
But it would still be good to put another message in case of error because and not to write in the file banned.cfg in case of error too.

rsKliPPy:
Well the Steam ID should be treated as 1 argument in my opinion. Looks like Valve didn't think so.
We could maybe check if an admin passed more than 6 arguments and deal with it the same way Valve did.

Arkshine:
I don't know if it's worth to support without quotes, it has been always with (even the documentation says with), but it could be a small welcome QoL change. Either that or we check and throw a message if steamid is incomplete as it's quite not acceptable to silently pass a broken steamid.

WPMGPRoSToTeMa:
> Looks like Valve didn't think so.

That is because Valve use the same parser for commands and various files (where `:` can be a token).

rsKliPPy:
> That is because Valve use the same parser for commands and various files (where : can be a token).

Yeah I know, but they should've done something different for arguments IMO.

PR #442, opened 2017-05-28
Minor flagman cleanups

IgnacioFDM:


configs/core.ini
```diff
@@ -40,3 +40,8 @@ max_binlog_size 20
 ; 2 - float comparisons
 ; 4 - float rounding
 optimizer 7
+
+; Admin commnand flag manager
```
In-line: *command

Arkshine: I did read like at least twice and I did not see it ahah, time to sleep I guess.

PR #443, opened 2017-06-20
Cstrike: Get offset for m_bMapHasBombTarget

In-line:
I guess that @Arkshine forgot to initialize `BombTargetDesc` in a445e806ea45bd9c09e2121c68d8d5273891ffa1 . 

But I have one issue with this code. For some unknown reason the `CommonConfig->GetOffsetByClass("CHalfLifeMultiplay", "m_bMapHasBombTarget", &BombTargetDesc)` returns false for me. It seems that my code is wrong or [offset](https://github.com/alliedmodders/amxmodx/blob/f31e0e405f41d616c40feba04eadea37f3877897/gamedata/common.games/gamerules.games/cstrike/offsets-chalflifemultiplay.txt) is wrong. So I need additional help with this. Tested under linux (6132 cs.so) without ReGameDLL.

### DIFF
--- a/gamedata/common.games/master.games.txt
+++ b/gamedata/common.games/master.games.txt
@@ -743,6 +743,25 @@
 		"game"  "czero"
 	}
 
+	"gamerules.games/cstrike/offsets-cgamerules.txt"
+	{
+		"game"  "cstrike"
+		"game"  "czero"
+	}
+
+	"gamerules.games/cstrike/offsets-chalflifemultiplay.txt"
+	{
+		"game"  "cstrike"
+		"game"  "czero"
+	}
+
+	"gamerules.games/cstrike/offsets-chalflifetraining.txt"
+	{
+		"game"  "cstrike"
+		"game"  "czero"
+	}
+
+
 
 	//
 	// Day Of Defeat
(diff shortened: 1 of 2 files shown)


Arkshine:
It's because the relevant configs files are not loaded. I guess it's my fault, for some reasons I created a `master.games.txt` file in `/gamerules.games/` directory which references all the files, but such files should be referenced in `/common.games/master.games.txt` instead (like the entities/hostages/others offsets).

To fix this, you need to either use `LoadGameConfigFile` with `common.games/gamerules.games` or to move the content of `/gamerules.games/master.games.txt` to `/common.games/master.games.txt`. The latter would make more sense, no reason to split.

In-line:
I haven't tested it yet, but I think it should be fine. Later it will be tested, if you need it.

Arkshine:
I think it's fine to delete the old file now you moved the references (you need to update fakemeta code by removing the second LoadGameConfigFile, and merging error description I guess)

In-line:
@Arkshine, what about third party modules relying on this?

In-line:
I think old file should remain for comparability reasons.

Arkshine:
Since it has been around for a long time, I guess it's okay to not delete it for compatibility. It's silly but not that much a big deal.

PR #446, opened 2017-07-03
Fix client_disconnected isn't called on mapchange when ReHLDS API ena…

WPMGPRoSToTeMa:
…bled

### DIFF
--- a/amxmodx/meta_api.cpp
+++ b/amxmodx/meta_api.cpp
@@ -120,6 +120,9 @@ int mState;
 int g_srvindex;
 
 CDetour *DropClientDetour;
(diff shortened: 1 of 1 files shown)


Arkshine:
I guess it looks good. Too bad there are no enable/disable functions similar to cdetour so it would not be needed to rely on such global variables.

Thanks!

WPMGPRoSToTeMa:
@Arkshine we need a class that will wrap detour and rehlds hook.

kristi-alb:
wrong sorry

Arkshine:
Should be fixed now, thanks to psychonic.

PR #449, opened 2017-07-26
Fix GetItemInfo incorrect associated handler

Arkshine:
Reported by PartialCloning here: https://forums.alliedmods.net/showthread.php?t=299753.

`GetItemInfo` returns an integer.


rsKliPPy:
Looks good. :ship:

PR #450, opened 2017-07-27
Fix broken cs_set_user_model update_index parameter for the default engine

Arkshine:
Related to #417. 
Reported by PartialCloning here: https://forums.alliedmods.net/showthread.php?t=299821.

Moved the code in the native for convenience.


rsKliPPy:
:sheep:

PR #452, opened 2017-08-07
Update Hamsandwich for Sven Coop 5.17

Arkshine:
WIP  :construction: 

Waiting for data from Sven Coop team to confirm data types of some functions.


Arkshine:
I guess it should be enough for Svencoop 5.17.

Can't sanely test everything. Just tested some to see if well aligned. 

Also, probably a few functions are missing, but well, can be added at another time.

PR #453, opened 2017-08-07
Adjust 'amxx cvars' command output

Arkshine:
Reading 'BOUND' and 'BOUNDED' is kind of confusing.
Showing the binds in the global list is not useful because it has no effects on others plugins.
Also 'BOUNDED" is now 'MIN MAX' for convenience and consistency.


rsKliPPy:
I'm happy.

PR #415, opened 2017-02-26
Add amx_searchcmd command in adminhelp plugin and do some clean up

Arkshine:
Essentially:
- Added `amx_searchcmd`: Similar to `amx_help` except you can provide a string to match partially the command names. Quite useful if you have a lot of commands.
- Moved the hardcoded `15` seconds for the help message to a new`amx_help_display_msg_time` cvar.
- The commands description are translated.
- Some general cleanup, using new stuff as well.

Feel free to correct the english if necessary, and to provide the translations in your language.

PR #454, opened 2017-08-21
Minor fixes in configs files.

Freeman-AM:
Omission related to https://github.com/alliedmodders/amxmodx/pull/415 and https://github.com/alliedmodders/amxmodx/pull/442.


Freeman-AM:
Added more fix in config files.

PR #457, opened 2017-09-03
Use smart pointers in NativeHandle class

Amaroq7:
* Use smart pointers instead of raw ones
* Use ``size_t`` type when returning and passing handles' ids

### DIFF
--- a/amxmodx/natives_handles.h
+++ b/amxmodx/natives_handles.h
@@ -11,6 +11,7 @@
 #define _NATIVES_NATIVES_HANDLES_H_
 
 #include <amtl/am-vector.h>
+#include <amtl/am-autoptr.h>
 
 // Note: All handles start at 1. 0 and below are invalid handles.
 //       This way, a plugin that doesn't initialize a vector or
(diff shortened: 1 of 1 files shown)


amxmodx/natives_handles.h
```diff
@@ -50,58 +43,58 @@ class NativeHandle
...
+			m_handles.append(ke::AutoPtr<T>(data));
 
 			return m_handles.length();
 		}
 
-		bool destroy(int handle)
+		bool destroy(size_t handle)
 		{
-			handle--;
+			--handle;
```
IgnacioFDM: :nauseated_face:

Amaroq7: I've done it to keep up consistency with ``lookup`` method where pre decrement operator is used. 😝

amxmodx/natives_handles.h
```diff
@@ -111,7 +104,6 @@ class NativeHandle
 				return false;
 			}
 
-			delete m_handles[handle];
```
Arkshine: Unless I'm wrong, this is still needed here, since the slots are reused, you are going to leak memory.

PR #460, opened 2017-09-23
Adding TSE module to AMXX distribution

SNMetamorph:


Arkshine:
Thanks for submitting your module.

Here few things after a first fast look:

VS
* This module project should be included in the main amxmodx solution as well (`Modules` filter)
* Toolset should include XP support (`v120_xp` at least)
* The _Header Files_ and _Source Files_ filters name are in Russian 
* The preprocessor defines should contain at least: `WIN32;_WINDOWS;_USRDLL;TSE_EXPORTS;HAVE_STDINT_H` and `DEBUG` if in debug mode, or `NDEBUG` if in release mode.
* You should add the `Pawn Includes` filter and referencing `tse.inc`.

AMBuild
* `AMBuilder` file is missing. This is the main way that AMXX uses to compile. You will have to update `AMBuildScript` to add the module there (end of file) and `support/PackageScript` to add the mod as well.
* `version.rc` is missing.

Module Config
* For consistency, `MODULE_VERSION` and `MODULE_AUTHOR` should have respectively the values of `AMXX_VERSION` and `"AMX Mod X Dev Team"`. You can add your name in the `version` command.
* `MODULE_LIBCLASS` should be empty here.

Game Data
* For any game data such as offsets/signatures/symbols, that would be welcomed to use a game data file. You can create one in `/gamedata/modules.games/game.ts.txt`. You can look at cstrike module to understand how to load such file.
* A `test` command is left. You can delete it along the `memory*` and `dllfuncs` files.

Code
* You should replace these `UINT`, `USHORT` and `ULONG` types with the standard ones `uint32_t`, and `uint16_t`.
* I'm not sure to like you put all the natives code in a header.
* You're using `pfnRegUserMsg` in some natives when you could register their ids.
* I would enable `PlayerPreThink` only if there are a least one of the forwards used in a plugin. You can check the engine module.

Documentation
* That would be welcomed if you could use the same documentation format as the other files, using /** */ and explaining the parameters/return as well. It's useful itself, but this is also required for the online documentation (http://www.amxmodx.org/api/).
* The license header is not correct. See the others files.
* No need `AMXX_VERSION_NUM` check.

There are probably other things to say, but it's a good start I guess.
Usually, you can just look at others modules/files to see how things are done.

Feel free to ask if you don't understand something.

WPMGPRoSToTeMa:
Why you don't use `MF_LogError` for errors, instead of returning some value?

SNMetamorph:
Hmm, for some reason did not pay any attention to this function. In the near future will deal with the logic of the this function, and will add to the module.

SNMetamorph:
What is the version.rc file? What is it's role? It needs to be generated or can I take this file from another module and change some values?

Arkshine:
Unless you plan something, memory* and dllfuncs* files should be removed.

SNMetamorph:
Okay, now do it.

modules/ts/tse/functions.cpp
```diff
@@ -0,0 +1,519 @@
...
+	cell *attachs = MF_GetAmxAddr(amx, params[5]);
+	cell *active_attachs = MF_GetAmxAddr(amx, params[6]);
+	*clip = Player(pid)->GetWeapPDataInt(WeaponsList[weapon].offsets.clip);
+	*ammo = Player(pid)->GetWeapPDataInt(WeaponsList[weapon].offsets.ammo);
+	*attachs = Player(pid)->GetWeapPDataInt(WeaponsList[weapon].offsets.clip + 3);
+	*active_attachs = Player(pid)->GetWeapPDataInt(WeaponsList[weapon].offsets.clip + 4);
+	return 1;
+}
+
+static cell AMX_NATIVE_CALL tse_isuserhasweap(AMX *amx, cell *params)
```
In-line: This function can be refactored to something like this.

```
static cell AMX_NATIVE_CALL tse_isuserhasweap(AMX *amx, cell *params)
{
    byte pid = params[1];
    
    if (!IsPlayerValid(amx, pid)) return -1;
    if (!Player(pid)->IsAlive()) return -1;
    
    byte weapon = params[2];
    
    if (!WeaponsList[weapon].offsets.clip) return -1;
    
    byte offset = !!WeaponsList[weapon].offsets.fmbase;
    cell result = Player(pid)->GetWeapPDataInt(WeaponsList[weapon].offsets.clip - offset) > 0;
    
    return result;
}
```

modules/ts/tse/functions.cpp
```diff
@@ -0,0 +1,519 @@
...
+	if (!Player(pid)->IsAlive()) return -1;
+	switch ((uint32_t)Player(pid)->GetPDataInt(462))
+	{
+		case 65536:
+			return 1;
+			break;
+		case 16777216:
+			return 2;
+			break;
+		case 16842752:
```
In-line: I think it's better to replace raw numbers with meaningful defines and `break` after `return` is redundant.

AMBuildScript
```diff
@@ -422,6 +422,7 @@ builder.RunBuildScripts(
     'modules/tfcx/AMBuilder',
     'modules/ts/tsfun/AMBuilder',
     'modules/ts/tsx/AMBuilder',
+	'modules/ts/tse/AMBuilder',
```
In-line: Wrong indentation.

modules/ts/tse/CPlayer.cpp
```diff
@@ -0,0 +1,190 @@
...
+	MDLL_KeyValue(ent, &params);
+
+	MDLL_Spawn(ent);
+	ent->v.origin = coord;
+	return ent;
+}
+
+edict_t *CreatePowerup(uint16_t type, Vector coord, short ttl) {
+	edict_t *ent = CREATE_NAMED_ENTITY((string_t)MAKE_STRING("ts_powerup"));
+	if (FNullEnt(ent)) return NULL;
```
In-line: I think it's better to use `nullptr`.

modules/ts/tse/CPlayer.h
```diff
@@ -0,0 +1,91 @@
...
+	inline float GetEffectsPhysicsSpeed() { return this->GetPDataFloat(87); }
+	inline void SetEffectsPhysicsSpeed(float val) { this->SetPDataFloat(87, val); }
+};
+
+// other methods
+extern CPlayer PlayersArray[33];
+edict_t *CreateWeapon(int id, Vector coord, short ttl, uint16_t clips, byte atcments);
+edict_t *CreatePowerup(uint16_t type, Vector coord, short ttl);
+
+inline byte IsPlayerValid(AMX *amx, int pl) {
```
In-line: Maybe `bool` as return type?

modules/ts/tse/functions.cpp
```diff
@@ -0,0 +1,519 @@
...
+	{ "tse_configmeleeatk", tse_configmeleeatk },
+	{ NULL, NULL }
+};
+
+static cell AMX_NATIVE_CALL tse_sendweapparams(AMX *amx, cell *params)
+{
+	byte pid = params[1];
+	if (!IsPlayerValid(amx, pid)) return 0;
+	char paramsstr[128];
+	sprintf(paramsstr, "%i %f %f %f %f %i", params[2], amx_ctof(params[3]), amx_ctof(params[4]), amx_ctof(params[5]), amx_ctof(params[6]), params[7]);
```
In-line: I think it's better to use `snprintf` here, because buffer can overflow.

plugins/include/tse.inc
```diff
@@ -0,0 +1,266 @@
...
+native tse_getuseractivepwup(idx);
+
+// Sets user powerup duration (in seconds), on error returns 0
+native tse_setuserpwupduration(idx, duration);
+
+// Sets user fake powerup effect (as though user had a speficied powerup and user activated it), on error returns 0
+native tse_setfakepwup(idx, pwupid, Float:duration);
+
+/* Configures fake powerup options (player and effects physical speed, aura distance)
+*  WARNING: THIS FUNCTION WORKS ONLY WITH FUNCTION tse_setfakepwup AND POWERUP TYPE TSE_PWUP_SLOWMO OR TS_PWUP_SLOWPAUSE ONLY!
```
In-line: You should emit native error if function is misused.

plugins/include/tse.inc
```diff
@@ -0,0 +1,266 @@
...
+
+// Returns player current items (TSE_ITEM_*), on error returns -1
+// WARNING: If you want check presence of who items, use them sum
+native tse_getusercurritems(idx);
+
+// Sets player current items (TSE_ITEM_*), on error returns 0
+native tse_setusercurritems(idx, val);
+
+// Returns player current weapon ID (TSE_WPN_*), on error returns -1
+native tse_getusercurrweap(idx);
```
In-line: Create tag for enum like
```
enum MY_ENUM
{
  ITEM1,
  ITEM2,
}
```
Then use it in functions.
```
native MY_ENUM:tse_makemegod(idx);
```

modules/ts/tse/functions.cpp
Arkshine: Note: there is `<amtl/am-string.h>` for that, see `ke::SafeSprintf`.

modules/ts/tse/functions.cpp
WPMGPRoSToTeMa: @In-line I prefer explicit:
`byte offset = !!WeaponsList[weapon].offsets.fmbase;`
->
`byte offset = WeaponsList[weapon].offsets.fmbase != 0 ? 1 : 0;`

plugins/include/tse.inc
SNMetamorph: Thanks for revision of my code, I will take action as soon as possible.

SNMetamorph: Thanks for revision of my code, I will take action as soon as possible.

SNMetamorph:
Thanks for revision of my code, I will take action as soon as possible.

modules/ts/tse/functions.cpp
```diff
@@ -0,0 +1,519 @@
...
+	if (!Player(pid)->IsAlive()) return 0;
+	Player(pid)->SetCash(params[2]);
+	return 1;
+}
+
+static cell AMX_NATIVE_CALL tse_getuserkevlar(AMX *amx, cell *params)
+{
+	byte pid = params[1];
+	if (!IsPlayerValid(amx, pid)) return -1;
+	if (!Player(pid)->IsAlive()) return -1;
```
In-line: You can create macros for this 2 checks to reduce copypaste

modules/ts/tse/functions.cpp
SNMetamorph: I applied changes in this aspect, thanks.

plugins/include/tse.inc
SNMetamorph: This method is likely to be true, but it is unusual and is rarely used in Pawn (for example see engine_const.inc, fakemeta_const.inc). Anyway, thanks for your helping!

Arkshine:
Are you still alive?

SNMetamorph:
Yes. I'm not show any activity because busy with education. 21st December, I will begin winter vacation and I will resume work on the module.

SNMetamorph:
How I can contain bitsums in gamedata files?

PR #461, opened 2017-10-02
Little cleanup

Amaroq7:
### Summary

- Remove useless natives such as:
```hcsardhnexsnu```
It seems to just test some experimental stuff and doesn't provide anything useful for scripter.
```is_user_hacking```, ```is_rukia_a_hag```
I believe these two were added as some kind of joke, but they're neither documented nor serve any useful purpose.
```amxx_setpl_curweap```
Not documented native. I can't find any situation where it can be used at the moment, maybe there are some, otherwise I'd opt for removing it.

- Get rid of AMX module support completely (which is broken anyway and has been since https://github.com/alliedmodders/amxmodx/commit/4609cb409b9d3679de9d120e17796156e546dd52)
If module doesn't have ```AMXX_Query``` function, it automatically gets marked as failed with ```MODULE_NOQUERY``` status.

- Remove FAKEMETA leftovers
Remove FAKEMETA defines which are unused now.

- Clean up ```GiveFnptrsToDll``` function
The ```GiveFnptrsToDll``` function was adjusted to work without ```.def``` under Windows. Its current look gives me a headache 😛. I'd like to make it look more readable and there are [3 possibilities](https://github.com/alliedmodders/amxmodx/pull/461#issuecomment-334595553).

- Move ```require_module``` native
I really don't see any reason why it should be exposed globally.

- Remove ```alloc_amxmemory```, ```free_amxmemory``` and ```strip_name``` functions
They're not used anywhere.


IgnacioFDM:
```is_user_hacking```, ```is_rukia_a_hag``` should be kept IMO since there can be plugins using those even if as a joke, and removing them would break them, especially considering it doesn't really hurt having them there.

I've used ```is_rukia_a_hag``` a few times for benchmarking purposes, which reminds me I should PR some performance improvements I did a few months ago but forgot.

I'm not really familiar with the rest to comment.

dvander:
drive by: just remove 'em. they are mean.

Arkshine:
As said Amaroq7 they serve no purpose. It was fun at some point, but that's it. Time to :fire: them.

I'm a little concerned about `amxx_setpl_curweap`. Unlike the others, it's a usable but undocumented native, which has been referenced in the 1.76 changelog. Though, Google doesn't give many references apart where dvander got this outrageous idea (https://forums.alliedmods.net/showthread.php?p=345424) , the changelog and 2 russian boards talking about the possibility to hack with it around `get_user_weapon`.

WPMGPRoSToTeMa:
> The GiveFnptrsToDll function was adjusted to work with fakemeta years ago, but that's no longer the case, so I think it's safe to make it back to its original state. Its current look gives me a headache 😛.

Fakemeta? Wow. I thought it was done to avoid using .def file in msvc.

Amaroq7:
Right, that change was made along with adding support for fakemeta, so I've
wrongly concluded it was related to fakemeta. I'll revert it later. Thanks
for pointing that out.

On 3 Oct 2017 11:27 p.m., "Artem Golubikhin" <notifications@github.com>
wrote:

The GiveFnptrsToDll function was adjusted to work with fakemeta years ago,
but that's no longer the case, so I think it's safe to make it back to its
original state. Its current look gives me a headache 😛.

Fakemeta? Wow. I thought it was done to avoid using .def file in msvc.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/alliedmodders/amxmodx/pull/461#issuecomment-333983772>,
or mute the thread

WPMGPRoSToTeMa:
@Amaroq7 but `.def` file is here, so idk why `GiveFnptrsToDll` is still so complicated.

Amaroq7:
I've done some testing and it seems that file is redundant and can be safely removed, probably should have been when that adjustment was introduced. Anyway after my change AMXX will fail and fails to load, because  ```GiveFnptrsToDll``` gets decorated and exported as ```_GiveFnptrsToDll@8```. It can be fixed by:

1. leaving the function as is.
2. making use of ```.def``` file again.
3. passing these options ```/EXPORT:GiveFnptrsToDll=_GiveFnptrsToDll@8,@1 /SECTION:.data,RW``` to linker.

I like the 3rd option the most, because there's no need to maintain extra file and we get cleaned up function.

WPMGPRoSToTeMa:
@Amaroq7 `.def` currently is unused?

WPMGPRoSToTeMa:
Can you also remove this?
https://github.com/alliedmodders/amxmodx/blob/779a4d9e59282428e04a2b5d6f37f2a7b9a5be36/modules/engine/engine.cpp#L560
And this:
https://github.com/alliedmodders/amxmodx/blob/779a4d9e59282428e04a2b5d6f37f2a7b9a5be36/modules/engine/amxxapi.cpp#L235

WPMGPRoSToTeMa:
@Amaroq7 oh, sorry, revert https://github.com/alliedmodders/amxmodx/pull/461#issuecomment-335028709. `memcpy` doesn't copy '\0'. You can also try to change it to normal string copy.

Arkshine:
Are you done @Amaroq7?

Amaroq7:
I think so, I've applied the same way to export ``GiveFnptrsToDll`` in modules (where it's needed) as is in the core. I wouldn't mind if someone could check that just to be sure everything gets loaded successfully.

Arkshine:
If the changes could be reflected in the VS projects, that would be welcomed as well.

Amaroq7:
They've been updated.

Arkshine:
Almost there. I don't see the Json module. You will likely need to rebase from master and add the changes to this module as well.

Amaroq7:
As I see Json module doesn't use ``USE_METAMOD`` define, so it doesn't export ``GiveFnptrsToDll`` at all, if we talk about adding ``/EXPORT`` to ``.vcxproj`` file.

Arkshine:
Oh right, then it's all good. Thanks!

PR #462, opened 2017-10-03
Fix a crash with overlapping newmenu

Arkshine:
Reported by @aron9forever on the forum: https://forums.alliedmods.net/showthread.php?t=301737.

Typically the situation where a global menu is destroyed and re-displayed.
Before displaying a menu, `menu_display` will try to close/destroy any previous menu. 
If it happens two same menu overlap, `menu_display` will destroy and display the same menu, resulting in a crash because the menu being deleted right before.

The proposed solution is to check the menu validity and erroring if the menu has been previously destroyed.

### DIFF
--- a/amxmodx/newmenus.cpp
+++ b/amxmodx/newmenus.cpp
@@ -809,15 +809,22 @@ static cell AMX_NATIVE_CALL menu_items(AMX *amx, cell *params)
 //page indices start at 0!
 static cell AMX_NATIVE_CALL menu_display(AMX *amx, cell *params)
 {
-	GETMENU(params[2]);
-	
+	auto handle = params[2];
+	GETMENU(handle);
+
 	int player = params[1];
 	int page = params[3];
 	CPlayer* pPlayer = GET_PLAYER_POINTER_I(player);
-	
+
 	if (!CloseNewMenus(pPlayer))
 	{
 		LogError(amx, AMX_ERR_NATIVE, "Plugin called menu_display when item=MENU_EXIT");
+		return 0;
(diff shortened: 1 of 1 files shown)


amxmodx/newmenus.cpp
```diff
@@ -809,18 +809,25 @@ static cell AMX_NATIVE_CALL menu_items(AMX *amx, cell *params)
...
+
 	if (!CloseNewMenus(pPlayer))
 	{
 		LogError(amx, AMX_ERR_NATIVE, "Plugin called menu_display when item=MENU_EXIT");
 		return 0;
 	}
 
+	if (!g_NewMenus[handle])
+	{
+		LogError(amx, AMX_ERR_NATIVE, "Invalid menu id %d (was previously destroyed).", params[2]);
```
Amaroq7: I think it's better to use here ```handle``` as well.

Nextra:
If I see this correctly, the menu handle is unnecessarily pre-fetched before the callback (and thus the potentially lethal side effect) is triggered by `CloseNewMenus`. `get_menu_by_id` and `GETMENU` already perform the `g_NewMenus` check, this patch essentially does it twice. If the function calls `CloseNewMenus` as soon as possible it should also fix the issue.

Arkshine:
You're right, `GETMENU` can simply be moved after `CloseNewMenus`. Not sure why I did not see that. 
Though, you could argue that checking the menu handle should be done first.

Oh well, not like it does matter much anyway. if a cleanup happens in this file, I will change it.

PR #463, opened 2017-10-05
Fix #395 backward compatibility issue by adding SetParamEntity2 native

Arkshine:
Unfortunately a bit late, but this fixes a backward compatibility issue introduced in #395.
Reported by Th3822 on the forum: https://forums.alliedmods.net/showthread.php?p=2512733#post2512733.

Proposed solution is to keep the old behavior and adding a new `SetParamEntity2` native which will reflect any change in the corresponding post forward.


rsKliPPy:
:sheep:

PR #466, opened 2017-10-28
Bump MySQL version to 5.5

Arkshine:
Bumps MySQL version to `5.5`.

OS | Old version | New version | Comment
--- | ------------- |-------------- | ---------
windows | `5.0.?` | `5.5.54`  | Latest available version before Microsoft Visual C++ 2008 Redistributable Package is required.
linux       | `5.1.?` | `5.5.58` | Latest available version.
mac        | `5.1.?` | `5.5.40` | Latest available version.

Requested by several users.

The main reason to not use `5.6+` is to keep Windows XP compatibility, avoid unnecessary prerequisite and be consistent across the platforms as much as possible.

Note: buildbot doesn't use `checkout-deps.sh`. Currently, the MySQL CDN doesn't have `.58` version so used `.57` for now.

### DIFF
--- a/modules/mysqlx/msvc12/mysqlx.vcxproj
+++ b/modules/mysqlx/msvc12/mysqlx.vcxproj
@@ -54,7 +54,7 @@
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <ClCompile>
       <Optimization>Disabled</Optimization>
-      <AdditionalIncludeDirectories>..\;..\..\..\public;..\..\..\public\sdk;..\..\..\public\amtl;..\..\third_party;..\..\third_party\hashing;..\..\..\..\mysql-5.0\include;..\mysql;..\sdk;..\thread;$(METAMOD)\metamod;$(HLSDK)\common;$(HLSDK)\engine;$(HLSDK)\dlls;$(HLSDK)\public;$(MYSQL5)\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>..\;..\..\..\public;..\..\..\public\sdk;..\..\..\public\amtl;..\..\third_party;..\..\third_party\hashing;..\..\..\..\mysql-5.5\include;..\mysql;..\sdk;..\thread;$(METAMOD)\metamod;$(HLSDK)\common;$(HLSDK)\engine;$(HLSDK)\dlls;$(HLSDK)\public;$(MYSQL55)\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PreprocessorDefinitions>WIN32;_CRT_SECURE_NO_DEPRECATE;_CRT_SECURE_NO_WARNINGS;_CRT_NONSTDC_NO_DEPRECATE;_ITERATOR_DEBUG_LEVEL=0;_DEBUG;_WINDOWS;_USRDLL;MYSQL2_EXPORTS;SM_DEFAULT_THREADER;HAVE_STDINT_H;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <MinimalRebuild>true</MinimalRebuild>
       <BasicRuntimeChecks>EnableFastChecks</BasicRuntimeChecks>
@@ -66,7 +66,7 @@
       <DebugInformationFormat>ProgramDatabase</DebugInformationFormat>
     </ClCompile>
     <Link>
-      <AdditionalDependencies>ws2_32.lib;..\..\..\..\mysql-5.0\lib\opt\mysqlclient.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies>ws2_32.lib;..\..\..\..\mysql-5.5\lib\mysqlclient.lib;%(AdditionalDependencies)</AdditionalDependencies>
       <IgnoreSpecificDefaultLibraries>LIBCMT;%(IgnoreSpecificDefaultLibraries)</IgnoreSpecificDefaultLibraries>
       <GenerateDebugInformation>true</GenerateDebugInformation>
       <ProgramDatabaseFile>$(OutDir)mysql2.pdb</ProgramDatabaseFile>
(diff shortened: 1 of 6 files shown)


modules/mysqlx/msvc12/mysqlx.vcxproj
```diff
@@ -78,7 +78,7 @@
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
     <ClCompile>
-      <AdditionalIncludeDirectories>..\;..\..\..\public;..\..\..\public\sdk;..\..\..\public\amtl;..\..\third_party;..\..\third_party\hashing;..\..\..\..\mysql-5.0\include;..\mysql;..\sdk;..\thread;$(METAMOD)\metamod;$(HLSDK)\common;$(HLSDK)\engine;$(HLSDK)\dlls;$(HLSDK)\public;$(MYSQL5)\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>..\;..\..\..\public;..\..\..\public\sdk;..\..\..\public\amtl;..\..\third_party;..\..\third_party\hashing;..\..\..\..\mysql-5.5\include;..\mysql;..\sdk;..\thread;$(METAMOD)\metamod;$(HLSDK)\common;$(HLSDK)\engine;$(HLSDK)\dlls;$(HLSDK)\public;$(MYSQL5)\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
IgnacioFDM: Shouldn't it be MYSQL55?

Arkshine:
I've adjusted the windows version and updated the first post for better readability.
I've also tested fastly with a plugin under windows and linux, and was working as expected.

How to cache with Travis?

Freeman-AM:
Mmmm this PR broke mac OSX automated builds.

psychonic:
@Freeman-AM indeed. It's on my list to fix.

psychonic:
The macOS builds are fixed now.

Destro-:
I in windows with version 5.5.? had problem, the timeout opcion not working. I think that to this day this problem still exists in all branch 5.5.

https://bugs.mysql.com/bug.php?id=36225
https://bugs.mysql.com/bug.php?id=41911 duplicate

Fix only in 5.6 ?
https://www.google.com/search?q=%2236225%22+site%3Ahttps%3A%2F%2Fdev.mysql.com%2Fdoc%2Frelnotes%2Fmysql%2F

IgnacioFDM:
This is probably going to be an unpopular opinion, but I think it's best to simply drop XP support and upgrade to the latest version [1]. XP has been EOL'd almost 4 years ago, most mainstream software (such as Google Chrome) dropped XP support years ago, and for anybody still running XP that for some reason can't upgrade to either Win7+ or some lightweight linux distro, having to run either 1.8.3 or the (as of today) latest dev build instead of future builds is probably the least of his worries. 

Even if we don't drop XP support because of this, we should probably ask ourselves, how much longer should we support XP? 1 year? 3 years? 5 years?

Sorry for the rant

[1] I couldn't get the absolute latest MySQL version to build because of some small issue I currently don't remember, but I'm running a close to latest version with no issues.

WPMGPRoSToTeMa:
> but I think it's best to simply drop XP support

The problem isn't Windows XP, but Windows Server 2003 I think.

@Arkshine do you know how they disabled Windows XP support? Maybe it can enabled back using some compile option?

Arkshine:
This might be this? https://github.com/mysql/mysql-server/commit/1261600865b7553242f63d088d57ec843adaa01f and https://github.com/mysql/mysql-server/commit/aa5fa2d06f247340e08a0f6b20c150ee5aad62b3

One of the reasons was also 5.5.55+ doesn't provide static libraries anymore by default and recompiling is annoying. I would not mind dropping XP/2003 support since the next version is major and it makes requirement could bump, but as said WPMGPRoSToTeMa, 2003 is likely still being widely used.

IgnacioFDM:
I would assume (pure conjecture, of course) that WS 2003 is less of an issue. WS 2003 has been EOL since july 2015, and those running windows server are almost surely more knowledgeable, security conscious and financially able to upgrade than some random kid who was gifted an old computer with XP.

In case of a requirement bump, I would think that Windows 7/Windows server 2008 (R2 ???) or above, and RHEL 6 or above would be reasonable.

In-line:
Just don't remove backward compatibility support without neccesity.
(Re)HLDS supports Windows XP so we should do the same.

On the other hand we can add a message about deprecation and remove legacy
support after some time.

Server administrators just want everything to work out of box. It will be
bad if after the update server starts to crash, without a special warning
or  an anouncement.

IgnacioFDM:
>Server administrators just want everything to work out of box. It will be
bad if after the update server starts to crash, without a special warning
or  an anouncement.

It's not like AMXModX has auto update or something like that. If a server admin updates to a newer version and it stops working he can just go back to the old version with basically no effort in a couple of minutes. And it's a dev version in fact, so you should be even less surprised.

And it's not like you are obliged to run the latest amxx version. Amxx is quite mature and stable (both as in features and as in bugs). The intersection of people who run WS 2003 and those who absolutely must use some new feature/fix not yet included in the as of today dev build is almost without a doubt null.

I don't see it as such a big deal. For that random guy that still uses WS 2003 and also frequently updates to the latest dev build, when he realizes it's not supported any longer he can just go back to what he was using and start planning to upgrade whenever he wishes. Whoever is still runing WS 2003 knows the upgrade is long due.

voed:
@In-line 
`Just don't remove backward compatibility support without neccesity.
(Re)HLDS supports Windows XP so we should do the same.`
Why? If someone cant update OS for any reasons and they NEED sql, they can use older AMXX builds. Its not about dropping support by all AMXX, its just about SQL library.

Arkshine:
There is also the possibility to go with the painful way by patching ourselves this specific bug, keeping 5.5 version and XP/2003 compatibility at least until the next AMXX 1.9 version is released. The main idea is to have an official version as a reference (and not an obscure dev build), warning the users of a potential requirement bump in a future release and giving them plenty of time.

PR #467, opened 2017-11-26
Fix open_dir not working properly recursively with valve file system

Arkshine:
Lazy fix. Saving the address of a local static variable can't properly work recursively.

### DIFF
--- a/amxmodx/file.cpp
+++ b/amxmodx/file.cpp
@@ -744,10 +744,16 @@ enum FileType
 
 struct DirectoryHandle
 {
-	DirectoryHandle(void* handle_, bool valvefs_) : handle(handle_), valvefs(valvefs_) {}
+	DirectoryHandle(CDirectory *handle_) : handle(handle_), valvefs(false) {}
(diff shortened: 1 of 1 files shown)


amxmodx/file.cpp
```diff
@@ -744,10 +744,16 @@ enum FileType
...
+	DirectoryHandle(CDirectory *handle_) : handle(handle_), valvefs(false) {}
+	DirectoryHandle(FileFindHandle_t handle_) : handle_vfs(handle_), valvefs(true) {}
 
-	void* handle;
-	bool  valvefs;
+	union
+	{
+		CDirectory*      handle;
+		FileFindHandle_t handle_vfs;
+	};
```
WPMGPRoSToTeMa: There is no variant in amtl?

Arkshine: Looks like not. Well, C++17 features are not yet a thing in AMTL. Feel free to contribute!

rsKliPPy:
As per @WPMGPRoSToTeMa 's suggestion, Variant should be used instead once (if) it makes it to AMTL.

🐑

PR #469, opened 2017-11-30
Add natives to access the command line

rsKliPPy:
`pfnCheckParm` has been there for years and it was never implemented in AMXX.
Instead of just adding it to Fakemeta's `EngFunc_*` and `FM_*`, I created new, more intuitive, natives.

### DIFF
--- a/amxmodx/commandline.cpp
+++ b/amxmodx/commandline.cpp
@@ -1,0 +1,100 @@
+#include "amxmodx.h"
+
+#define CHECK_COMMANDLINE_AVAILABILITY()                                        \
+    if (!IsCommandLineAvailable())                                              \
+    {                                                                           \
+        LogError(amx, AMX_ERR_NATIVE, "The command line is not available.");    \
+        return 0;                                                               \
+    }
+
+static bool IsCommandLineAvailable()
+{
+    return g_engfuncs.pfnCheckParm != nullptr;
+}
+
+// native bool:IsCommandLineAvailable();
+static cell AMX_NATIVE_CALL IsCommandLineAvailable(AMX *amx, cell *params)
+{
+    return static_cast<cell>(IsCommandLineAvailable());
+}
+
+// native bool:FindCommandLineParam(const param[]);
+static cell AMX_NATIVE_CALL FindCommandLineParam(AMX *amx, cell *params)
+{
+    enum {arg_count, arg_param};
+
+    CHECK_COMMANDLINE_AVAILABILITY();
+
+    int length;
+    const char *param = get_amxstring(amx, params[arg_param], 0, length);
+    if (length == 0)
+    {
+        return 0;
+    }
+    return g_engfuncs.pfnCheckParm(param, nullptr);
+}
+
+// native GetCommandLineParam(const param[], value[], maxlen, const defaultValue[]);
+static cell AMX_NATIVE_CALL GetCommandLineParam(AMX *amx, cell *params)
+{
+    enum {arg_count, arg_param, arg_value, arg_maxlen, arg_defaultValue};
+
+    CHECK_COMMANDLINE_AVAILABILITY();
+
+    int length;
+    char *value;
+    const char *param = get_amxstring(amx, params[arg_param], 0, length);
+    if (length == 0 || !g_engfuncs.pfnCheckParm(param, &value))
+    {
+        return set_amxstring(amx, params[arg_value], get_amxstring(amx, params[arg_defaultValue], 0, length), params[arg_maxlen]);
+    }
+
+    return set_amxstring(amx, params[arg_value], value, params[arg_maxlen]);
+}
+
+// native GetCommandLineParamInt(const param[], defaultValue);
+static cell AMX_NATIVE_CALL GetCommandLineParamInt(AMX *amx, cell *params)
+{
+    enum {arg_count, arg_param, arg_defaultValue};
+    
+    CHECK_COMMANDLINE_AVAILABILITY();
+
+    int length;
+    char *value;
+    const char *param = get_amxstring(amx, params[arg_param], 0, length);
+    if (length == 0 || !g_engfuncs.pfnCheckParm(param, &value))
+    {
+        return params[arg_defaultValue];
+    }
+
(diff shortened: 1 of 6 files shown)


WPMGPRoSToTeMa:
Maybe `Get*` functions should be suffixed with `OrDefault`?

amxmodx/commandline.cpp
```diff
@@ -0,0 +1,100 @@
...
+// native bool:FindCommandLineParam(const param[]);
+static cell AMX_NATIVE_CALL FindCommandLineParam(AMX *amx, cell *params)
+{
+    enum {arg_count, arg_param};
+
+    CHECK_COMMANDLINE_AVAILABILITY();
+
+    int length;
+    const char *param = get_amxstring(amx, params[arg_param], 0, length);
+    if (length == 0)
```
WPMGPRoSToTeMa: What is the purpose of `length` check?

Arkshine: In case the param is empty (for some reasons) to avoid unnecessary code execution in the engine, even though it doesn't matter, but since you got the length already available, there is no reason to no use it.

WPMGPRoSToTeMa: @Arkshine it is a bit confusing, it makes me think like `EngCheckParm` can produce errors if parameter is empty string. So it should be explicitly commented with something like this: `// For optimization purposes`.

Arkshine:
IMO `OrDefault` seems unnecessarily verbose. Native names are already quite long also.

WPMGPRoSToTeMa:
> IMO `OrDefault` seems unnecessarily verbose. Native names are already quite long also.

In my opinion `Get...` functions should return value or throw an error. Implicitly return some default value (which is not explicitly defined on the call side) is some unclear. IMHO there are 3 kinds of get functions:
`Get...` returns a value or throws an error if param can't be found.
`Get...OrDefault` returns a value or returns a default value if param can't be found.
`TryGet...` returns `true` and returns value byref or returns `false` if param can't be found.

rsKliPPy:
> IMO OrDefault seems unnecessarily verbose. Native names are already quite long also.
This sums it up. Even SourceMod uses exactly the same names for their CommandLine set of natives. You aren't really even supposed to use the function unless you read the documentation which clearly states how it works.

amxmodx/commandline.cpp
rsKliPPy: Do we really have to do this?

Arkshine:
@WPMGPRoSToTeMa, Sure, ideally, but the logic doesn't necessarily apply to AMXX context.

* You can explicit param if you want
* We don't have object notation, we have to care about name length for readability
* Keeping consistency is also good reason
* Documentation is clear

```SourcePawn
new string[16];
GetCommandLineParamIntOrDefault("-something", string, charsmax(string), "");
new value = GetCommandLineParamIntOrDefault("-something", 0);
new Float:value = GetCommandLineParamFloatOrDefault("-something", 0.0);
```

```SourcePawn
new string[16];
GetCommandLineParamInt("-something", string, charsmax(string));
new value = GetCommandLineParamInt("-something");
new Float:value = GetCommandLineParamFloat("-something");
```

Definitvely the latter. I understand the former is semantically better, but l would prefer to place a consencus on implicit but common and optional default value rather than an ideal prototype but with a cumbersome name.

Also, you can and is encouraged to explicit the param like:
```SourcePawn
new value = GetCommandLineParamInt("-something", .defaultValue = 42);
```

amxmodx/commandline.cpp
WPMGPRoSToTeMa: I can't understand why we need this optimization in function that is not critical to performance. And also who will call this function with empty param?

Arkshine: I can't understand why it's so dramatic to add such check. The engine code doesn't check for an empty string and will loop/compare for nothing. We have an unused length variable available which can be used, even if it's not critical and probability someone passes an empty string is low, why it's so bad to use it?

amxmodx/commandline.cpp
```diff
@@ -0,0 +1,100 @@
...
+#define CHECK_COMMANDLINE_AVAILABILITY()                                        \
+    if (!IsCommandLineAvailable())                                              \
+    {                                                                           \
+        LogError(amx, AMX_ERR_NATIVE, "The command line is not available.");    \
+        return 0;                                                               \
+    }
+
+static bool IsCommandLineAvailable()
+{
+    return g_engfuncs.pfnCheckParm != nullptr;
```
WPMGPRoSToTeMa: This check won't work, because amxmodx will just crash in `GiveFnptrsToDll` or `pfnCheckParm` will be filled with garbage. I mean old metamods, like 1.19 where `pfnCheckParm` isn't present.

rsKliPPy: What solution do you propose?

WPMGPRoSToTeMa: @rsKliPPy looks like metamod developers didn't update any API version that can be checked in plugin, so there is no legal way.

rsKliPPy: So what do we do? Scrap this? Bump Metamod requirement? Something else?

Arkshine: In 1.8.2 release notes, it was clearly stated to upgrade to metamod 1.21-am.
This would not be a surprise that 1.9 would require the same version.

Documenting this properly would be likely enough, but having a way to gracefully warn the user would be welcomed though. If both metamod could bump the API version, we could make 1.9 requiring this version.

WPMGPRoSToTeMa:
Ok, there should be trade-off between explicitness and brevity.

amxmodx/commandline.cpp
WPMGPRoSToTeMa: @Arkshine I need just a comment that this check doesn't change function logic.

amxmodx/commandline.cpp
WPMGPRoSToTeMa: We can remove `IsCommandLineAvailable` native function if amxx supports only last hlds build and metamod >= 1.21-am.

PR #471, opened 2017-12-03
Fix a menu recursion issue

Arkshine:
Reported by JusTGo: https://forums.alliedmods.net/showthread.php?t=275819.

This partially reverts https://bugs.alliedmods.net/show_bug.cgi?id=3199 (https://github.com/alliedmodders/amxmodx/commit/5cb07b900ae566f41f2a6ce00a7605bee48c22e5), a fix is to prevent a context where you would overwrite a default menu (such as buy menu), and once you move out of the zone where such default menu would appear, this essentially fucks up your custom menu in some ways.

By sending `menuselect 10`, this would normally set the player's `m_iMenu` member to 0, therefore avoiding a previous default menu to act. It's actually a fix specific to CS and which doesn't work on a player who is not yet fully joined (team + appearance applied) because `m_iMenu` is only set if you are either fully joined or not stuck in team/appearance menu.

Thre regression happens when you overwrite an initial menu when the player is not yet fully joined, once `menuselect 10` is sent, the game will send again the default menu, and therefore an infinite loop happens. Typically, this should happen only with team and appearance menu.

The proposed fix is simply to use `m_iMenu` directly for CS only intead.

Here a small test plugin to reproduce the error:
```Pawn
#include <amxmodx>

public plugin_init()
{
    register_message(get_user_msgid("ShowMenu"), "OnShowMenu");
}

public OnShowMenu(msgid, dest, id)
{
    static menu_text[64];
    get_msg_arg_string(4, menu_text, charsmax(menu_text));

    if (contain(menu_text, "#Team_Select") >= 0)
    {
        MainMenu(id);
        return PLUGIN_HANDLED;
    }

    return PLUGIN_CONTINUE;
}

public MainMenu(id)
{
    new const menu = menu_create("Menu", "@HandlerMainMenu");
    menu_additem(menu, "Item");
    menu_display(id, menu);
}

@HandlerMainMenu(id, menu, item)
{
    if (++item == 1)
    {
        MainMenu(id);
    }

    menu_destroy(menu);

    return PLUGIN_HANDLED;
}
```

### DIFF
--- a/amxmodx/CGameConfigs.h
+++ b/amxmodx/CGameConfigs.h
@@ -166,6 +166,19 @@ class CGameConfigManager : public IGameConfigManager
 		StringHashMap<ITextListener_SMC*> m_customHandlers;
 };
 
+#define GET_OFFSET(classname, member)												\
+	static int member = -1;															\
+	if (member == -1)																\
+	{                                                                               \
+		TypeDescription type;                                                       \
+		if (!CommonConfig->GetOffsetByClass(classname, #member, &type) || type.fieldOffset < 0)\
+		{																			\
+			LogError(amx, AMX_ERR_NATIVE, "Invalid %s offset. Native %s is disabled", #member, __FUNCTION__);\
+			return 0;																\
+		}																			\
+		member = type.fieldOffset;                                                  \
+	}
+
 extern CGameConfigManager ConfigManager;
 extern IGameConfig *CommonConfig;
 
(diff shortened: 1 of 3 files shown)


Arkshine:
@Nextra, if you don't mind.

Arkshine:
You scared me, I thought I actually used `m_iTeam` in my code.  Not sure why I fuck up each time my commit message!

Arkshine:
Should be fine.

PR #474, opened 2017-12-30
Hungarian language update

regener:
The important parts have been updated* and the missing parts added according to my best knowledge.

*admincmd.txt was UTF8 instead of UTF8 without BOM, I'm not sure it was a mistake or not.


Arkshine:
Thanks!

PR #475, opened 2017-12-31
Add a flags update forward and a override param in set_user_flags

OciXCrom:
Added a forward called "client_flags_updated" which will be called when the client's flags are modified using "set_user_flags" or "remove_user_flags". I also added another param in the "set_user_flags" function called "override", which if set to true will remove any other flags from the user - I found this necessary because the name of the function is "SET_user_flags", but it actually adds them to the current ones. This way coders can actually SET the flags instead of add them to the current ones.

### DIFF
--- a/amxmodx/amxmodx.cpp
+++ b/amxmodx/amxmodx.cpp
@@ -2841,7 +2841,7 @@ static cell AMX_NATIVE_CALL get_user_flags(AMX *amx, cell *params) /* 2 param */
 	return GET_PLAYER_POINTER_I(index)->flags[id];
 }
 
-static cell AMX_NATIVE_CALL set_user_flags(AMX *amx, cell *params) /* 3 param */
+static cell AMX_NATIVE_CALL set_user_flags(AMX *amx, cell *params) /* 4 param */
 {
 	int index = params[1];
 
@@ -2860,8 +2860,13 @@ static cell AMX_NATIVE_CALL set_user_flags(AMX *amx, cell *params) /* 3 param */
 
 	if (id > 31)
 		id = 31;
-
-	pPlayer->flags[id] |= flag;
+		
+	if(params[4])
+		pPlayer->flags[id] = flag;
+	else
+		pPlayer->flags[id] |= flag;
+		
+	executeForwards(FF_ClientFlagsUpdated, static_cast<cell>(pPlayer->index), pPlayer->flags[id]);
 
 	return 1;
 }
@@ -2887,7 +2892,8 @@ static cell AMX_NATIVE_CALL remove_user_flags(AMX *amx, cell *params) /* 3 param
 		id = 31;
 
 	pPlayer->flags[id] &= ~flag;
-
+	executeForwards(FF_ClientFlagsUpdated, static_cast<cell>(pPlayer->index), pPlayer->flags[id]);
+	
 	return 1;
 }
 
--- a/plugins/include/amxmodx.inc
+++ b/plugins/include/amxmodx.inc
@@ -160,6 +160,16 @@ forward client_connect(id);
 forward client_connectex(id, const name[], const ip[], reason[128]);
 
 /**
+ * Called after the client's admin flags are changed.
+ *
+ * @param id        Client index
+ * @param flags     New client flags
+ *
+ * @noreturn
+ */
+forward client_flags_updated(id, flags);
+
+/**
  * Called when the client gets a valid SteamID.
  *
  * @note This may occur before or after client_putinserver has been called.
@@ -1814,7 +1824,7 @@ native task_exists(id = 0, outside = 0);
  * @error           If the index is not within the range of 0 to MaxClients, an
  *                  error will be thrown.
  */
-native set_user_flags(index, flags = -1, id = 0);
+native set_user_flags(index, flags = -1, id = 0, override = false);
 
 /**
  * Returns the client's admin flags as a bitflag sum.
@@ -1826,6 +1836,7 @@ native set_user_flags(index, flags = -1, id = 0);
  *
  * @param index     Client index, 0 to set flags of server
  * @param id        Flag set id, ranging from 0 to 31
+ * @param override  If set to true, all other flags will be removed
  *
  * @noreturn
  * @error           If the index is not within the range of 0 to MaxClients, an
--- a/amxmodx/amxmodx.h
+++ b/amxmodx/amxmodx.h
@@ -333,6 +333,7 @@ extern int FF_InconsistentFile;
 extern int FF_ClientAuthorized;
 extern int FF_ChangeLevel;
 extern int FF_ClientConnectEx;
+extern int FF_ClientFlagsUpdated;
 
 extern bool g_coloredmenus;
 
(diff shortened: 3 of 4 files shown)


plugins/include/amxmodx.inc
```diff
@@ -1826,6 +1836,7 @@ native set_user_flags(index, flags = -1, id = 0);
  *
  * @param index     Client index, 0 to set flags of server
  * @param id        Flag set id, ranging from 0 to 31
+ * @param override  If set to true, all other flags will be removed
```
Amaroq7: Misplaced param's description.

amxmodx/amxmodx.cpp
```diff
@@ -2860,8 +2860,13 @@ static cell AMX_NATIVE_CALL set_user_flags(AMX *amx, cell *params) /* 3 param */
 
 	if (id > 31)
 		id = 31;
-
-	pPlayer->flags[id] |= flag;
+		
+	if(params[4])
```
Nextra: You need to check for parameter availability here.

plugins/include/amxmodx.inc
```diff
@@ -159,6 +159,16 @@ forward client_connect(id);
...
 
+/**
+ * Called after the client's admin flags are changed.
+ *
+ * @param id        Client index
+ * @param flags     New client flags
+ *
+ * @noreturn
+ */
+forward client_flags_updated(id, flags);
```
Amaroq7: I'd also pass id of set here, so we'll know which set of flags has been updated.

plugins/include/amxmodx.inc
OciXCrom: Oops, I always get confused when viewing .inc files because of the way the description is placed above the native. Fixed.

amxmodx/amxmodx.cpp
OciXCrom: Sorry, I'm fairly new to this. How do I exactly add that check?

plugins/include/amxmodx.inc
OciXCrom: I never knew different sets of flag existed until I saw the actual .cpp file. Are different sets ever used? Because if not, it would be rather unnecessary to have to write the set argument every time we use the forward.

amxmodx/amxmodx.cpp
Amaroq7: Here's an example
https://github.com/alliedmodders/amxmodx/blob/7b3646a012e4cc330c17328e13b31fec6a3280a8/amxmodx/amxmodx.cpp#L1443

plugins/include/amxmodx.inc
Amaroq7: We don't have to write all params in forward, they can be skipped, so if someone doesn't depend on sets, they can safely omit last param.

rsKliPPy: I knew they existed but I never knew their purpose. The doc says:
> AMXX stores multiple sets of flags internally, but only flag set
0 is actively used. You should not change the value of the third
parameter from the default.

I almost feel that we should just ignore flag sets and use a single one.

Amaroq7: Even if AMXX doesn't use other values than 0, it still may be used by 3rd plugins for their some specific permission system, so they don't mess up with the AMXX's one. I think including that param wouldn't do much harm, but it's just my opinion.

WPMGPRoSToTeMa: @Amaroq7 we can add it later if someone really uses sets.

OciXCrom: Actually if they can really be skipped when using the forwards (haven't actually checked this), then there's really no problem if that param can be passed too. I'll check this tomorrow and add it if that's the case.

amxmodx/amxmodx.cpp
OciXCrom: Is it actually necessary? The 4th parameter has a default value of "false" if it's not used. Just curious why it's needed.

WPMGPRoSToTeMa: @OciXCrom you will need to recompile all plugins using `get_players`.

Nextra: @OciXCrom Plugins that don't get compiled against the new *.inc will pass you 3 parameters, and core needs to handle that situation gracefully.

OciXCrom: @Nextra @Amaroq7 @WPMGPRoSToTeMa - Thanks, I added the necessary check.

plugins/include/amxmodx.inc
OciXCrom: @Amaroq7 @rsKliPPy @WPMGPRoSToTeMa - I added the flag set as the last parameter in the forward, so people can skip it if not needed. I also passed the old flags so people can see which flags were changed.

OciXCrom:
I added a check whether the new flags are different from the old ones, since there's not really a point in executing the forward if they are the same.

In-line:
I think something like `bool addFlags = true` will be clearer. 

In the first place why should `set_user_flags` do not override?
Maybe deprecating `set_user_flags` and adding 2 more natives will be better?
`add_user_flags`,`override_user_flags`.

OciXCrom:
It should have added them in the first place, but it's too late now. I don't think there's really a point in deprecating it just because of the name.

WPMGPRoSToTeMa:
> I think something like `bool addFlags = true` will be more clear.

Or `SetFlagsMode:mode = SetFlagsMode_Add`.

OciXCrom:
Hmm, I'm starting to like that idea. Maybe we can deprecate both `set_user_flags` and `remove_user_flags` and add a new one that will cover all 3 cases (set, add, remove) with the bool mentioned by @WPMGPRoSToTeMa. This bool can also be added to many other functions, e.g. `set_user_health` for easier usage. Any suggestions for the new flags function?

asherkin:
This PR makes two independent changes, and you're talking about adding even more - each change should be a separate PR. I suggest retargeting this to be only the forward, and make a separate PR if you want to change how `set_user_flags ` works.

OciXCrom:
Well, the PR is mainly about the forward, and the other change is just a small improvement to a function that's calling it. If I made them separate, one would become incompatible if the other one got merged, so I found it to be better this way. I won't be adding any more stuff until this gets approved (if it does). If it's really a problem I'll remove the `set_user_flags` change and make a separate PR after this one gets reviewed.

eat1k:
@Arkshine will it be merged?

plugins/include/amxmodx.inc
```diff
@@ -159,6 +159,18 @@ forward client_connect(id);
...
+ * Called after the client's admin flags are changed.
+ *
+ * @param id        Client index
+ * @param oldflags  Old client flags
+ * @param newflags  New client flags
+ * @param set       Flag set id
+ *
+ * @noreturn
+ */
+forward client_flags_updated(id, oldflags, newflags, set);
```
In-line: Maybe use same terminology that is used in `set_user_flags`?
Client ID is `index`
Set ID is just `id`

voed:
@Arkshine bump. Should it be splitted into 2 PRs?

OciXCrom:
@Arkshine if you wouldn't mind.

PR #476, opened 2018-01-01
Add 'j' and 'k' flags in get_user_flags for matching with admin flags

OciXCrom:
"j" - match with all of the specified admin flags
"k" - match with any of the specified admin flags

### DIFF
--- a/amxmodx/amxmodx.cpp
+++ b/amxmodx/amxmodx.cpp
@@ -2248,6 +2248,7 @@ static cell AMX_NATIVE_CALL get_players(AMX *amx, cell *params) /* 4 param */
 {
 	int iNum = 0;
 	int ilen;
+	int pflags = 0;
 	char* sptemp = get_amxstring(amx, params[3], 0, ilen);
 	int flags = UTIL_ReadFlags(sptemp);
 
@@ -2268,6 +2269,12 @@ static cell AMX_NATIVE_CALL get_players(AMX *amx, cell *params) /* 4 param */
 				team = g_teamsIds.findTeamIdCase(sptemp);
 		}
 	}
+	
+	if ((flags & 512) || (flags & 1024))
+	{
+		sptemp = get_amxstring(amx, params[4], 0, ilen);
+		pflags = UTIL_ReadFlags(sptemp);
+	}
 
 	for (int i = 1; i <= gpGlobals->maxClients; ++i)
 	{
@@ -2281,6 +2288,10 @@ static cell AMX_NATIVE_CALL get_players(AMX *amx, cell *params) /* 4 param */
 			if ((flags & 16) && (pPlayer->teamId != team))
 				continue;
 			if ((flags & 128) && (pPlayer->pEdict->v.flags & FL_PROXY))
+				continue;
+			if ((flags & 512) && ((pPlayer->flags[0] & pflags) != pflags))
+				continue;
+			if ((flags & 1024) && (!(pPlayer->flags[0] & pflags)))
 				continue;
 			if (flags & 32)
 			{
--- a/plugins/include/amxmodx.inc
+++ b/plugins/include/amxmodx.inc
@@ -1387,11 +1387,13 @@ native get_playersnum(flag = 0);
  *                    "g" - match case insensitive
  *                    "h" - do not include HLTV proxies
  *                    "i" - include connecting clients
- * @param team      String to match against if the "e" or "f" flag is specified
+ *                    "j" - match with all of the specified admin flags
+ *                    "k" - match with any of the specified admin flags
+ * @param string    String to match against if the "e", "f", "j" or "k" flag is specified
  *
  * @noreturn
  */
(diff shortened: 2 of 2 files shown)


amxmodx/amxmodx.cpp
```diff
@@ -2282,6 +2289,10 @@ static cell AMX_NATIVE_CALL get_players(AMX *amx, cell *params) /* 4 param */
 				continue;
 			if ((flags & 128) && (pPlayer->pEdict->v.flags & FL_PROXY))
 				continue;
+			if ((flags & 512) && ((pPlayer->flags[0] & pflags) != pflags))
+				continue;
+			if ((flags & 1024) && (!(pPlayer->flags[0] & pflags)))
```
Mistrick: Maybe add consts for all flags?

rsKliPPy:
You should also update `get_players_ex` to match this change.

OciXCrom:
@rsKliPPy - I honestly didn't know that function even existed. Will do it tomorrow.

rsKliPPy:
It's alright, it's a fairly recent addition.

OciXCrom:
@rsKliPPy - done.

plugins/include/amxmisc.inc
```diff
@@ -859,13 +859,15 @@ stock set_task_ex(Float:time, const function[], id = 0, const any:parameter[] =
...
+ *                    GetPlayers_MatchAllFlags - match with all of the specified admin flags
+ *                    GetPlayers_MatchAnyFlags - match with any of the specified admin flags
+ * @param string    String to match against if the "e", "f", "j" or "k" flag is specified
  *
  * @noreturn
  */
-stock get_players_ex(players[MAX_PLAYERS], &num, GetPlayersFlags:flags = GetPlayers_None, const team[] = "")
+stock get_players_ex(players[MAX_PLAYERS], &num, GetPlayersFlags:flags = GetPlayers_None, const string[] = "")
 {
 	new strFlags[10];
```
WPMGPRoSToTeMa: Looks like this should be updated too.

OciXCrom: @WPMGPRoSToTeMa - change what exactly? I changed the name to "string" because "team" doesn't really fit the situation.

WPMGPRoSToTeMa: @OciXCrom 10.

OciXCrom: @WPMGPRoSToTeMa - Ah, I see. Fixed.

plugins/include/amxmodx.inc
```diff
@@ -1387,11 +1387,13 @@ native get_playersnum(flag = 0);
...
  *                    "i" - include connecting clients
- * @param team      String to match against if the "e" or "f" flag is specified
+ *                    "j" - match with all of the specified admin flags
+ *                    "k" - match with any of the specified admin flags
+ * @param string    String to match against if the "e", "f", "j" or "k" flag is specified
  *
  * @noreturn
  */
-native get_players(players[MAX_PLAYERS], &num, const flags[] = "", const team[] = "");
+native get_players(players[MAX_PLAYERS], &num, const flags[] = "", const string[] = "");
```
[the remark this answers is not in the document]
OciXCrom: @Arkshine I changed the param name back to `team`. I don't think however that `get_players` shouldn't be updated. `get_players_ex` is a stock that is depending on `get_players`.

```pawn
stock get_players_ex(players[MAX_PLAYERS], &num, GetPlayersFlags:flags = GetPlayers_None, const team[] = "")
{
	new strFlags[12];
	get_flags(_:flags, strFlags, charsmax(strFlags));
	get_players(players, num, strFlags, team);
}
```

If we don't update `get_players`, this change will have no effect.

Arkshine:
It was fine to keep 'string' for `get_players_ex`, it was introduced in dev recently. If possible we should have a generic term here and actually something a bit more meaningful, maybe 'argument' or 'arg' ; don't know.

OciXCrom:
@Arkshine In my opinion it was fine for `get_players` too. I doubt that anyone will use parameter skipping here since all previous parameters in the function must be written in order for the `team` parameter to even be used. Simply said, `get_players(iPlayers, iPnum, .team = "CT")` is not valid, so chances are nobody will experience problems with backwards compatibility here.

Anyways, I'll change it back to `string` for `get_players_ex` if you want. I chose the name `string` because the description for the param says `string to match against`. I couldn't figure out a more appropriate name either.

Arkshine:
`string` is fine.

What do you mean by `get_players(iPlayers, iPnum, .team = "CT")` not being valid? It _is_ valid.

OciXCrom:
@Arkshine I put back `string` for `get_players_ex`.

I meant that it won't have an effect because the "e" flag isn't specified so most likely people won't do that.

WPMGPRoSToTeMa:
Arg name should also be fixed for `get_playersnum_ex`.

PR #477, opened 2018-01-01
Macedonian & Bulgarian translation fix

OciXCrom:
Macedonian - fixed usage of specific letters (c => ch; s => sh, etc) and some minor mistakes. Some parts were written in Serbian (the author probably used that language as a reference).

Bulgarian - I don't know if the person who made the translation was 8 years old or he simply failed the first grade in school. He litreally took the liberty of replacing letters with numbers (e.g. "E" => "3") and failed basic grammar things.

I also fixed some other minor mistakes that I noticed in other languges, such as missing commas, quotes, etc. Two files were set to UTF8 without BOM, so I changed that too.


rsKliPPy:
> Bulgarian - I don't know if the person who made the translation was 8 years old or he simply failed the first grade in school. He litreally took the liberty of replacing letters with numbers (e.g. "E" => "3") and failed basic grammar things.
I think HL/CS didn't support UTF-8 chat at one point. Even now the console can't display non-ASCII charcaters.

OciXCrom:
Still, "ч" can be written as "ch" instead of "4", which is what I did in both languages. I don't want to convert them to cyrillic because the console won't display those letters.

Freeman-AM:
I'm french and i don't get why you removed some " . " at the end of the sentence. It's correct to end a sentence this way.

Is it for consistency between languages ?

OciXCrom:
It is correct in all languages (I guess), but I did it for consistency. Since the original translation didn't end with a ".", it should be the same for all languages, even though this bugs me too.

Freeman-AM:
Alright ;)

WPMGPRoSToTeMa:
@OciXCrom 
> I don't want to convert them to cyrillic because the console won't display those letters.

Do you mean the server console or ...?

OciXCrom:
@WPMGPRoSToTeMa - the game's console. It doesn't display cyrillic characters.

WPMGPRoSToTeMa:
@OciXCrom client console supports most unicode symbols. The problem occurs when chat message moves into console after timeout. The client code should filter only color codes, but it also filters all non-ASCII characters because the code is bugged. I have a patch for windows and linux clients that fixes this problem.

Arkshine:
Thanks!

PR #479, opened 2018-01-30
Update miscstats.sma

all85100:
Fixed display of your team name when the remaining enemy is displayed


all85100:
error

Soyrce:
error:
Error: Undefined symbol "MAX_PLAYERS" on line 60
Error: Invalid array size (negative or zero) on line 88

Freeman-AM:
Excluding @Soyrce comment, who is reporting errors much more on his side such as outdated includes or some other fail on his side, there is effectively a bug concerning enemy remaining. But updating it this way isn't the way to do i believe. Cause this part is not clear at all, at the first time you read it. A more complete rewrite should be done. I will see what can i do in another PR.

regener:
Could be closed I think, it is fixed since last commits.
https://github.com/alliedmodders/amxmodx/commits/master/plugins/cstrike/miscstats.sma

PR #480, opened 2018-02-10
Added ADMIN_SUPER flag

OciXCrom:
Users with flag "x" (ADMIN_SUPER) are able to bypass the ADMIN_IMMUNITY flag on commands that are properly made with `cmd_target`.

### DIFF
--- a/plugins/include/amxconst.inc
+++ b/plugins/include/amxconst.inc
@@ -94,6 +94,7 @@ public stock const Float:NULL_VECTOR[3];
 #define ADMIN_LEVEL_H       (1<<19) /* flag "t" */
 #define ADMIN_MENU          (1<<20) /* flag "u" */
 #define ADMIN_BAN_TEMP      (1<<21) /* flag "v" */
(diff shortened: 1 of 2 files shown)


rsKliPPy:
All plugins have to be recompiled to make this applicable. If `cmd_target` was a native, and it sadly isn't, it would've been fine. I'd say this is a problem, because some commands will ignore immunity and some won't. We really need more opinions on this one.

OciXCrom:
I don't think that's really a problem. It's probably safer to compile them after updating anyways. Still, compiling all plugins takes only a few seconds, so that shouldn't stop us from introducing a new feature.

afwn90cj93201nixr2e1re:
And what about cases when someone already taked this bit for using it inside amxmodx? Since 1.8.2 updates.
Also brekable.

OciXCrom:
@afwn90cj93201nixr2e1re That's extremely unlikely. Even if it does happen, it's even more unlikely that it will break anything that badly. Plus, these are default unused AMXX flags, so we might as well do something with them.

A similar change was already made with the new "v" flag for tempban.

PR #481, opened 2018-02-20
Build assembly files with AMBuild.

dvander:
This isn't the prettiest thing but it's better than having these awful object files in the tree. Yasm must be installed on the system. (We could alternately require nasm which would remove the codesize hack).

r?'ing DS for build stuff and Arkshine to see if this would disrupt AMX Mod X developers.


dvander:
Actually, I think it's reasonable to require nasm. It's in macports/apt and we can add it to checkout-deps.sh.

In-line:
– Hm. So we have 19 build systems.
– Yes, but no one has features that I need.
– Okey, now we have 20 build systems.

rsKliPPy:
> – Hm. So we have 19 build systems.
> – Yes, but no one has features that I need.
> – Okey, now we have 20 build systems.

This change was long overdue actually. To be honest it should have been this way from the very beginning.

Some Wikis should probably be updated as a consequence of this change.

AMBuildScript
```diff
@@ -369,11 +395,42 @@ class AMXXConfig(object):
     binary = context.compiler.Program(name)
     return self.AddVersioning(binary)
 
+  def AddAssembly(self, context, binary, input_file, output_file, includes=[], extra_argv=[]):
+    if builder.target_platform == 'windows':
+      obj_type = 'win32'
+    elif builder.target_platform == 'linux':
+      obj_type = 'elf'
```
rsKliPPy: Maybe have it as `elf32` as well to match the other two, or remove `32` from the other two. Is there a reason for this one to be different?

Arkshine:
@TheDS If possible, even if it looks good, I would prefer that you say something since david r?'d you.

PR #483, opened 2018-03-01
Fix: fakemeta hook not being unregistered

IgnacioFDM:


Freeman-AM:

Arkshine:
There is an SSL issue when the windows builder tries to access the amxx repository. It's reported and someone will likely fix that soon enough.

Arkshine:
It's now fixed, thanks to pyschonic.

PR #484, opened 2018-03-01
Comment MODULE_RELOAD_ON_MAP_CHANGE on modules that use USE_METAMOD

IgnacioFDM:
Currently amxmodx doesn't reload modules that use metamod, so MODULE_RELOAD_ON_MAP_CHANGE being defined or not doesn't make any difference.
Even though this commit currently doesn't change any behavior, amxmodx might in the future support reloading modules that use metamod, and all these modules in their current state either fail completely because of erroneous cleanup, or are untested (and most likely leak stuff).
Furthermore, these changes make it clear that these modules are in fact not being reloaded.


Arkshine:
Hm okay. At some point, documentation could be updated as well, but fine as it is for now. Thanks.

Freeman-AM:

PR #485, opened 2018-03-01
Fix OnPluginsUnloaded not being called for reloadable modules.

IgnacioFDM:
Modules are loaded before plugins, and can hook OnPluginsLoaded. Therefore, they should be unloaded AFTER plugins, and be able to hook OnPluginsUnloaded. This was not the case for reloadable modules.
This affects nvault module, whose OnPluginsUnloaded function wasn't being called.


Arkshine:
It affects nVault by command only since by default it's not reloaded. 

That's said, it should be good, thanks!

Freeman-AM:

PR #487, opened 2018-03-16
Using negative values for generating random RGB values in (D)HUD messages

OciXCrom:
A very simple change that allows users to generate random colors for their HUD and DHUD messages by using a negative number for any of the RGB components. For example, using `set_hudmessage(-1, -1, -1)` will generate a fully random colored message. Another example is using `set_hudmessage(255, 0, -150)` which will generate a random amount for the blue component ranging from 150 to 255.

### DIFF
--- a/amxmodx/amxmodx.cpp
+++ b/amxmodx/amxmodx.cpp
@@ -480,9 +480,9 @@ static cell AMX_NATIVE_CALL set_hudmessage(AMX *amx, cell *params) /* 11 param *
 	g_hudset.r2 = 255;
 	g_hudset.g2 = 255;
 	g_hudset.b2 = 250;
-	g_hudset.r1 = static_cast<byte>(params[1]);
-	g_hudset.g1 = static_cast<byte>(params[2]);
-	g_hudset.b1 = static_cast<byte>(params[3]);
+	g_hudset.r1 = params[1] < 0 ? RANDOM_LONG(abs(params[1]), 255) : static_cast<byte>(params[1]);
+	g_hudset.g1 = params[2] < 0 ? RANDOM_LONG(abs(params[2]), 255) : static_cast<byte>(params[2]);
+	g_hudset.b1 = params[3] < 0 ? RANDOM_LONG(abs(params[3]), 255) : static_cast<byte>(params[3]);
 	g_hudset.x = amx_ctof(params[4]);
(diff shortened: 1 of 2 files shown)


justgo97:
what happens when some one use bigger value then 255 like -999

OciXCrom:
@justgo97 - the same that happens when you use a bigger value right now.

IgnacioFDM:
Maybe I'm missing something but what's the point? Why not just use random_num() from the plugin?

OciXCrom:
A good example is using a configuration file, cvars or some other method for changing the HUD colors. You will need to manually add a code in the plugin that will check if the value is negative and generate a random number if it is. This makes it easier and I don't see a problem why it shouldn't be added.

IgnacioFDM:
I think it's unneeded but w/e

WPMGPRoSToTeMa:
IMHO full random is bad. It is better to use HSV with fixed Saturation and Value and random Hue (may be without some ranges that aren't displayed clearly).

OciXCrom:
@WPMGPRoSToTeMa - wait, there's hue and saturation? I always thought the RGB color is the only thing that can be edited related to the color.

IgnacioFDM:
You have to do the conversion

rsKliPPy:
RGB can easily be converted to HSV and vice versa. Some operations are just easier and produce "better" results in HSV color space.

OciXCrom:
I'm not very familiar with any of those, I just made it to produce a random value with the current method. By the way, is it possible to produce a HUD message in black?

rsKliPPy:
It is impossible, that's why complete randomness is a bad idea.

OciXCrom:
Exactly, that's why I made it so you can choose from which value to start, instead of it being from 0 to 255 every time, thus having the chance of getting 0 or some other low number which won't display a visible message.

WPMGPRoSToTeMa:
> that's why I made it so you can choose from which value to start, instead of it being from 0 to 255 every time

IMO it gives nothing good. (1, 1, 1) is still almost invisible. If you will start from, for example, 63, then you can't get "pure" color like (255, 0, 0).

Arkshine:
As others people stated, this doesn't seem like a good idea to have that in the core and the way you can choose the start value is quite weird as well.

PR #490, opened 2018-07-06
Fix for precache_event

voed:
Blame to #250


Freeman-AM:
I don't see that much @Arkshine lately. I hope he still there to do some quick reviews about small bugs.

Arkshine:
Thanks!

PR #492, opened 2018-07-11
Fix player transfering

shelru:


Arkshine:
Could you provide some explanation for this, please?

SmileYzn:
register_event also use enable_event and disable_event functions

SmileYzn:
Probably it is because after admin change player team, he can back into old team??

Freeman-AM:
It was already done @SmileYzn since my last commit on this plugin.
i bet the guy didn't even notice that the same offset is used with gamedata way few line after : 
set_ent_data(player, "CBasePlayer", "m_bTeamChanged", true);
It may completly block what he call fix depanding of conditions.
Anyway, I didn't read the whole plugin but double calling the same offset isn't the right way to do.
Plus it misses fakemeta checks.

Arkshine:
It might be related to https://github.com/s1lentq/ReGameDLL_CS/blob/master/regamedll/dlls/client.cpp#L1933-L1941, from `HandleMenu_ChooseTeam`.

Basically, If you already died and changed teams once, the transfer will fail.

shelru:
> Basically, If you already died and changed teams once, the transfer will fail.
Right.

Ok. Bye. But there an error. We must change m_bTeamChanged to false before change team, and remove `if (g_fakemeta)
				{
					set_ent_data(player, "CBasePlayer", "m_bTeamChanged", true);
				}`
Coz we don't need it.

Arkshine:
Why did you close?

shelru:
Coz im lazy;) And i can't find m_bHasChangeTeamThisRound in regamedll(strange).

Arkshine:
As you can read above, `m_bHasChangeTeamThisRound ` is `m_bTeamChanged` ...

PR #495, opened 2018-07-12
Fix a buffer issue in RegisterHam

Arkshine:
Reported by Darthan & KliPPy.

Easy & lazy fix for a classic buffer issue.

`REMOVE_ENTITY` invokes `pfnOnFreeEntPrivateData` which plugins can hook and `function` string is used after that but it is pointing to the AMXX static buffer. Basically, hooking this forward and doing stuff inside which relies on the AMXX buffer, could invalid all RegisterHam calls.

The proposed fix is to copy locally the string.


rsKliPPy:
:boat: \o/

PR #497, opened 2018-07-13
Add new preprocessor constant "__FILE__"

wopox1337:
[The code for the implementation I looked at the guys from SA:MP.](https://github.com/pawn-lang/compiler/commit/d15fc2224be20eac391062d1c1898b027a168713)

Example of work and implementation:
image

Example easy code:
```
    #include <amxmodx>
    
    public plugin_init()
    {
       server_print("^n^t^tConstants:");
       server_print("__DATE__: %s", __DATE__);
       server_print("__TIME__: %s", __TIME__);
       server_print("cellbits: %d", cellbits);
       server_print("cellmax: %d", cellmax);
       server_print("cellmin: %d", cellmin);
       server_print("__Pawn: %d", __Pawn);
       server_print("debug: %d", debug);
       server_print("__LINE__: %d", __LINE__);
       server_print("__BINARY_PATH__: %s", __BINARY_PATH__);
       server_print("__BINARY_NAME__: %s", __BINARY_NAME__);

       server_print("__FILE__: %s", __FILE__);
    }
```

### DIFF
--- a/compiler/libpc300/sc1.c
+++ b/compiler/libpc300/sc1.c
@@ -585,9 +585,9 @@ int pc_compile(int argc, char *argv[])
             sc_needsemicolon ? "true" : "false",
             sc_tabsize);
     pc_writeasm(outf,string);
-    setfiledirect(inpfname);
   } /* if */
   /* do the first pass through the file (or possibly two or more "first passes") */
+  setfiledirect(inpfname);
   sc_parsenum=0;
   inpfmark=pc_getpossrc(inpf_org);
   do {
@@ -612,6 +612,7 @@ int pc_compile(int argc, char *argv[])
     fline=skipinput;            /* reset line number */
     sc_reparse=FALSE;           /* assume no extra passes */
     sc_status=statFIRST;        /* resetglobals() resets it to IDLE */
+	setfileconst(inpfname);
     if (strlen(incfname)>0) {
       if (strcmp(incfname,sDEF_PREFIX)==0) {
         plungefile(incfname,FALSE,TRUE);    /* parse "default.inc" */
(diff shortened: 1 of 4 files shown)


compiler/libpc300/sc1.c
```diff
@@ -612,6 +612,7 @@ int pc_compile(int argc, char *argv[])
     fline=skipinput;            /* reset line number */
     sc_reparse=FALSE;           /* assume no extra passes */
     sc_status=statFIRST;        /* resetglobals() resets it to IDLE */
+	setfileconst(inpfname);
```
Arkshine: :pray:

Arkshine:
Looks like you can do the same way as `__BINARY_PATH__` has been implemented. I would prefer this way since it doesn't require a new function such as `add_constant_string` that I can't review. I will do it later.

We should probably have instead `__FILE_PATH__` and  `__FILE_NAME__` similar to `__BINARY_PATH__` and __BINARY_NAME__.

I'm concerned about the PATH version of both though. Is it actually useful to know the whole path? Do you guys see an usage?

Arkshine:
After more thinking, right now we have:

* `__BINARY_NAME__`	: admin.amxx
* `__BINARY_PATH__`	:  /home/hlds/cstrike/addons/amxmodx/plugins/admin.amxx
* `__FILE__ ` :  myplugin.sma

I think we should remove  `__BINARY_PATH__`. There are two reasons: 
1) The path is set at compilation time, which means it can be any location (own computer or server) on either windows, linux or osx. From server-side point of view, this information is pointless. 
2) Some people is not going to like having their directories displayed (especially if compiled from your computer)

About `__FILE__`, It would be nice to have `__FILE_PATH__` where the path would be relative to `/include`.  Actually the path given after the `#include`. It can be useful if your plugin includes one and more files, and among them some have the same name (`/dir1/file.sma`, `/dir2/file.sma`).

PR #498, opened 2018-07-15
The "&id" parameter in LookupLangKey() should be just "id"

OciXCrom:
The id parameter doesn't have to and isn't supposed to save the value in a variable, so the & in front is wrong. This argument is meant to tell the function which language to use, not to extract the language from the lang file which doesn't make sense.


Arkshine:
You can't remove it if we want to keep backward compatibility.

Originally `&` is used because internally a function was requiring the parameter address so `id` is passed by reference (it is for sure an arguable design decision).

In a recent PR, we cleaned up the code but we still keep the `&` for backward compatibility reason. If an old compiled plugin would execute the native after your PR, the id value passed will be the address of the parameter instead of the actual value and the new code won't be able to find the player's language (internally you would need to retrieve the actual value from the address).

OciXCrom:
But, was this function ever passing the parameter by reference or was it the same from the beginning? I checked amxmodx.cpp and right now it doesn't seem to be requiring the & tag. I don't think that much will change if we remove the & if nobody ever used the function to retreive the id parameter.

Arkshine:
It was originally implemented with `&`, because of the reason stated above. This function is used often and is used in base plugins as well.

Removing it now means breaking backward compatibility for old compiled plugins and for the next 1.9 version, I would prefer to keep it. It could be planned in another major version, but at the moment, the next version should not change any behavior or break functionalities.

Nextra:
The & in the .inc causes the compiler to generate different code, passing an address instead of the value. It does not matter that AMXX Core does not write to the address: It still uses `get_amxaddr` to retrieve the value indirectly. This can not change, as this behavior was introduced some 11 years ago in 201d65749e849c08e7826cc96145ff71810f4d99

WPMGPRoSToTeMa:
We can make it `const` again. I can't understand why #222 removed `const`.

Arkshine:
Good question. Don't remember why it has been changed. But I agree `const` should be kept. Because 1) `&` is a trick for an internal use only, it makes sense to say that value can't change  2) it breaks backward compatibility as you stated.

Thanks for reporting! Will fix ASAP.

PR #501, opened 2018-07-23
nvault.inc documentation fix

OciXCrom:
Almost all of the natives in this file are marked with `@noreturn` even though all of them have proper return values in different scenarios, especially the `nvault_get` function.


Arkshine:
You don't need to explicit a `@return` when the only 'error' value is preceded with a runtime error. You will never use that return value.

OciXCrom:
You're right. Should I remove those changes or let them be? The `nvault_get` and `nvault_prune` have proper return values that were missing.

Arkshine:
Well, yes, you should keep only the relevant information and without the handle part.

OciXCrom:
Done.

Arkshine:
I guess it's okay. Thanks!

PR #503, opened 2018-07-24
Fix a buffer issue in server_changelevel forward caused by engine_changelevel native

Arkshine:
Related to #104.
Reported on the forum: https://forums.alliedmods.net/showthread.php?t=309368.

Again another buffer issue.


rsKliPPy:
:sheep:

PR #504, opened 2018-07-24
Fix client_print_color where the color is ignored if string starts with an unicode character

Arkshine:
Related to #338.
Reported on the forum: https://forums.alliedmods.net/showthread.php?t=309314

Because of the conversion `cell` -> `char` in `format_amxstring`, utf-8 unsiged bytes will be interpreted as negative value, ie `0xc3` -> `0xffffc3`, resulting in the check failure.

Proposed fix is to cast the char as unsigned.


rsKliPPy:
:sheep: :dash:

PR #506, opened 2018-07-24
Add @note in get_user_msgid()

OciXCrom:
This function should not be used inside plugin_precache(), as it will return 0 if the server was just started.

https://forums.alliedmods.net/showthread.php?t=308592


plugins/include/amxmodx.inc
```diff
@@ -2142,6 +2142,8 @@ native random_num(a, b);
  * @note Example usage: get_user_msgid("TextMsg")
  * @note The message id is unique as long as the server is running, but might
  *       change between updates. They should not be hardcoded into plugins.
+ * @note This function should not be used inside plugin_precache(), as it
```
Arkshine: It's kind of weirdly phrased. The context should be first. My English is horrible but I would go like:

 _On first server start, this function returns 0 inside plugin_precache(). Consider hooking RegUserMsg in order to retrieve the message id_.

Something like that with a better english if not good.

OciXCrom: @Arkshine Done.

PR #464, opened 2017-10-08
Fixed passing a temporary string to LIGHT_STYLE

WPMGPRoSToTeMa:

PR #508, opened 2018-07-28
Fix a buffer issue in EngFunc_LightStyle

Arkshine:
Reported on the forum: https://forums.alliedmods.net/showthread.php?t=309459.
Related to #464.

Similar to `set_lights` issue.  The original engine saves the AMXX string pointer as it is, and therefore if the value in the AMXX buffer changes, it will reflect on the light as well.

I believe ReHLDS doesn't have this issue since the string is copied in another array.

### DIFF
--- a/modules/fakemeta/engfunc.cpp
+++ b/modules/fakemeta/engfunc.cpp
@@ -12,8 +12,11 @@
 //
 
 #include "fakemeta_amxx.h"
+#include <engine_strucs.h>
 
 TraceResult g_tr;
+
+ke::AString LightStyleBuffers[MAX_LIGHTSTYLES];
 
 //by mahnsawce from his NS module
 static cell AMX_NATIVE_CALL engfunc(AMX *amx, cell *params)
@@ -595,8 +598,13 @@ static cell AMX_NATIVE_CALL engfunc(AMX *amx, cell *params)
 	case	EngFunc_LightStyle:			// void )			(int style, const char* val);
 		cRet = MF_GetAmxAddr(amx,params[2]);
 		iparam1=cRet[0];
-		temp = MF_GetAmxString(amx,params[3],0,&len);
-		(*g_engfuncs.pfnLightStyle)(iparam1,temp);
+		if (iparam1 < 0 || iparam1 >= ARRAYSIZE(LightStyleBuffers))
+		{
+			MF_LogError(amx, AMX_ERR_NATIVE, "Invalid style %d", iparam1);
+			return 0;
+		}
+		LightStyleBuffers[iparam1] = MF_GetAmxString(amx, params[3], 0, &len);
+		(*g_engfuncs.pfnLightStyle)(iparam1, LightStyleBuffers[iparam1].chars());
 		return 1;
 
 


justgo97:
In my old tests this bug happens with ReHLDS too, unless they changed somthing in latest versions it should occur with the test plugin

Arkshine:
It has been fixed: https://github.com/dreamstalker/rehlds/pull/532

modules/fakemeta/engfunc.cpp
```diff
@@ -595,8 +596,8 @@ static cell AMX_NATIVE_CALL engfunc(AMX *amx, cell *params)
 	case	EngFunc_LightStyle:			// void )			(int style, const char* val);
 		cRet = MF_GetAmxAddr(amx,params[2]);
 		iparam1=cRet[0];
-		temp = MF_GetAmxString(amx,params[3],0,&len);
-		(*g_engfuncs.pfnLightStyle)(iparam1,temp);
+		string = MF_GetAmxString(amx,params[3],0,&len);
+		(*g_engfuncs.pfnLightStyle)(iparam1, string.chars());
```
WPMGPRoSToTeMa: `string` is temporary, `LightStyle` requires static string.

Arkshine: What about `ALLOC_STRING`? Or should we use a global var specifically for that?

modules/fakemeta/engfunc.cpp
```diff
@@ -12,9 +12,12 @@
 //
 
 #include "fakemeta_amxx.h"
+#include <engine_strucs.h>
 
 TraceResult g_tr;
 
+char LightStyleBuffers[MAX_LIGHTSTYLES];
```
WPMGPRoSToTeMa: I meant an array of `AString`, not `char`. `MAX_LIGHTSTYLES` means lightstyle count, not lightstyle max size like `MAX_LIGHTSTYLE_SIZE`.

modules/fakemeta/engfunc.cpp
```diff
@@ -595,8 +598,9 @@ static cell AMX_NATIVE_CALL engfunc(AMX *amx, cell *params)
 	case	EngFunc_LightStyle:			// void )			(int style, const char* val);
 		cRet = MF_GetAmxAddr(amx,params[2]);
 		iparam1=cRet[0];
-		temp = MF_GetAmxString(amx,params[3],0,&len);
-		(*g_engfuncs.pfnLightStyle)(iparam1,temp);
+		cRet = MF_GetAmxAddr(amx, params[3]);
+		strncopy(LightStyleBuffers, cRet, sizeof(LightStyleBuffers));
+		(*g_engfuncs.pfnLightStyle)(iparam1, LightStyleBuffers);
```
WPMGPRoSToTeMa: So, `iparam1` should be an index to `LightStyleBuffers` array.

Arkshine:
Does it look good now,  @WPMGPRoSToTeMa?

WPMGPRoSToTeMa:
:shipit:

PR #509, opened 2018-07-28
Add constants for stats functions

OciXCrom:
> 0 - Kills
> 1 - Deaths
> 2 - Headshots
> 3 - Teamkills
> 4 - Shots
> 5 - Hits
> 6 - Damage
> 7 - Rank

> 0 - total defusions
> 1 - bomb defused
> 2 - bomb plants
> 3 - bomb explosions

These numbers are the same for all stats functions present in the csstats.inc file and the stats libraries for other games as well. I decided to add stats for accessing the data without having to use "magic numbers" that need to be memorized or searched in the files.

I added them in amxconst.inc marked with STATS_*.

> /* example usage */
> new iStats[8], iHits[8]
> get_user_stats(id, iStats, iHits)
> client_print(id, print_chat, "Your rank is %i with %i kills and %i deaths.", iStats[STATS_RANK], iStats[STATS_KILLS], iStats[STATS_DEATHS])


Arkshine:
If you are using an enum, as it is, you can be sure there will be name collisions due to plugins declaring same constants with defines.

So either you use defines as well (redeclaration of defines with same value doesn't throw a warning), or you rename the prefix like `STATSX_`. I don't have a preference, both are fine, might be better to rename to avoid any potential issues though.

Also, having a constant for all types to replace the hardcoded `8` would be welcomed. Not sure how to name it though.

OciXCrom:
@Arkshine I changed them to STATSX_*. About the 8 - is it actually supposed to be 8 in all of the functions? Take for example the `get_user_astats` function - the documentation says it only has 6 fields:

>* @note The fields in the statistics are:
>*          0 - Kills
>*          1 - Deaths
>*          2 - Headshots
>*          3 - Teamkills
>*          4 - Shots
>*          5 - Hits
>*          6 - Damage

...so the argument should be 7, not 8. Furthermore, the dodstats.inc file even has 9 for some reason.

Anyways, I chose these two names as defines:

>#define MAX_STATS_CLIENT 8
>#define MAX_STATS_OBJECTIVE 4

Arkshine:
8 is just a max, some natives just don't set the rank.

For dod, we should probably have a different enum based on the default values.
Looks like a points field is inserted before the rank.
Constants name could start with DODX_ or maybe STATSX_DOD_, not sure.
Like:
```
enum
{ 
    STATSX_DOD_KILLS = STATSX_KILLS,
    ...
    STATSX_DOD_POINTS,
    STATSX_DOD_RANK
```

You will need new max for DoD as well.

By the way, you replaced the limit for hitbody, but it's wrong. It's based on HIT_* constants.
We don't have yet a max constants for them. Might as well add one generic (8) and one for CS (9).

Anything can be discussed.

OciXCrom:
@Arkshine There, it should be fine now. I added these in dodconst.inc:

```pawn
/* DoD stats constants */
enum {
	DODX_KILLS = 0,
	DODX_DEATHS,
	DODX_HEADSHOTS,
	DODX_TEAMKILLS,
	DODX_SHOTS,
	DODX_HITS,
	DODX_DAMAGE,
	DODX_POINTS,
	DODX_RANK
}

/* Constants for maximum number of stats in DoD */
#define DODX_MAX_STATS 9
#define DODX_MAX_BODYHITS 8
#define DODX_MAX_OBJECTIVE 4
```

And these in amxconst.inc and I used them in the stats files for the other games:

```pawn
/**
 * Constants for maximum number of stats
 */
#define STATSX_MAX_STATS 8
#define STATSX_MAX_BODYHITS 8
#define STATSX_MAX_OBJECTIVE 4
```

WPMGPRoSToTeMa:
What about changing macro-constants to `const`?

Arkshine:
Yea, I don't mind `const`.

rsKliPPy:
I would honestly `const`-ify everything as a new PR if that change wouldn't break anything.

OciXCrom:
@WPMGPRoSToTeMa @Arkshine

If you're talking about changing `define` to `const` in all .inc files, I guess it can be done in a new PR as @rsKliPPy said.

Arkshine:
Few things in my mind:
* Is there a reason `_MAX_` is not inside the `enum` as well?
* It would be likely better to create `csstats_const.inc` 
* You defined the max objectives, but looks like there is no enum for them

OciXCrom:
@Arkshine 

1. I figured it's more appropriate to leave only the possible return values in the `enum` and have the `_MAX_` ones outside so users don't get confused when looking at the `enum`. Shall I put the `_MAX_` inside too?

2. Only the objective based stats are specific for `cstrike`, it would be rather unnecessary to have an entire file for only 5 lines of code. I think it would be better to move them in `cstrike_const.inc` instead.

3. What do you mean? It's right there in `amxconst.inc`:

```pawn
/**
 * Constants for objective based statistics
 */
enum
{
	STATSX_TOTAL_DEFUSIONS = 0,
	STATSX_BOMBS_DEFUSED,
	STATSX_BOMBS_PLANTED,
	STATSX_BOMB_EXPLOSIONS
}
```

Arkshine:
1. Well, my reasoning is the value itself is not important and is the result of the number of stats types. It would be logic to find this constant at the end of the `enum` since it's auto-incremented. If you prefer, you can have a new line before, and/or even a comment. This constant will be displayed also in the concerned natives, so there is a little chance it would be confusing. Unless there are some unforeseen advantages to separate, I think inside the enum is fine.

2. `cstrike `!= `csx`. Different modules, different includes. There is no issue creating a new file, it's invisible for the user. `amxconst.inc` should be only for generic/core constants. It makes sense to have a new `csstats_const.inc` file here. DoD module has already its own `dodconst.inc` file.

3. You're right, probably did not expect to find them in `amxconst.inc`.

OciXCrom:
@Arkshine 

There, I moved the `_MAX_` ones to the `enum`, except `STATSX_MAX_BODYHITS` because its number is out of the `enum` sequence so it won't look pretty.

I also created `csstats_const.inc` and moved the objective based constants in it.

Arkshine:
Is there a reason you did not apply the change on `DODX_` and `STATSX_` enum is not moved?

OciXCrom:
@Arkshine 

I forgot about DoD, it's fixed now. I didn't move the `STATSX_` enum because it's global for all games, so it should stay in `amxconst.inc`. Moving it to `csstats_const.inc` will make it for CS only. I changed the `HIT_` constant too.

Welp, I didn't expect such a simple PR to get so complicated.

Arkshine:
Yeah, if you don't do things properly at first try, it will be ugly to fix later on. Thanks for your patience!

Arkshine:
I believe some stat plugins could be updated as well to reflect those changes, but it can be done later in another PR.

SmileYzn:
Description from get_user_astats 

`@param wpnindex  Attacker client index, or 0 to retrieve the statistics from all attackers`

What should be? wpnindex >> attacker ??

OciXCrom:
@SmileYzn hmm, that's a bummer. I believe it should be "attacker" because the function requires 2 client indexes, not a weapon index. Not sure how we should proceed about this since the mistake was not made in this PR, it was there from before. On one hand, changing the argument name can break backwards compatibility, but on the other hand it's wrong to leave it like this. @Arkshine

PR #511, opened 2018-08-01
Add default values for menu_item_getinfo's arguments

OciXCrom:
This function has many arguments and 99% of the time people don't use all of them at once. Here's how I personally almost always use this function:

> new szKey[5], szUnused[2], iUnused
> menu_item_getinfo(iMenu, iItem, iUnused, szKey, charsmax(szKey), szUnused, charsmax(szUnused), iUnused)

This is just ridiculous. I need to create variables named "unused" just to store the values that I don't need from the function. Adding default values to those arguments would make this function much easier to use:

> new szKey[5]
> menu_item_getinfo(iMenu, iItem, _, szKey, charsmax(szKey))

### DIFF
--- a/plugins/include/newmenus.inc
+++ b/plugins/include/newmenus.inc
@@ -221,7 +221,7 @@ native menu_find_id(menu, page, key);
  * @return				1 on success, 0 on failure.
  * @error				Invalid menu resource.
  */
-native menu_item_getinfo(menu, item, &access, info[], infolen, name[]="", namelen=0, &callback);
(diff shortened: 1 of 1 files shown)


Arkshine:
At first, I was going to say it doesn't make much sense to allow `menu_item_getinfo(menu, item)` and it was not dramatic to require at least the first param, but this function can be used to check if a menu item is valid as well from the return value. So, I guess it's okay having all the params optional.

rsKliPPy:
I was always wondering why we haven't done this each time I use this native.

PR #512, opened 2018-08-03
Execute amxx.cfg before plugin_init to keep compatibility after #266

Arkshine:
Reported on the forum.
Related to #266.

Originally, the `amxx.cfg` file was executed in `admin.sma` and specifically inside `plugin_init`.
This means any plugins declared after could use their cvars in `plugin_init` with values updated from `amxx.cfg`.

After #266, the file was executed after all `plugin_init` calls resulting in a behavior change. Not a dramatic issue, more an inconvenience.

Proposed fix is to execute `amxx.cfg` before `plugin_init`. I don't see any downside to this, or is there one?

### DIFF
--- a/amxmodx/meta_api.cpp
+++ b/amxmodx/meta_api.cpp
@@ -657,10 +657,11 @@ void C_ServerActivate_Post(edict_t *pEdictList, int edictCount, int clientMax)
 		pPlayer->Init(pEdictList + i, i);
 	}
 
+	CoreCfg.ExecuteMainConfig();    // Execute amxx.cfg
+
 	executeForwards(FF_PluginInit);
 	executeForwards(FF_PluginCfg);
 
-	CoreCfg.ExecuteMainConfig();    // Execute amxx.cfg
 	CoreCfg.ExecuteAutoConfigs();   // Execute configs created with AutoExecConfig native.
 	CoreCfg.SetMapConfigTimer(6.1); // Prepare per-map configs to be executed 6.1 seconds later.
 	                                // Original value which was used in admin.sma.


justgo97:
whats the point of having amxx.cfg before plugin_init ? 
i mean for exemple if i have a plugin which registres a cvar in plugin_init and is set to a default value will the cvar change if i add it to amxx.cfg with a different value ?

WPMGPRoSToTeMa:
I think this has only effect on startup (or when new plugin is loaded), because `register_cvar` sets value only for new cvar, not for already registered. (cvars live until the server is fully rebooted)

WPMGPRoSToTeMa:
Does amxx 1.8.3 `register_cvar` change cvar value even if it is already exist? In 1.8.2 it doesn't.

WPMGPRoSToTeMa:
`register_cvar` can change cvar value only on startup, isn't it? So, we will have different cvar values on startup and mapchange?

Arkshine:
This PR is merely keeping a similar behavior before #266.

Example:
* You register the cvars in `plugin_precache` and do stuffs in `plugin_init`.
  * _Before_: Your cvars values in `plugin_init` are updated from the config file
  * _After_: They are not

Whatever at first server start or after an edit in the config, the behavior has changed and you will need to restart/changemap depending the stuff you are doing in `plugin_init` (someone reported the issue yesterday: https://forums.alliedmods.net/showthread.php?t=309682)

Another example, if you register and do stuffs in `plugin_init` and you edit the config file, you will need to restart/changemap two times, whereas before only one was enough.

It's probably better to execute the `amxx.cfg` after `plugin_init` but the behahior change is going to be annoying with the existing plugins. Also, people should stop doing stuff with cvar in this forward, and are encouraged instead to use `AutoExecConfig` and/or `OnConfigsExecuted`.

WPMGPRoSToTeMa:
In my opinion the correct behavior should be as follows:
1. The value in `register_cvar` should have low priority.
2. The value in config should have medium priority.
3. And the value that is set directly from the plugin code should have high priority.
These rules are only for `plugin_precache`-`plugin_cfg` period.
Idk how to implement this and can it break plugins. But I think this is pretty expected behavior.

Arkshine:
It's already more or less what it does currently, but priority is not enforced for sure and `amxx.cfg` is originally executed in a weird place.

Anyway, the next AMXX version should not have any behavior change. We could consider for another version to clean/fix/improve things and prepare/notify people of a possible behavior change, but right now, the only purpose of this PR is just to have a similar behavior as before. I would like to avoid people reporting bugs as someone already did on the forum.

rsKliPPy:
If no one has any objections then we can :ship: it, I guess this is the preferred behavior for the best backwards compatibility.

PR #513, opened 2018-08-07
Fix wrong message type information for engclient_print in amxmodx.inc

OciXCrom:
>@param type      Message type, see print_* destination constants in

`print_*` constants are invalid for `engclient_print` and will output either an error or a wrong destination type when used. The right constants to use are `engprint_*`.


Arkshine:
Thanks!

PR #514, opened 2018-08-09
Add get_user_rendering

OciXCrom:
Apart from `set_user_rendering` being the only function from the fun module that doesn't have a counter function, `get_user_rendering` is extremely needed in many situations. Take for example a plugin that attaches an entity to the player - we should be able to get the player's rendering so we can set the same rendering on the entity.

### DIFF
--- a/plugins/include/fun.inc
+++ b/plugins/include/fun.inc
@@ -126,6 +126,26 @@ native set_user_origin(index, const origin[3]);
 native set_user_rendering(index, fx = kRenderFxNone, r = 0, g = 0, b = 0, render = kRenderNormal, amount = 0);
 
 /**
+ * Gets player's rendering mode.
+ *
+ * @note A really useful render modes reference:
+ *		 https://sites.google.com/site/svenmanor/rendermodes
+ *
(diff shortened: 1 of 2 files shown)


Arkshine:
I guess it's okay, the native makes sense. Looks good.

plugins/include/fun.inc
```diff
@@ -125,6 +125,26 @@ native set_user_origin(index, const origin[3]);
...
+ * @param g             Variable to store the amount of green color
+ * @param b             Variable to store the amount of blue color
+ * @param render        Variable to store the render mode
+ * @param amount        Variable to store the render amount
+ *
+ * @noreturn
+ * @error               If player is not connected or not within the range
+ *                      of 1 to MaxClients.
+ */
+native get_user_rendering(index, &fx = 0, &r = 0, &g = 0, &b = 0, &render = 0, &amount = 0);
```
Arkshine: You could use the same default constant `kRenderFxNone` and `kRenderNormal`.

OciXCrom: @Arkshine Done.

PR #515, opened 2018-08-09
[WIP] Third attempt for cmd_targetex

OciXCrom:
First attempt was made by HamletEagle.
Second attempt is over here and looks abandoned - #368
So, here's the third attempt to add this stock into AMXX.

Bear in mind that I haven't done too much testing on all the specific cases featured with the stock, but I haven't noticed any bugs so far. The PR is "work in progress", so I would appreciate some feedback on it. I decided to post it earlier since I'm leaving on a vacation tomorrow and the stock is *almost* complete.

For those of you that don't know, the goal of this stock is to allow admins to use arguments such as @all, @ct, @t and so on with player targeting commands. Here's a list of all arguments that I made for the stock:

- @aim - targets the player that the admin is aiming at
- @all - targets all players
- @alive - targets alive players
- @bots - targets all bots
- @dead - targets dead players
- @humans - targets all humans
- @me - targets the admin himself
- @spectating - targets the client that the admin is spectating
- @view - targets all clients in the admin's field of view

In addition, a team name can be passed into any argument that is supposed to target a group of players. For example, @alivect will target all alive players.

Another addition is the option to exclude yourself from the command by adding the "!" symbol right after the "@". For example, @!all will target everyone except the admin who used the command.

I decided to create a separate .inc file because it requires engine.inc. I'm not sure if the compiler will actually read any of the natives in engine.inc if `cmd_targetex` is used in the plugin, if not, we may move it in amxmisc.inc where the original `cmd_target` is.

Let me know if there's something that can be optimized (there probably is) and if you have any ideas about adding additional arguments (@closest and @radius are ones I have in mind).

### DIFF
--- a/plugins/include/targetex.inc
+++ b/plugins/include/targetex.inc
@@ -1,0 +1,335 @@
+// vim: set ts=4 sw=4 tw=99 noet:
+//
+// AMX Mod X, based on AMX Mod by Aleksander Naszko ("OLO").
+// Copyright (C) The AMX Mod X Development Team.
+//
+// This software is licensed under the GNU General Public License, version 3 or higher.
+// Additional exceptions apply. For full license details, see LICENSE.txt or visit:
+//     https://alliedmods.net/amxmodx-license
+
+//
+// cmd_targetex stock
+//
+
+#if defined _targetex_included
+	#endinput
+#endif
+#define _targetex_included
+
+#include <amxmisc>
+#include <engine>
+
+static const TargetEx_GroupArguments[][] = { "all", "alive", "bots", "dead", "humans", "view" }
+static TargetEx_ArgLens[sizeof(TargetEx_GroupArguments)];
+
+enum TargetexFlags (<<= 1)
+{
+	TARGETEX_NONE = 0,
+	TARGETEX_OBEY_IMM_SINGLE = 1,
+	TARGETEX_OBEY_IMM_GROUP,
+	TARGETEX_NO_SELF,
+	TARGETEX_NO_GROUPS,
+	TARGETEX_NO_BOTS,
+	TARGETEX_NO_ALIVE,
+	TARGETEX_NO_DEAD
+}
+
+enum TargetexTeams
+{
+	TARGETEX_E_CODE[5],
+	TARGETEX_E_NAME[10],
+	TARGETEX_E_SUFFIX[6],
+	TARGETEX_E_LEN,
+	TARGETEX_E_LEN_PLUS_ONE,
+}
+
+static const Targetex_Teams[][TargetexTeams] =
+{
+	{ "t", "TERRORIST", "_T" },
+	{ "ct", "CT", "_CT" },
+	{ "spec", "SPECTATOR", "_SPEC" }
+}
+
+/**
+ * Processes a generic target pattern and tries to match it to a client or a
+ * group of clients based on filtering flags and the usage of special arguments.
+ *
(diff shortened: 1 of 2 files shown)


rsKliPPy:
This can probably just be in `amxmisc.inc`, we don't need a new file.
I haven't checked thoroughly, but because this is in "core" it shouldn't work only for CS teams, it should be game independent. Either through some configuration file, somehow, or with natives to register a new `@selector`.

OciXCrom:
It's because it includes engine.inc as well, I don't know if that's a big deal or not.

It won't be only for CS. I didn't make the team names hardcoded, instead I added them in an enum list which can be filled with the team names for other games and it will work just fine. I don't know the names for the other games, so a list of them would be appreciated.

I agree it's better if it was a native, but my C++ "skills" aren't too good, I even struggled doing it in Pawn and we can see that other people did too (the previous attempts).

OciXCrom:
@rsKliPPy It's because it includes engine.inc as well, I don't know if that's a big deal or not.

It won't be only for CS. I didn't make the team names hardcoded, instead I added them in an enum list which can be filled with the team names for other games and it will work just fine. I don't know the names for the other games, so a list of them would be appreciated.

I agree it's better if it was a native, but my C++ "skills" aren't too good, I even struggled doing it in Pawn and we can see that other people did too (the previous attempts).

OciXCrom:
I removed the [WIP] tag since I think it's pretty much good to go from coding perspective. I'll just need a list of all the team names used in AMXX games so I can add them in the file.

```pawn
static const Targetex_Teams[][TargetexTeams] =
{
	{ "ct", "CT" },
	{ "t", "TERRORIST" },
	{ "spec", "SPECTATOR" }
}
```

I did a bunch of testing and no problems were found.

OciXCrom:
@rsKliPPy There won't be any problem if an invalid team name is used. It will simply print in the console that no players were found. It uses `get_players_ex` to find the players and this function doesn't produce errors or any kind of problems if the team name doesn't exist. With that being said, if you try using a CS team in DoD for example, it will simply tell you that no players were found.

WPMGPRoSToTeMa:
@OciXCrom for team filters I see two approaches:
1. If text after @ can't be matched with any pattern, then treat it as team name and pass it to `get_players_ex`.
2. Add prefix for team patterns like `@team_%teamname%` or `@team%teamname%`, where `%teamname%` is a team name.

OciXCrom:
@WPMGPRoSToTeMa @rsKliPPy Right now it's working like this:

1. It's looping the entire array that holds the teams and if it finds a match it adds `_TEAM` to the language key (e.g. `ct`).
2. It checks if any other argument is specified (e.g. `@alivect`). If not (it's only `@ct`), it proceeds directly to affecting the entire team.

This way there's no errors or wrong matches. The group argument names don't collide with the team names. The only reason I don't like the second approach is because the patterns are longer and admins will need to write more text (e.g. `@team_ct`), whereas typing only `@ct` is much convenient and used in custom plugins at the moment so people are already used to it.

WPMGPRoSToTeMa:
@OciXCrom IIRC Half-Life has dynamic teams, so I think it's not possible to store teams for all the games.

OciXCrom:
@WPMGPRoSToTeMa Weren't the teams in HL named RED and BLUE? If they can't be added we'll just ignore them, not a big deal. It's fine as long as they're working for CS and other games that have valid team names. We just need to dig out the list.

Nord1cWarr1or:
Any news?

PR #517, opened 2018-08-17
Add get_playersnum_ex

justgo97:
example code : 

```
#include <amxmodx>
#include <amxmisc>

public plugin_init( )
{
    register_concmd( "GetAliveCount", "cmd_GetAliveCount" )
}

public cmd_GetAliveCount( )
{
    new AliveCount = get_playersnum_ex(GetPlayers_ExcludeDead)

    server_print("AliveCount:%d", AliveCount)
}
```

### DIFF
--- a/plugins/include/amxmisc.inc
+++ b/plugins/include/amxmisc.inc
@@ -869,3 +869,31 @@ stock get_players_ex(players[MAX_PLAYERS], &num, GetPlayersFlags:flags = GetPlay
 	get_flags(_:flags, strFlags, charsmax(strFlags));
 	get_players(players, num, strFlags, team);
 }
+
+/**
+ * Returns the number of clients on the server that matches the specified flags.
+ *
+ * @note Example retrieving all alive CTs:
+ *       new AliveCt = get_playersnum_ex(GetPlayers_ExcludeDead | GetPlayers_MatchTeam, "CT")
+ *
+ * @param flags     Optional filtering flags (enum GetPlayersFlags); valid flags are:
+ *                    GetPlayers_None - No filter (Default)
+ *                    GetPlayers_ExcludeDead - do not include dead clients
+ *                    GetPlayers_ExcludeAlive - do not include alive clients
+ *                    GetPlayers_ExcludeBots - do not include bots
+ *                    GetPlayers_ExcludeHuman - do not include human clients
+ *                    GetPlayers_MatchTeam - match with team
+ *                    GetPlayers_MatchNameSubstring - match with part of name
+ *                    GetPlayers_CaseInsensitive - match case insensitive
+ *                    GetPlayers_ExcludeHLTV - do not include HLTV proxies
+ *                    GetPlayers_IncludeConnecting - include connecting clients
+ * @param team      String to match against if the "e" or "f" flag is specified
+ *
+ * @return          Number of clients on the server that matches the specified flags
+ */
+stock get_playersnum_ex(GetPlayersFlags:flags = GetPlayers_None, const team[] = "")
+{
+	new PlayersNum;
+	get_players_ex(_, PlayersNum, flags, team);
+	return PlayersNum;
+}
--- a/plugins/include/amxmisc.inc
+++ b/plugins/include/amxmisc.inc
@@ -863,7 +863,7 @@ stock set_task_ex(Float:time, const function[], id = 0, const any:parameter[] = 
  *
  * @noreturn
  */
-stock get_players_ex(players[MAX_PLAYERS], &num, GetPlayersFlags:flags = GetPlayers_None, const team[] = "")
+stock get_players_ex(players[MAX_PLAYERS] = {}, &num, GetPlayersFlags:flags = GetPlayers_None, const team[] = "")
 {
 	new strFlags[10];
 	get_flags(_:flags, strFlags, charsmax(strFlags));
--- a/plugins/include/amxmisc.inc
+++ b/plugins/include/amxmisc.inc
@@ -871,7 +871,7 @@ stock get_players_ex(players[MAX_PLAYERS] = {}, &num, GetPlayersFlags:flags = Ge
 }
 
 /**
- * Returns the number of clients on the server that matches the specified flags.
+ * Returns the number of clients on the server that match the specified flags.
  *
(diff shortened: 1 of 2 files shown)


plugins/include/amxmodx.inc
```diff
@@ -1391,7 +1391,7 @@ native get_playersnum(flag = 0);
  *
  * @noreturn
  */
-native get_players(players[MAX_PLAYERS], &num, const flags[] = "", const team[] = "");
+native get_players(players[MAX_PLAYERS] = 0, &num, const flags[] = "", const team[] = "");
```
WPMGPRoSToTeMa: I think `{}` should be better.

justgo97: noticed plugin size got smaller

justgo97: had to read it a couple of times to understand it, but its good that we got rid of the buffer creation and copying operations but is editing the get_players function in amxmodx.cpp to check if the first param is a valid array will be better then current code ?

rsKliPPy: It's not possible to do that (unless we introduce something like `NULL_ARRAY`) as far as I know. The only solution would be to add a new parameter, but in the end it doesn't matter - the native has to do the filtering anyway, copying to the array and increasing the counter is trivial.

Arkshine:
What about having something dedicated for that? Also, we should probably encourage the use of flags as letters.

Maybe a `get_playersnum_ex` stock using `get_players_ex`? It would be more intuitive to use as a return value. Any thoughts?

justgo97:
somthing like this ?

OciXCrom:
It would probably be better if you revert the argument change that you made to `get_players_ex` since you added `get_playersnum_ex`.

justgo97:
@OciXCrom that will not work and will give an error because get_playersnum_ex depends on it

plugins/include/amxmodx.inc
justgo97: @rsKliPPy i see, well its trivial but in some cases this may be called a lot in small delay tasks for example, i guess another native in amxmodx.cpp is the best option if we want to do the most optimized way

OciXCrom:
You can add the array in the stock itself:

```pawn
stock get_playersnum_ex(GetPlayersFlags:flags = GetPlayers_None, const team[] = "")
{
	new Players[MAX_PLAYERS], PlayersNum;
	get_players_ex(Players, PlayersNum, flags, team);
	return PlayersNum;
}
```

I wouldn't say it's a problem, but having both functions being able to do the exact same thing seems redundant. Having only `get_playersnum_ex` do this is good enough.

justgo97:
IMO creating the array should be avoided and they can do the same thing with or without the argument change

rsKliPPy:
Adding the default parameter will create the array on each call anyway, it's really the same.

justgo97:
Well its just the idea of having more code without using it in any way doesn't look right

WPMGPRoSToTeMa:
It's pretty cheap:
```Pawn
stock fun1(arr[32] = {}, &count) {
	get_players(arr, count);
}

public fun2() {
	new count;
	fun1(_, count);
	return count;
}
```
```Asm
PROC  fun1
    push.c           var_001  ;  ref to Variable  
    push.c           var_000  ;  ref to Variable  
    push.s              0x10
    push.s               0xC
    push.c              0x10
    sysreq.c     get_players
    stack               0x14
    zero.pri  
    retn      
ENDP

PROC  fun2
    push.c               0x0
    pushaddr            -0x4
    const.pri        str_001  ;  ref to String  
    heap                0x80
    movs                 0x0
    push.alt  
    push.c               0x8
    call                fun1
    heap               -0x80
    load.s.pri          -0x4
    stack                0x4
    retn      
ENDP
```

plugins/include/amxmodx.inc
```diff
@@ -1391,7 +1391,7 @@ native get_playersnum(flag = 0);
  *
  * @noreturn
  */
-native get_players(players[MAX_PLAYERS], &num, const flags[] = "", const team[] = "");
+native get_players(players[MAX_PLAYERS] = {}, &num, const flags[] = "", const team[] = "");
```
Arkshine: {} should be removed here.

justgo97: done.

plugins/include/amxmisc.inc
```diff
@@ -863,9 +863,37 @@ stock set_task_ex(Float:time, const function[], id = 0, const any:parameter[] =
...
-stock get_players_ex(players[MAX_PLAYERS], &num, GetPlayersFlags:flags = GetPlayers_None, const team[] = "")
+stock get_players_ex(players[MAX_PLAYERS] = {}, &num, GetPlayersFlags:flags = GetPlayers_None, const team[] = "")
 {
 	new strFlags[10];
 	get_flags(_:flags, strFlags, charsmax(strFlags));
 	get_players(players, num, strFlags, team);
 }
+
+/**
+ * Returns the number of clients on the server that matches the specified flags.
```
OciXCrom: match*

justgo97: Fixed

plugins/include/amxmisc.inc
```diff
@@ -863,9 +863,37 @@ stock set_task_ex(Float:time, const function[], id = 0, const any:parameter[] =
...
+ *                    GetPlayers_ExcludeDead - do not include dead clients
+ *                    GetPlayers_ExcludeAlive - do not include alive clients
+ *                    GetPlayers_ExcludeBots - do not include bots
+ *                    GetPlayers_ExcludeHuman - do not include human clients
+ *                    GetPlayers_MatchTeam - match with team
+ *                    GetPlayers_MatchNameSubstring - match with part of name
+ *                    GetPlayers_CaseInsensitive - match case insensitive
+ *                    GetPlayers_ExcludeHLTV - do not include HLTV proxies
+ *                    GetPlayers_IncludeConnecting - include connecting clients
+ * @param team      String to match against if the "e" or "f" flag is specified
```
Arkshine: Letters should be replaced here too.

justgo97: Oh yeah, copied them from get_players_ex description, should i change it too ? or keep it to another PR ?

PR #519, opened 2018-08-24
CStrike: Fix a missing RemoveEntityHashValue function address check

Arkshine:
Just a silly typo.


WPMGPRoSToTeMa:
Related to #417 since the typo was introduced there.

PR #523, opened 2018-08-26
Add message natives

OciXCrom:
The goal of this PR is to make the usage of `message_begin` functions a lot easier, especially the `SVC_TEMPENTITY` ones.

Instead of having to do all of this just to create a single beam:

```pawn
new iStartPos[3], iEndPos[3];
get_user_origin(id, iStartPos);
get_user_origin(id, iEndPos, 3);

message_begin(MSG_ALL, SVC_TEMPENTITY);
write_byte(TE_BEAMPOINTS);
write_coord(iStartPos[0]);
write_coord(iStartPos[1]);
write_coord(iStartPos[2]);
write_coord(iEndPos[0]);
write_coord(iEndPos[1]);
write_coord(iEndPos[2]);
write_short(g_iSprite);
write_byte(0);
write_byte(30);
write_byte(10);
write_byte(10);
write_byte(0);
write_byte(0);
write_byte(0);
write_byte(255);
write_byte(75);
write_byte(0);
message_end();
```

With using the new natives, we can do it as simple as this:

```pawn
new iStartPos[3], iEndPos[3];
get_user_origin(id, iStartPos);
get_user_origin(id, iEndPos, 3);

te_create_beam_between_points(iStartPos, iEndPos, g_iSprite)
```

This saves a lot of time that you would spend searching for the `write_` arguments, experimenting with the numbers and having to write all 20 arguments 1 by 1 and having a huge chance of messing some of them up.

Only the mandatory arguments in the natives are required (e.g. starting position, sprite index, etc) and all other ones have a default value and are optional.

All of the natives share 2 same arguments - `receiver` and `bool:reliable` - both optional and the second one is always the last argument in the natives. The first one controls who will see the message and the second one controls whether or not to use the reliable channel. With this being said, if `receiver` is set to `0` and `reliable` is `true`, the native will use `MSG_ALL` as the message destination.

I also spent some time creating a video that previews all of the `te_` functions to make it easier for users to understand how they work and what they look like - https://youtu.be/szW-bSMPuyQ

In addition, I found some mistakes while testing in the `message_const.inc` file that I fixed, e.g. the `TE_DLIGHT` parameter had a non-existant `brightness` argument that will crash the server if used (I found this out the hard way).

For anyone willing to test things out, my fantastically coded plugin that I used for testing this PR can be found here - https://pastebin.com/6z5rhc3d

The list of added natives is as follow:

1. Some basic natives found in `msgnatives.inc`:

- draw_ammo_pickup_icon
- draw_weapon_pickup_icon
- draw_status_icon
- draw_train_controls
- send_geiger_signal
- hide_hud_elements
- fade_user_screen
- shake_user_screen
- set_user_fov

2. Natives for Counter-Strike found in `cstrike.inc`:

- cs_draw_progress_bar
- cs_play_reload_sound
- cs_set_hud_icon
- cs_set_user_shadow

3. Temporary entity natives found in `tempent.inc`:

- te_create_beam_between_points
- te_create_beam_from_entity
- te_create_gunshot
- te_create_explosion
- te_create_tar_explosion
- te_create_smoke
- te_create_tracer
- te_create_beam_between_entities
- te_create_sparks
- te_create_lava_splash
- te_create_teleport_splash
- te_create_colored_explosion
- te_place_decal_from_bsp_file
- te_create_implosion
- te_create_model_trail
- te_display_additive_sprite
- te_create_beam_sprite
- te_create_beam_ring
- te_create_beam_disk
- te_create_beam_cylinder
- te_create_following_beam
- te_display_glow_sprite
- te_create_beam_ring_between_entities
- te_create_tracer_shower
- te_create_dynamic_light
- te_create_entity_light
- te_draw_line
- te_create_box
- te_remove_all_beams_from_entity
- te_create_large_funnel
- te_create_bloodstream
- te_draw_blood_line
- te_spray_blood
- te_place_brush_decal
- te_create_bouncing_model
- te_create_explode_model
- te_create_break_model
- te_place_gunshot_decal
- te_create_sprite_spray
- te_create_armor_ricochet
- te_place_player_spray
- te_create_bubble_box
- te_create_bubble_line
- te_display_falling_sprite
- te_place_world_decal
- te_create_projectile
- te_create_sprite_shower
- te_emit_sprite_from_player
- te_create_particle_burst
- te_create_fire_field
- te_attach_model_to_player
- te_remove_all_player_attachments
- te_create_multi_gunshot
- te_create_user_tracer

I skipped `TE_LIGHTNING` and `TE_FIZZ` because I couldn't find a working example.
I skipped `TE_TEXTMESSAGE` as well because of obvious reasons.


plugins/include/cstrike.inc
```diff
@@ -1231,6 +1231,84 @@ stock bool:cs_is_valid_itemid(id, bool:weapon_only = false)
...
+ * @param duration      How long (in seconds) until the bar is fully filled
+ * @param startpercent  Bar starting percentage (0-100)
+ * @param reliable      If true, the message will be sent via the reliable
+ *                      channel, otherwise it will use the unreliable one
+ *
+ * @noreturn
+ * @error               If "id" is non-zero and the client isn't
+ *                      connected, an error will be thrown.
+ */
+native cs_draw_progress_bar(id, duration, startpercent = 0, bool:reliable = true)
```
WPMGPRoSToTeMa: Missing semicolon.

plugins/include/cstrike.inc
```diff
@@ -1231,6 +1231,84 @@ stock bool:cs_is_valid_itemid(id, bool:weapon_only = false)
...
+ *                      otherwise it will play "weapons/generic_reload.wav".
+ * @param volume        Volume amount (0 - 255)
+ * @param reliable      If true, the message will be sent via the reliable
+ *                      channel, otherwise it will use the unreliable one
+ *
+ * @noreturn
+ * @error               If "id" is non-zero and the client isn't
+ *                      connected, an error will be thrown.
+ */
+native cs_play_reload_sound(id, bool:shotgun = false, volume = 100, bool:reliable = true)
```
WPMGPRoSToTeMa: Same and same below.

plugins/include/message_const.inc
```diff
@@ -117,6 +117,55 @@ enum
...
 
+/**
+ * Flags used in draw_status_icon()
+ */
+enum StatusIconFlags
+{
+	StatusIcon_Hide = 0,
+	StatusIcon_Show,
+	StatusIcon_Flash
+}
```
WPMGPRoSToTeMa: Possibly missing semicolon, but compiler doesn't pay attention for it. Same below.

plugins/include/message_const.inc
```diff
@@ -799,6 +861,19 @@ enum
  */
 #define TE_BREAKMODEL               108
 
+/**
+ * Flags used in te_create_break_model()
+ */
+#define BreakModel_TypeMask                  0x4F
```
WPMGPRoSToTeMa: Caps?

plugins/include/msgnatives.inc
```diff
@@ -0,0 +1,206 @@
...
+ * @param a             Color alpha (brightness) (0 - 255)
+ * @param reliable      If true, the message will be sent via the reliable
+ *                      channel, otherwise it will use the unreliable one
+ *
+ * @noreturn
+ * @error               If "id" is non-zero and the client isn't
+ *                      connected, an error will be thrown.
+ */
+
+native fade_user_screen(id, Float:duration = 1.0, Float:fadetime = 0.5, ScreenFadeFlags:flags = ScreenFade_FadeIn, r = 0, g = 0, b = 255, a = 75, bool:reliable = true)
```
WPMGPRoSToTeMa: Why default `b` is 255?

plugins/include/msgnatives.inc
```diff
@@ -0,0 +1,206 @@
...
+ * @param duration      Shake duration (in seconds) (0 - 16)
+ * @param frequency     Delay between each shake (0 - 16)
+ * @param reliable      If true, the message will be sent via the reliable
+ *                      channel, otherwise it will use the unreliable one
+ *
+ * @noreturn
+ * @error               If "id" is non-zero and the client isn't
+ *                      connected, an error will be thrown.
+ */
+native shake_user_screen(id, Float:amplitude = 3.0, Float:duration = 3.0, Float:frequency = 1.0, bool:reliable = true)
```
WPMGPRoSToTeMa: What is the reason for these default values?

plugins/include/msgnatives.inc
```diff
@@ -0,0 +1,206 @@
...
+ * @param id            Client index or 0 for all clients
+ * @param fov           Field of view degrees (0 - 255)
+ * @param reliable      If true, the message will be sent via the reliable
+ *                      channel, otherwise it will use the unreliable one
+ *
+ * @noreturn
+ * @error               If "id" is non-zero and the client isn't
+ *                      connected, an error will be thrown.
+ */
+native set_user_fov(id, fov = 0, bool:reliable = true)
```
WPMGPRoSToTeMa: Why default `fov` is `0`?

plugins/include/tempent.inc
```diff
@@ -0,0 +1,1289 @@
...
+ * @param receiver      Client index that will be able to see the beam
+ *                      or 0 for all clients
+ * @param reliable      If true, the message will be sent via the reliable
+ *                      channel, otherwise it will use the unreliable one
+ *
+ * @noreturn
+ * @error               If "receiver" is non-zero and the client isn't
+ *                      connected, an error will be thrown.
+ */
+native te_create_beam_between_points(startpos[3], endpos[3], sprite, startframe = 0, framerate = 30, life = 10, width = 10, noise = 0, r = 0, g = 0, b = 255, a = 75, speed = 0, receiver = 0, bool:reliable = true)
```
WPMGPRoSToTeMa: I can't understand your preferences for default values, same below.

plugins/include/msgnatives.inc
OciXCrom: @WPMGPRoSToTeMa Why not? It's better than having no color at all.

plugins/include/msgnatives.inc
OciXCrom: @WPMGPRoSToTeMa It produces a simple and short shake, setting them to 1.0 is barely noticeable.

plugins/include/msgnatives.inc
OciXCrom: @WPMGPRoSToTeMa Because it resets the fov back to normal? Default values should do that in cases like this.

plugins/include/tempent.inc
OciXCrom: @WPMGPRoSToTeMa I don't see what the problem is. I set the duration to 1 second in all natives, chose a blue color for most of them and kept the other parameters similar in al the natives (e.g. width is 10 everywhere). As you can see from the video, the beams have a perfect set of parameters for displaying the default sprite as most people would use it. I can't just put 0 in all of them, there's no point of having them optional if I do that.

plugins/include/cstrike.inc
OciXCrom: @WPMGPRoSToTeMa I'll run a "pragma semicolon 1" test tomorow and add them everywhere I forgot to.

plugins/include/message_const.inc
OciXCrom: @WPMGPRoSToTeMa IMO it looks better this way, I'll change it to const or enum though.

plugins/include/tempent.inc
WPMGPRoSToTeMa: @OciXCrom I think better just add a note with an example of using it.

plugins/include/msgnatives.inc
WPMGPRoSToTeMa: @OciXCrom then maybe we need something like `reset_user_fov`?
And also why `user`, but not `player` or `client`? Function description says: "Changes the **_client_**'s field...".
There is also a mention of sniper scope, but that's not true for Half-Life.

plugins/include/msgnatives.inc
WPMGPRoSToTeMa: @OciXCrom I think it is better to remove default values and add an example to description.

plugins/include/message_const.inc
WPMGPRoSToTeMa: @OciXCrom tagged enum should fit good here, tagged constants should be good too.

amxmodx/msgnatives.cpp
```diff
@@ -0,0 +1,208 @@
...
+static cell AMX_NATIVE_CALL draw_ammo_pickup_icon(AMX *amx, cell *params)
+{
+	enum args { arg_numargs, arg_id, arg_ammoid, arg_amount, arg_reliable };
+
+	int index = params[arg_id];
+
+	if(!check_msg_receiver(amx, index))
+		return 0;
+
+	MESSAGE_BEGIN(get_msg_destination(index, params[arg_reliable] != 0), gmsgAmmoPickup, NULL, index ? TypeConversion.id_to_edict(index) : NULL);
```
WPMGPRoSToTeMa: `nullptr` maybe? Same below.

modules/cstrike/cstrike/CstrikeNatives.cpp
```diff
@@ -2011,7 +2011,9 @@ static cell AMX_NATIVE_CALL cs_draw_progress_bar(AMX *amx, cell *params)
 	int index = params[1];
 
 	if(index)
+	{
 		CHECK_PLAYER(index);
+	}
```
WPMGPRoSToTeMa: That's because `CHECK_PLAYER` macro isn't correctly written. Look at [RETURN_META macro](https://github.com/theAsmodai/metamod-r/blob/0fff97941f46ccb113d4525e36ef9d814b3ccba7/metamod/extra/example/include/metamod/meta_api.h#L59-L60), it doesn't have this problem.

amxmodx/tempent.cpp
```diff
@@ -0,0 +1,1504 @@
...
+static cell AMX_NATIVE_CALL te_create_user_tracer(AMX *amx, cell *params)
+{
+	enum args { arg_numargs, arg_position, arg_velocity, arg_life, arg_color, arg_length, arg_receiver, arg_reliable };
+
+	int index = params[arg_receiver];
+
+	if(!check_msg_receiver(amx, index))
+		return 0;
+
+	// Values outside the range 0-12 will throw an error
```
WPMGPRoSToTeMa: So... why it isn't thrown here?

plugins/include/tempent.inc
OciXCrom: @WPMGPRoSToTeMa No, one of the goals of these natives is to prevent the usage of all ~15 arguments. Most of the times you won't need all of them, so having default values that have the most common settings is a huge help. Take `set_hudmessage` for example - it has totally random default values.

amxmodx/msgnatives.cpp
OciXCrom: @WPMGPRoSToTeMa Is there an actual difference? I saw that `NULL` is used the most in `amxmodx.cpp` so I decided to go with that one.

modules/cstrike/cstrike/CstrikeNatives.cpp
OciXCrom: @WPMGPRoSToTeMa Yeah, I didn't have this problem when I compiled it on my computer as well.

amxmodx/tempent.cpp
OciXCrom: @WPMGPRoSToTeMa I don't think it has to. Since all other natives accept values from 0 to 255, this one being the only one limited from 0 to 12 will probably get forgotten by users, so better use `clamp` to limit it automatically.

plugins/include/cstrike.inc
OciXCrom: @WPMGPRoSToTeMa It's done now.

plugins/include/message_const.inc
OciXCrom: @WPMGPRoSToTeMa I changed it to `enum`, but I'm not sure if I can use (<< 1) in the tag for HEX values. I assume you know?

amxmodx/msgnatives.cpp
Arkshine: When we can, we modernize the code.

`nullptr` is a new keyword introduced in C++11. `nullptr`  is meant as a replacement to `NULL` . `nullptr` provides a type-safe pointer value representing an empty (null) pointer.

For new code like that, you should always use `nullptr`.

plugins/include/msgnatives.inc
OciXCrom: @WPMGPRoSToTeMa Take a look at the natives in the `fun` module - most of them have the reset value as default. For example, `set_user_gravity` has its `gravity` param set to `1.0` by default, so people can simply write `set_user_gravity(id)` in order to reset it. There's no need for people to research what the reset value for each native is before using it - let it be in the default value so they will know right away.

OciXCrom: @WPMGPRoSToTeMa I chose `user` for the same reason - respecting the `fun` module. It too has `user` in the native names and `client` in the description..

amxmodx/msgnatives.cpp
OciXCrom: @Arkshine Okay, I changed it.

plugins/include/msgnatives.inc
Arkshine: I don't the others mod, but in CS, the default value is 90.

OciXCrom: @Arkshine 0 seems to have the same effect as 90.

plugins/include/message_const.inc
WPMGPRoSToTeMa: @OciXCrom hex doesn't make difference, of course you can use `(<< 1)`.

plugins/include/tempent.inc
```diff
@@ -0,0 +1,1289 @@
...
+ * @param receiver      Client index that will be able to see the effect
+ *                      or 0 for all clients
+ * @param reliable      If true, the message will be sent via the reliable
+ *                      channel, otherwise it will use the unreliable one
+ *
+ * @noreturn
+ * @error               If "receiver" is non-zero and the client isn't
+ *                      connected, an error will be thrown.
+ */
+native te_create_user_tracer(position[3], velocity[3], life = 1, color = 106, length = 1, receiver = 0, bool:reliable = true)
```
WPMGPRoSToTeMa: `color = 106`???

amxmodx/tempent.cpp
WPMGPRoSToTeMa: @OciXCrom there is another way - we can allow to put RGB and then choose the nearest color to it.

OciXCrom: @WPMGPRoSToTeMa That would be cool if it would work correctly. Can you show an example?

plugins/include/message_const.inc
OciXCrom: @WPMGPRoSToTeMa What about `0x4F`? It outputs 79.

amxmodx/tempent.cpp
WPMGPRoSToTeMa: @OciXCrom I think it is better to just add a tagged enum for `color` param, it is more explicit and safe.

plugins/include/msgnatives.inc
WPMGPRoSToTeMa: @OciXCrom is it `90` for all the games on GoldSrc? As I can see from the `SetFOV` client handler code the value `0` sets FOV to the default value, I hope this behavior is the same across the games. So it is better to stick with `0`.

plugins/include/message_const.inc
WPMGPRoSToTeMa: @OciXCrom you can define it as:
```Pawn
BreakModel_TypeMask = BreakModel_Glass | BreakModel_Metal | BreakModel_Flesh | BreakModel_Wood | BreakModel_Concrete
```
or:
```Pawn
const BreakModelFlags:BreakModel_TypeMask = BreakModel_Glass | BreakModel_Metal | BreakModel_Flesh | BreakModel_Wood | BreakModel_Concrete;
```

amxmodx/AMBuilder
```diff
@@ -100,6 +100,8 @@ binary.sources = [
   'CGameConfigs.cpp',
   'gameconfigs.cpp',
   'CoreConfig.cpp',
+  'msgnatives.cpp',
```
WPMGPRoSToTeMa: test

amxmodx/tempent.cpp
OciXCrom: @WPMGPRoSToTeMa Good idea, I added them. However I don't see any visible difference between 6, 8, 9, 10 and 11, so I named all of them `TracerColor_Pink%num%`.

plugins/include/msgnatives.inc
OciXCrom: @WPMGPRoSToTeMa Okay, I switched it back to 0 again, if that's what the code says. There's no visible difference in CS between 0 and 90, so it shouldn't be a problem.

plugins/include/message_const.inc
OciXCrom: @WPMGPRoSToTeMa Thanks, I added the first one in the end of the `enum`.

rsKliPPy:
Regarding naming:
Only enumerations that actually represent flags should be named `<something>Flags`. The naming usually implies that you can do bitwise operations on them.
Enumerations that represent "types" or similar that you can choose one of should be singular. `TracerColors` > `TraceColor`. You can see that in `cstrike_const.inc` (with an ancient exception of `CsTeams`, which is awful).

Suggestions:
- `StatusIconFlags` > `StatusIconDisplay`
- `TrainControlFlags` > `TrainControlSpeed`
- `BreakModelFlags` > `BreakModelType`
- `TracerColors` > `TracerColor`

plugins/include/message_const.inc
```diff
@@ -672,6 +720,12 @@ enum
  */
 #define TE_LARGEFUNNEL              100
 
+/**
+ * Flags used in te_create_large_funnel()
```
rsKliPPy: Why are these not in an enumeration as well?

OciXCrom: @rsKliPPy Not sure to be honnest. I changed them now.

OciXCrom:
@rsKliPPy You're right. I updated the `enum` names.

Arkshine:
If you plan to make a lot of changes in a PR, you should
* Discuss beforehand
* Split the PR in smaller PRs if possible.
* Avoid adding all your new changes in one commit. Make smaller and logical commits.
* Fix separately any original bugs found in another PR.

Random thoughts to be **discussed**:
* About `te_` natives. This doesn't feel right to have them as natives. The only reason you would want to keep them is to error on invalid player's index.  Looking at `message_begin`, even though the engine will skip if the player is not connected, it will crash if the index is incorrect. For some reasons, this native doesn't error on invalid player.  If this native is fixed, I would likely put all those `te_` natives into stocks instead. 
* Same for the others natives, this kind of shortcut to just wrap things should be likely stocks.
* Should we use emessage?
* About `set_user_fov`. This is not the proper way to do that. The message is associated to `m_iFOV`, `m_iClientFOV` and `pev->fov`. Those should be set as well. Also, you have to care if a mod do something different. For example, CS sends also a HLTV message <sup id="a1">[[1]](#f1)</sup>. A CS version might be needed, I don't know. Also, since using class members, it might be a good idea to avoid to put such stuff into the core.
* About `hide_hud_elements`, same as above. It's associated to `m_iHideHUD` and `m_iHideClientHUD`. ReGameDLL added some fix as well <sup id="a2">[[2]](#f2)</sup> Same remark as above. 
* About `shake_user_screen`, it could be more interesting to base the code on `UTIL_ScreenShake` <sup id="a3">[[3]](#f3)</sup>, maybe.
* About `send_geiger_signal`, I don't know if it's really useful to send the message alone. It's also associated to `m_flgeigerDelay`, `m_flgeigerRange`, and `m_igeigerRangePrev`. And to trigger the message, you just need to set a value `m_igeigerRange` != 1000 <sup id="a4">[[4]](#f4)</sup>. This is the same behavior for fov and hidehud btw where the game already checks if a message needs to be sent. Though the geiger message is just about sending a short sound (from 30ms to 300ms max), so it would be kind of silly to redo the logic to send several message when it's already available. Don't know.
* About `cs_set_hud_icon`: Are you saying you can provide an interval without providing the associated icon? If not active, only the first byte should be sent.
* About `cs_reload_sound`, maybe more logic could be added like does `CBasePlayer::ReloadSound` <sup id="a5">[[5]](#f5)</sup>.
* About `cs_draw_progress_bar`, you have the `m_progressStart` and 'm_progressEnd' members associated. You should also care for spectators. CS has already two functions for that <sup id="a6">[[6]](#f6)</sup>
* About the style, please put a space after `if`, `for`, etc.  Also, put the native header above the function and finally put brackets even if one line for consistency.
__

<sup id="f1">__1__</sup> _CS code example with `SetFOV`_ [↩](#a1) 
<details>
  <summary>Details</summary>

```C
   if (m_iFOV != m_iClientFOV)
   {
	// cache FOV change at end of function, so weapon updates can see that FOV has changed
	pev->fov = m_iFOV;

	MESSAGE_BEGIN(MSG_ONE, gmsgSetFOV, nullptr, pev);
		WRITE_BYTE(m_iFOV);
	MESSAGE_END();

	MESSAGE_BEGIN(MSG_SPEC, gmsgHLTV);
		WRITE_BYTE(ENTINDEX(edict()));
		WRITE_BYTE(m_iFOV);
	MESSAGE_END();
   }
```
</details>

__

<sup id="f2">__2__</sup> _CS code example with `HideWeapon`_ [↩](#a2) 
<details>
  <summary>Details</summary>

```C
	if (m_iHideHUD != m_iClientHideHUD)
	{
		MESSAGE_BEGIN(MSG_ONE, gmsgHideWeapon, nullptr, pev);
			WRITE_BYTE(m_iHideHUD);
		MESSAGE_END();

#ifdef REGAMEDLL_FIXES
		if (m_iHideHUD && !(m_iHideHUD & HIDEHUD_OBSERVER_CROSSHAIR))
		{
			if (m_iClientHideHUD < 0)
				m_iClientHideHUD = 0;

			int hudChanged = m_iClientHideHUD ^ m_iHideHUD;
			if (hudChanged & (HIDEHUD_FLASHLIGHT | HIDEHUD_HEALTH | HIDEHUD_TIMER | HIDEHUD_MONEY))
			{
				MESSAGE_BEGIN(MSG_ONE, gmsgCrosshair, nullptr, pev);
					WRITE_BYTE(0);
				MESSAGE_END();
			}
		}
#endif

		m_iClientHideHUD = m_iHideHUD;
	}
```
</details>

__

<sup id="f3">__3__</sup> _UTIL_ScreenShake`_ [↩](#a3) 
<details>
  <summary>Details</summary>

```C
void UTIL_ScreenShake(const Vector &center, float amplitude, float frequency, float duration, float radius)
{
	int i;
	float localAmplitude;
	ScreenShake shake;

	shake.duration = FixedUnsigned16(duration, (1<<12));
	shake.frequency = FixedUnsigned16(frequency, (1<<8));

	for (i = 1; i <= gpGlobals->maxClients; i++)
	{
		CBaseEntity *pPlayer = UTIL_PlayerByIndex(i);
		if (!pPlayer || !(pPlayer->pev->flags & FL_ONGROUND))
			continue;

		localAmplitude = 0;
		if (radius > 0)
		{
			Vector delta = center - pPlayer->pev->origin;
			float distance = delta.Length();

			if (distance < radius)
				localAmplitude = amplitude;
		}
		else
			localAmplitude = amplitude;

		if (localAmplitude)
		{
			shake.amplitude = FixedUnsigned16(localAmplitude, 1<<12);

			MESSAGE_BEGIN(MSG_ONE, gmsgShake, nullptr, pPlayer->edict());
				WRITE_SHORT(shake.amplitude);
				WRITE_SHORT(shake.duration);
				WRITE_SHORT(shake.frequency);
			MESSAGE_END();
		}
	}
}
```
</details>

__

<sup id="f4">__4__</sup> _UpdateGeigerCounter`_ [↩](#a4) 
<details>
  <summary>Details</summary>

```C
#define GEIGERDELAY 0.25

void CBasePlayer :: UpdateGeigerCounter( void )
{
	BYTE range;

	// delay per update ie: don't flood net with these msgs
	if (gpGlobals->time < m_flgeigerDelay)
		return;

	m_flgeigerDelay = gpGlobals->time + GEIGERDELAY;
		
	// send range to radition source to client

	range = (BYTE) (m_flgeigerRange / 4);

	if (range != m_igeigerRangePrev)
	{
		m_igeigerRangePrev = range;

		MESSAGE_BEGIN( MSG_ONE, gmsgGeigerRange, NULL, pev );
			WRITE_BYTE( range );
		MESSAGE_END();
	}

	// reset counter and semaphore
	if (!RANDOM_LONG(0,3))
		m_flgeigerRange = 1000;

}
```
</details>

__

<sup id="f5">__5__</sup> _ReloadSound`_ [↩](#a5) 
<details>
  <summary>Details</summary>

```C
void CBasePlayerWeapon::ReloadSound()
{
	CBasePlayer *pPlayer = nullptr;
	while ((pPlayer = UTIL_FindEntityByClassname(pPlayer, "player")))
	{
		if (pPlayer->IsDormant())
			break;

		if (pPlayer == m_pPlayer)
			continue;

		float distance = (m_pPlayer->pev->origin - pPlayer->pev->origin).Length();
		if (distance <= MAX_DIST_RELOAD_SOUND)
		{
			MESSAGE_BEGIN(MSG_ONE, gmsgReloadSound, nullptr, pPlayer->pev);
				WRITE_BYTE(int((1.0f - (distance / MAX_DIST_RELOAD_SOUND)) * 255.0f));
			if (!Q_strcmp(STRING(pev->classname), "weapon_m3") || !Q_strcmp(STRING(pev->classname), "weapon_xm1014"))
				WRITE_BYTE(0);
			else
				WRITE_BYTE(1);
			MESSAGE_END();
		}
	}
}
```
</details>

__

<sup id="f6">__6__</sup> _SetProgressBarTime` and `SetProgressBarTime2`_ [↩](#a6) 
<details>
  <summary>Details</summary>

```C
void CBasePlayer::SetProgressBarTime(int time)
{
	if (time)
	{
		m_progressStart = gpGlobals->time;
		m_progressEnd = time + gpGlobals->time;
	}
	else
	{
		m_progressStart = 0;
		m_progressEnd = 0;
	}

	MESSAGE_BEGIN(MSG_ONE, gmsgBarTime, nullptr, pev);
		WRITE_SHORT(time);
	MESSAGE_END();

	int playerIndex = entindex();
	CBaseEntity *pEntity = nullptr;

	while ((pEntity = UTIL_FindEntityByClassname(pEntity, "player")))
	{
		if (FNullEnt(pEntity->edict()))
			break;

		CBasePlayer *pPlayer = GetClassPtr<CCSPlayer>((CBasePlayer *)pEntity->pev);

		if (pPlayer->GetObserverMode() == OBS_IN_EYE && pPlayer->pev->iuser2 == playerIndex)
		{
			MESSAGE_BEGIN(MSG_ONE, gmsgBarTime, nullptr, pPlayer->pev);
				WRITE_SHORT(time);
			MESSAGE_END();
		}
	}
}

void CBasePlayer::SetProgressBarTime2(int time, float timeElapsed)
{
	if (time)
	{
		m_progressStart = gpGlobals->time - timeElapsed;
		m_progressEnd = time + gpGlobals->time - timeElapsed;
	}
	else
	{
		timeElapsed = 0;
		m_progressStart = 0;
		m_progressEnd = 0;
	}

	short iTimeElapsed = (timeElapsed * 100.0 / (m_progressEnd - m_progressStart));

	MESSAGE_BEGIN(MSG_ONE, gmsgBarTime2, nullptr, pev);
		WRITE_SHORT(time);
		WRITE_SHORT(iTimeElapsed);
	MESSAGE_END();

	int playerIndex = entindex();
	CBaseEntity *pEntity = nullptr;

	while ((pEntity = UTIL_FindEntityByClassname(pEntity, "player")))
	{
		if (FNullEnt(pEntity->edict()))
			break;

		CBasePlayer *pPlayer = GetClassPtr<CCSPlayer>((CBasePlayer *)pEntity->pev);

		if (pPlayer->GetObserverMode() == OBS_IN_EYE && pPlayer->pev->iuser2 == playerIndex)
		{
			MESSAGE_BEGIN(MSG_ONE, gmsgBarTime2, nullptr, pPlayer->pev);
				WRITE_SHORT(time);
				WRITE_SHORT(iTimeElapsed);
			MESSAGE_END();
		}
	}
}
```
</details>

plugins/include/message_const.inc
```diff
@@ -1151,6 +1151,25 @@ enum BreakModelFlags
...
+/**
+ * Colors used in te_create_user_tracer()
+ */
+enum TracerColors
+{
+	TracerColor_White = 0,
+	TracerColor_Red,
+	TracerColor_Green,
+	TracerColor_Blue,
+	TracerColor_Yellow,
```
WPMGPRoSToTeMa: Isn't it black?

amxmodx/tempent.cpp
WPMGPRoSToTeMa: @OciXCrom here the list of colors:
```
0 - FFFFFF
1 - FF0000
2 - 00FF00
3 - 0000FF
4 - 000000
5 - FFA711
6 - FF825A
7 - 373C90
8 - FF825A
9 - FF8C5A
10 - C8825A
11 - FF7846
12 - 000000
```
You can convert them to color names using [this](https://www.color-blindness.com/color-name-hue/) (for example).

WPMGPRoSToTeMa:
@Arkshine anchors don't work.

amxmodx/tempent.cpp
OciXCrom: @WPMGPRoSToTeMa Are you sure this list is correct? The 12th color definitely isn't #000000.

plugins/include/message_const.inc
OciXCrom: @WPMGPRoSToTeMa Nope, unless it's uncolored - https://i.imgur.com/1rqQ1r7.jpg
But the 12th color definitely isn't black as well - https://i.imgur.com/3rbtdlx.jpg

OciXCrom:
@Arkshine I first started making these as stocks but then I decided to remake them as natives. If I'm not mistaken, natives are called faster than stocks, right? If you want to have them as stocks instead, I have the stock version uploaded [url=https://forums.alliedmods.net/showthread.php?t=310276]here[/url], but I think we should stick to natives instead. I'll look at the other things you suggested tomorrow and I'll try to make them. I'm not that great at C++ but I'll see what I can do. I though `MESSAGE_BEGIN` will change all other things associated. For `emessage`, we can also include another optional parameter at the end of every native (e.g. `bool:hookable`).

Arkshine:
I'm not asking, we need to discuss it before. I don't have the universal answer to everything. The PR has too many changes, we need to figure out first what to keep, what to trash, what to change, etc.

rsKliPPy:
The performance should be the least of your concern when deciding `stock` vs `native`, especially when it's something as insignificant as this which you won't (and can't) execute too much of. The primary concern is whether you want it to be compiled into plugins (`stock`) or have the implementation separate so it can be changed at any time without compilation (`native`).
I honestly think these deserve to be `stock` functions as well.

plugins/include/message_const.inc
WPMGPRoSToTeMa: @OciXCrom yeah, 12th color is buggy and invalid, it'll be `000080` on Windows client and `000000` on Linux client (what about Mac? lol). 4th color is based on `tracerred`, `tracergreen`, `tracerblue` and `traceralpha` cvars.

OciXCrom:
@Arkshine @rsKliPPy Okay then. I can convert them to stocks if you believe there won't be any performance  improvements if they are natives. I'll wait for the discussion before I do that.

amxmodx/tempent.cpp
WPMGPRoSToTeMa: @OciXCrom answered here https://github.com/alliedmodders/amxmodx/pull/523#discussion_r214147046.

plugins/include/message_const.inc
OciXCrom: @WPMGPRoSToTeMa Thanks for this information, I updated the documentation.

plugins/include/message_const.inc
```diff
@@ -1164,15 +1164,15 @@ enum TracerColor
...
-	TracerColor_Pink2,
-	TracerColor_Pink3,
-	TracerColor_Pink4,
-	TracerColor_Pink5,
-	TracerColor_DarkBlue
+	TracerColor_Coral2,
+	TracerColor_Coral3,
+	TracerColor_Coral4,
+	TracerColor_Coral5,
+	TracerColor_Invalid
```
WPMGPRoSToTeMa: I think we can just remove it.

plugins/include/message_const.inc
```diff
@@ -1164,15 +1164,15 @@ enum TracerColor
...
 	TracerColor_Orange,
-	TracerColor_Pink,
+	TracerColor_Coral,
 	TracerColor_Purple,
-	TracerColor_Pink2,
-	TracerColor_Pink3,
-	TracerColor_Pink4,
-	TracerColor_Pink5,
-	TracerColor_DarkBlue
+	TracerColor_Coral2,
```
WPMGPRoSToTeMa: I think we can remove it, because it is just a duplicate of 6 (`TracerColor_Coral`).

plugins/include/message_const.inc
```diff
@@ -1164,15 +1164,15 @@ enum TracerColor
 	TracerColor_Red,
 	TracerColor_Green,
 	TracerColor_Blue,
-	TracerColor_Yellow,
+	TracerColor_User,
```
WPMGPRoSToTeMa: I think we can add a comment with its default color.

plugins/include/message_const.inc
```diff
@@ -1164,15 +1164,15 @@ enum TracerColor
...
 	TracerColor_Purple,
-	TracerColor_Pink2,
-	TracerColor_Pink3,
-	TracerColor_Pink4,
-	TracerColor_Pink5,
-	TracerColor_DarkBlue
+	TracerColor_Coral2,
+	TracerColor_Coral3,
+	TracerColor_Coral4,
+	TracerColor_Coral5,
```
WPMGPRoSToTeMa: I don't know how to properly name them, color databases that I found can't distinguish them.

ClaudiuHKS:
Hard work man, how much time did it take?
This will definitely reduce the size of the future plugins.. and increase the execution speed..

OciXCrom:
@ClaudiuHKS Took about 10 days. The weather was bad on my vacation so I had nothing smarter to do. I don't about the improved size and speed if we decide to make them as `stocks` instead.

wopox1337:
bump

seypaa:
Still waiting

wopox1337:
Why still not merged? @Arkshine

PR #531, opened 2018-08-29
Add register_chatcmd

OciXCrom:
Just a simple stock to make registering chat commands easier.

Instead of doing this:

```pawn
register_clcmd("say /command", "function", ADMIN_SLAY, "description")
register_clcmd("say_team /command", "function", ADMIN_SLAY, "description")
```

We can just do this:

```pawn
register_chatcmd("/command", "function", ADMIN_SLAY, "description")
```

It comes in handy when plugins use multiple chat commands.

### DIFF
--- a/plugins/include/amxmisc.inc
+++ b/plugins/include/amxmisc.inc
@@ -863,11 +863,39 @@ stock set_task_ex(Float:time, const function[], id = 0, const any:parameter[] = 
  *
  * @noreturn
  */
-stock get_players_ex(players[MAX_PLAYERS], &num, GetPlayersFlags:flags = GetPlayers_None, const team[] = "")
+stock get_players_ex(players[MAX_PLAYERS] = {}, &num, GetPlayersFlags:flags = GetPlayers_None, const team[] = "")
 {
 	new strFlags[10];
 	get_flags(_:flags, strFlags, charsmax(strFlags));
 	get_players(players, num, strFlags, team);
+}
+
+/**
+ * Returns the number of clients on the server that match the specified flags.
+ *
+ * @note Example retrieving all alive CTs:
+ *       new AliveCt = get_playersnum_ex(GetPlayers_ExcludeDead | GetPlayers_MatchTeam, "CT")
+ *
+ * @param flags     Optional filtering flags (enum GetPlayersFlags); valid flags are:
+ *                    GetPlayers_None - No filter (Default)
+ *                    GetPlayers_ExcludeDead - do not include dead clients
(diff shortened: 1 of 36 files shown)


plugins/include/amxmisc.inc
```diff
@@ -869,3 +869,32 @@ stock get_players_ex(players[MAX_PLAYERS], &num, GetPlayersFlags:flags = GetPlay
...
+ *
+ * @param client_cmd    Command to register
+ * @param function      Callback function
+ * @param flags         Admin privilege flags required
+ * @param info          Command description
+ * @param FlagManager   0 opts out of flag manager, 1 opts in, -1 selects
+ *                      automatically
+ * @param info_ml       If true, the parameter "info" will be looked up as multilingual key
+ *
+ * @noreturn
```
OciXCrom: TODO: return command index.

WPMGPRoSToTeMa: @OciXCrom do you think it is possible? There are two commands registered inside.

OciXCrom: @WPMGPRoSToTeMa I'm pretty sure the second index will always be the first index minus one, so we can just return the index of the first one and state in the documentation how to retreive the second one.

rsKliPPy: That's honestly not intuitive, better return them by ref, optionally. Other than that, you shouldn't ever have to do such arithmetic, it's closely tied to the implementation details - if they change (for example we start using handles instead) it will break.

Also commands register by AMXX have indices in negative - if one command is -10, the next one is -11. So the order is reversed.

OciXCrom: @rsKliPPy You're right. I changed it to store them by ref.

Btw I know about the reversed order, that's why I made it return the "say_team" index and use +1 to get the "say" index.

rsKliPPy: Oh sorry, I was tripping it said "- 1" for some reason.

OciXCrom:
@ClaudiuHKS Chat commands can't be called from the server console, so there's no reason to use `register_concmd`.

The difference is that `clcmds` don't get shown in `amx_help` at all. I don't know why it's made like that in the first place.

ClaudiuHKS:
I didn't say anything about register_srvcmd..

afwn90cj93201nixr2e1re:
Seems like useless.

OciXCrom:
@afwn90cj93201nixr2e1re Please elaborate. Why should we use two *almost* identical lines of code for such a often used thing instead of one?

PR #532, opened 2018-08-30
Make the "body" param in get_user_aiming() optional

OciXCrom:
This param is skipped most of the times. I personally have never had to use it even though I use this function all the time.


OciXCrom:
@Arkshine Done.

PR #533, opened 2018-08-30
Add precache_player_model

OciXCrom:
Since the player model path is always hardcoded (models/player/name/name.mdl) and the functions that set player models demand only the model name instead of the full path, it's nice to have a simple stock to do all the work for us. This stock also checks for a T.mdl file which will prevent a crash if the user didn't add support for these kinds of models.

### DIFF
--- a/plugins/include/amxmisc.inc
+++ b/plugins/include/amxmisc.inc
@@ -869,3 +869,30 @@ stock get_players_ex(players[MAX_PLAYERS], &num, GetPlayersFlags:flags = GetPlay
 	get_flags(_:flags, strFlags, charsmax(strFlags));
 	get_players(players, num, strFlags, team);
 }
+
+/**
+ * Precaches a player model file.
+ *
+ * @note Can only be used inside of the plugin_precache() forward.
+ * @note Also searches for a "T.mdl" file and precaches that one as well if found.
+ * @note Player models MUST be placed in the following directory:
+ *       "models/player/%name%/%name%.mdl" where "name" is the name of the model.
+ *
+ * @param name  Name of the model file without the .mdl extension (e.g. "admin")
+ * @param id    Variable to store the "T.mdl" cache id if found
+ *
+ * @return      Unique cache id of the model
+ * @error       If called outside of the plugin_precache() forward, an error is
+ *              thrown from the precache_model() native.
+ */
+stock precache_player_model(const name[], &id = 0)
+{
+	new model[128];
+	formatex(model, charsmax(model), "models/player/%s/%sT.mdl", name, name);
+
+	if(file_exists(model))
+		id = precache_model(model);
+
+	replace_string(model, charsmax(model), "T.mdl", ".mdl");
+	return precache_model(model);
+}


WPMGPRoSToTeMa:
Why `T.mdl` is precached here, but not precached in `precache_model`?

Arkshine:
It would be more useful to have native to read any model to figure out if there are external textures (`T.mdl` or `t.mdl`) and groups (`01.mdl`, `02.mdl`, etc.). I believe ReHLDS already does that now.

Arkshine:
It would be also welcomed to upgrade the existing `precache_model` to read any model to figure out if there are external textures (`T.mdl` or `t.mdl`) and groups (`01.mdl`, `02.mdl`, etc.). I believe ReHLDS already does that now. Then you could base this stock on the upgraded `precache_model`.

plugins/include/amxmisc.inc
```diff
@@ -869,3 +869,30 @@ stock get_players_ex(players[MAX_PLAYERS], &num, GetPlayersFlags:flags = GetPlay
...
+ */
+stock precache_player_model(const name[], &id = 0)
+{
+	new model[128];
+	formatex(model, charsmax(model), "models/player/%s/%sT.mdl", name, name);
+
+	if(file_exists(model))
+		id = precache_model(model);
+
+	replace_string(model, charsmax(model), "T.mdl", ".mdl");
```
WPMGPRoSToTeMa: I don't like this way, it works because GoldSrc allows only one dot in file path. Imagine if model path is `models/player/superT.mdl/superT.mdl.mdl`.

plugins/include/amxmisc.inc
```diff
@@ -869,3 +869,30 @@ stock get_players_ex(players[MAX_PLAYERS], &num, GetPlayersFlags:flags = GetPlay
...
+ * @param id    Variable to store the "T.mdl" cache id if found
+ *
+ * @return      Unique cache id of the model
+ * @error       If called outside of the plugin_precache() forward, an error is
+ *              thrown from the precache_model() native.
+ */
+stock precache_player_model(const name[], &id = 0)
+{
+	new model[128];
+	formatex(model, charsmax(model), "models/player/%s/%sT.mdl", name, name);
```
WPMGPRoSToTeMa: Some models have `t.mdl` instead of `T.mdl`, this can be a problem on OS with case-sensitive filesystem (like Linux) https://github.com/dreamstalker/rehlds/commit/6715402425e2247818086a51f9662c0605447177.

plugins/include/amxmisc.inc
```diff
@@ -869,3 +869,30 @@ stock get_players_ex(players[MAX_PLAYERS], &num, GetPlayersFlags:flags = GetPlay
...
+
+	if(file_exists(model))
```
WPMGPRoSToTeMa: This is not the way how it should be checked, you need to check if model really needs it https://github.com/dreamstalker/rehlds/commit/9080b9b7e82fd4bdb84b9f965f940728144b1925#diff-462e3f794ed90c12c86f5d8d5b725c4aR5405.

plugins/include/amxmisc.inc
```diff
@@ -869,3 +869,30 @@ stock get_players_ex(players[MAX_PLAYERS], &num, GetPlayersFlags:flags = GetPlay
...
+		id = precache_model(model);
```
WPMGPRoSToTeMa: It is better to precache it via `precache_generic`, because `precache_generic` doesn't have limit like `precache_model` (on ReHLDS).

plugins/include/amxmisc.inc
OciXCrom: @WPMGPRoSToTeMa I changed it to check the last characters only

plugins/include/amxmisc.inc
OciXCrom: @WPMGPRoSToTeMa I changed it to generic.

plugins/include/amxmisc.inc
OciXCrom: @WPMGPRoSToTeMa Which I assume can't be done with a stock?

OciXCrom:
@Arkshine @WPMGPRoSToTeMa I'll leave it up to you guys to update `precache_model` because I don't think I'll be able to do it. I'll stick to the stock.

plugins/include/amxmisc.inc
```diff
@@ -891,8 +919,8 @@ stock precache_player_model(const name[], &id = 0)
 	formatex(model, charsmax(model), "models/player/%s/%sT.mdl", name, name);
 
 	if(file_exists(model))
-		id = precache_model(model);
+		id = precache_generic(model);
 
-	replace_string(model, charsmax(model), "T.mdl", ".mdl");
+	replace_string(model[strlen(model) - 4], charsmax(model), "T.mdl", ".mdl");
```
WPMGPRoSToTeMa: Still not the best option, magic `4` should be replaced with something like `sizeof("T.mdl") - 1`. But anyway `replace_string` doesn't fit good for me.

Arkshine:
Will do it later if you can't.

plugins/include/amxmisc.inc
OciXCrom: @WPMGPRoSToTeMa I changed it to `charsmax` because `sizeof` would output `6` instead.

What do you suggest using instead `replace_string`? I couldn't figure out a way to manually insert a character without using functions.

plugins/include/amxmisc.inc
```diff
@@ -897,3 +897,31 @@ stock get_playersnum_ex(GetPlayersFlags:flags = GetPlayers_None, const team[] =
...
+ */
+stock precache_player_model(const name[], &id = 0)
+{
+	new model[128];
+	formatex(model, charsmax(model), "models/player/%s/%sT.mdl", name, name);
+
+	if(file_exists(model))
+		id = precache_generic(model);
+
+	static const extension[] = "T.mdl";
```
justgo97: why static ?

OciXCrom: @justgo97 It doesn't need to be defined more than once. It acts like a global variable with a local scope.

OciXCrom:
@Arkshine Sounds good.

plugins/include/amxmisc.inc
justgo97: static is bad here because its lifetime (or "extent") is the entire run of the program while these stock will be only use in one forward.

WPMGPRoSToTeMa:
I think support for model texture should be done not in `precache_model`, but at some other side, like `PF_precache_model_I` hook in order to precache model texture for all models, not only for precached from plugins (e.g. from amxx modules).

Arkshine:
That would be ideal, but not everyone uses ReHLDS and we can't expect any additional features from Valve.

plugins/include/amxmisc.inc
WPMGPRoSToTeMa: `static` is fully OK for local array/string constants. There is no other option:
1. `new` isn't appropriate here because it does copying on each function call.
2. Global constant is bad because we use this constant only in this function, so there is no reason for it to be global.

plugins/include/amxmisc.inc
WPMGPRoSToTeMa: @OciXCrom `copy` maybe.

plugins/include/amxmisc.inc
justgo97: yeah but this string constant is pretty small in size and in its use having it static means it will stay in memory all the time while its only needed at the 1-2 seconds of map start for plugin_precache forward ?

plugins/include/amxmisc.inc
```diff
@@ -921,6 +921,7 @@ stock precache_player_model(const name[], &id = 0)
 	if(file_exists(model))
 		id = precache_generic(model);
 
-	replace_string(model[strlen(model) - 5], charsmax(model), "T.mdl", ".mdl");
+	static const extension[] = "T.mdl";
+	replace_string(model[charsmax(extension)], charsmax(model), extension, ".mdl");
```
WPMGPRoSToTeMa: `charsmax` isn't named good for this case, it would be better to have something like `const_strlen`.

plugins/include/amxmisc.inc
WPMGPRoSToTeMa: @justgo97 it will stay in memory with both `new` and `static`, but `new` will also copy data from "global" constant to local on each call.

justgo97: oh, i thought with new it will be deleted after local scope is finished.

justgo97: so new is better for memory in this case ?

WPMGPRoSToTeMa: @justgo97 no, `new = data + stack`, `static = data`.

WPMGPRoSToTeMa:
@Arkshine we can implement it in amxmodx for non-ReHLDS.

plugins/include/amxmisc.inc
justgo97: @WPMGPRoSToTeMa I mean in overall which is better for memory ?

```
public plugin_init()
{
    func1()

    func2()
}

public func1()
{
    new const Block1[] = "I_M_GOING_TO_STAY_IN_MEMORY_JUST_FOR_THIS_FUNC"
    new const Block2[] = "I_M_GOING_TO_STAY_IN_MEMORY_JUST_FOR_THIS_FUNC"
    new const Block3[] = "I_M_GOING_TO_STAY_IN_MEMORY_JUST_FOR_THIS_FUNC"
    new const Block4[] = "I_M_GOING_TO_STAY_IN_MEMORY_JUST_FOR_THIS_FUNC"
    new const Block5[] = "I_M_GOING_TO_STAY_IN_MEMORY_JUST_FOR_THIS_FUNC"
    new const Block6[] = "I_M_GOING_TO_STAY_IN_MEMORY_JUST_FOR_THIS_FUNC"
}

public func2()
{
    static const Block1[] = "I_M_GOING_TO_STAY_IN_MEMORY_FOR_EVER"
    static const Block2[] = "I_M_GOING_TO_STAY_IN_MEMORY_FOR_EVER"
    static const Block3[] = "I_M_GOING_TO_STAY_IN_MEMORY_FOR_EVER"
    static const Block4[] = "I_M_GOING_TO_STAY_IN_MEMORY_FOR_EVER"
    static const Block5[] = "I_M_GOING_TO_STAY_IN_MEMORY_FOR_EVER"
    static const Block6[] = "I_M_GOING_TO_STAY_IN_MEMORY_FOR_EVER"
}
```

WPMGPRoSToTeMa: @justgo97 if consider only permanent memory, then they are both the same.

justgo97: so for non-permanent memory new is better, which is an advantage for new over static ?

rsKliPPy: It doesn't matter.

PR #536, opened 2018-08-30
Remove GetModelCollisionBox, SetModelCollisionBox and SetModelBoundingBox natives

Arkshine:
`SetModelCollisionBox` and `SetModelBoundingBox` are confusing, does not do what you would expect, and redundant. No one use them. So, time to remove them.

`GetModelCollisionBox` doesn't seem to be useful as such values are rarely set in the model. Not sure why. It looks like only values per sequence are actually used.

As for `GetModelBoundingBox`, not sure. Likely will remove Model_DefaultSize mode as this min/max values seems to be never set. Also, someone suggested a way to get the min/max from of all the sequences, but don't know if useful?

### DIFF
--- a/modules/fakemeta/misc.cpp
+++ b/modules/fakemeta/misc.cpp
@@ -149,42 +149,6 @@ enum
 	Model_CurrentSequence	= -1,
 };
 
-// GetModelCollisionBox( index, Float:mins[3], Float:maxs[3] );
-static cell AMX_NATIVE_CALL GetModelCollisionBox(AMX *amx, cell *params)
-{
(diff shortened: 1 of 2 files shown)


ClaudiuHKS:
Maybe some people use them. Only remove them if you're sure they don't work at all.

Arkshine:
They have been introduced in dev. It's expected changes happen. Here, I know no use it because the size used from the model is almost never set. Only the size per sequence seems to be relevant. Also, Set* natives are really pointless and misleading, it does just SetSize.

PR #542, opened 2018-09-02
Add automatic precaching of model textures, groups and sounds

Arkshine:
An attempt to allow automatic precaching for model textures (`*T.mdl`) / groups (`*01.mdl`, etc.) and weapon sounds, similar to what does already ReHLDS.

Few remarks:
* Small QoL mainly for convenience and to trigger less crash
* This targets only vanilla engine as it exists in ReHLDS. It would have been ideal to have this in the engine, but considering GoldSrc is in maintenance mode only, new features are not going to happen. 
* The reason that the code is global and located in `ServerActivate` is to allow the modules as well.
* The code relies on a temporary fake entity because of afaik, unless done manually, there is no direct way other than using `pfnGetModelPtr` which require an edict (internally, it retrieves `model_t*` from `pev->modelindex`). 
* Unlike ReHLDS, there is no cvar to choose to precache weapon sounds. Not sure if necessary. Why is there a cvar?
* Actually, do we need cvars for people to opt-in automatic precaching?

To be discussed.

### DIFF
--- a/amxmodx/meta_api.cpp
+++ b/amxmodx/meta_api.cpp
@@ -34,6 +34,7 @@
 #include "CoreConfig.h"
 #include <resdk/mod_rehlds_api.h>
 #include <amtl/am-utility.h>
+#include <studio.h>
 
 plugin_info_t Plugin_info =
 {
@@ -156,6 +157,9 @@ int FF_ClientConnectEx = -1;
 
 IFileSystem* g_FileSystem;
 HLTypeConversion TypeConversion;
+
+bool HasReHLDS;
+server_t *Server;
 
 bool ColoredMenus(const char *ModName)
 {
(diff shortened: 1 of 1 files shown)


WPMGPRoSToTeMa:
> The code relies on a temporary fake entity because of afaik, unless done manually, there is no direct way other than using pfnGetModelPtr which require an edict (internally, it retrieves model_t* from pev->modelindex).

Then maybe just create a variable with the `edict_t` type? I think we can rely on implementation details here.

Arkshine:
You're right, not sure why I did not occur to me.

Any others suggestions or thoughts on it? What about the cvars?

PartialCloning:
Missing *T.mdl results in the server crashing, missing sounds don't. Weapon sound precache or precaching sounds attached to models should have a cvar and more importantly should be opt in, not opt out. Like REHLDS.

Expect zombie server owners to complain about the new amxmodx crashing their server because of too many resources, where as the old one worked fine.

Arkshine:
Well, we can go with upgrading `precache_model` native, and adding `precache_weapon_sound` native for example. For modules, adding `MF_PrecacheModel` and `MF_PrecacheSound` or something like that.

WPMGPRoSToTeMa:
> and adding `precache_weapon_sound` native for example.

@Arkshine it's not only for weapons.

Arkshine:
Then I guess `precache_model_sound` :P.

Anyway, please give feedbacks, should we go with the natives way?

justgo97:
Yeah, this should be just natives and shouldn't change any HLDS behavior

Arkshine:
Alright.

WPMGPRoSToTeMa:
Maybe we should go with both cvar and native?

PR #543, opened 2018-09-02
Adjust version for manual build and update VS projects

Arkshine:


modules/fun/fun.h
```diff
@@ -103,7 +103,7 @@ class CPlayer
 
 	private:
 
-		int  bodyHits_[kMaxClients + 1] {};
+		int  bodyHits_[kMaxClients + 1];
```
WPMGPRoSToTeMa: How is it related to this PR?

Arkshine: Not sure. It's possible I was playing with PVS-Studio and forgot about this change.

PR #544, opened 2018-09-02
Add ammo constants for max clip ammo and max backpack ammo

OciXCrom:
To prevent using magic numbers. Most usual case is giving a weapon to a player:

```pawn
give_item(id, "weapon_ak47")
cs_set_user_bpammo(id, CSW_AK47, 30)
```

The number 30 should have a constants, as well as all others. I know there's `cs_get_weapon_info`, but there's no need to use a function here.


OciXCrom:
Aaaand the old commits popped up again. If someone knows how to fix that, let me know, if not, so be it.

rsKliPPy:
The function is the exactly because you shouldn't hardcode that stuff, this is just going backwards.

wopox1337:
Why not give names to enums? This enums can declare array sizes

OciXCrom:
@rsKliPPy In order to use that function, the value of it will need to be cached in a variable in order for it to be optimized.

```pawn
static const some_ammo = cs_get_weapon_info(CSW_AK47, CS_WEAPONINFO_GUN_CLIP_SIZE)
give_item(id, "weapon_ak47")
cs_set_user_bpammo(id, CSW_AK47, some_ammo)
```

Imagine what you have to do when you have multiple weapons - it's very inconvenient. These values never change by default so it's better to have them as constants.

You can get the weapon cost with the same function, but we still have price constants as well.

@wopox1337 It will give a tag mismatch when used in one of the existing functions, e.g. `cs_set_user_bpammo`.

rsKliPPy:
You won't be setting ammo every frame, chill.

ClaudiuHKS:
Since there is a cstrike_const.inc file, I'd agree implementing these constants.

OciXCrom:
@rsKliPPy I'm aware of that, but I'm also aware that plugin approvers will say to precache the value for maximum optimization and "good practice".

HamletEagle:
I don't think you have to cache the result. Using the native directly also accounts for plugins altering WeaponInfoStruct on the fly to change bpammo/price/etc(not very likely to happen, but still).

Arkshine:
^ That's mainly the reason.

That's said, having constants to get the original and default value could make sense but doesn't seem really useful at first glance.

WPMGPRoSToTeMa:
@wopox1337 
> This enums can declare array sizes

What do you mean?

PR #546, opened 2018-09-03
Compiler: Add __FILE__ and remove __BINARY_PATH__ constants

Arkshine:
Basically, done #497 in a different way using the existing code rather than introducing redundant code.

`__FILE__` can be either the main plugin (e.g. `main.sma`) or any file that the main plugin includes with the relative path (what you define after `#include` , e.g. `myinclude.inc`, or `mydir/myinclude.inc`.

`__BINARY_PATH__` is removed because the absolute path is not useful as it can vary, and some people will not like having their personal directories being displayed.

Should I rename `__BINARY_NAME__` to `__BINARY__` ?

### DIFF
--- a/compiler/libpc300/sc.h
+++ b/compiler/libpc300/sc.h
@@ -521,6 +521,10 @@ SC_FUNC symbol *add_constant(char *name,cell val,int vclass,int tag);
 SC_FUNC void exporttag(int tag);
 SC_FUNC void sc_attachdocumentation(symbol *sym);
 SC_FUNC int get_actual_compound(symbol *sym);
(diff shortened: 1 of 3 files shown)


ClaudiuHKS:
Yes **__FILE__** and **__BINARY__** look nice. No need of **_NAME**.

WPMGPRoSToTeMa:
Too late, but maybe `__FILE__` should be `__SOURCE__`? (by analogy with `__BINARY__`)

Arkshine:
An alias sounds good. It's true it's not that obvious.

PR #548, opened 2018-09-04
Add cs_get_weaponbox_item native

HamletEagle:
The purpose is to add a native to retrieve the weapon entity that was packed into a weaponbox. This is quite a common thing to do in plugins(for example when creating custom weapons or blocking touch for specific weapons).
Moreover, we already have a native to retrieve the type of an armoury_entity so it makes sense to have one for weaponbox too.

This is the first time when I make a PR, I hope everything is fine. If I messed something let me know.

### DIFF
--- a/modules/cstrike/cstrike/CstrikeNatives.cpp
+++ b/modules/cstrike/cstrike/CstrikeNatives.cpp
@@ -1998,6 +1998,34 @@ static cell AMX_NATIVE_CALL cs_get_user_weapon(AMX *amx, cell *params)
 	return 0;
 }
 
+// native cs_get_wpnbox_weapon(weaponboxIndex);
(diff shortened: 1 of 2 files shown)


HamletEagle:
cs_get_weaponbox_item sound much better.

Arkshine:
I assume you tested it?

HamletEagle:
Of course I did, I never post anything unless it's tested. If you are referring to the != check, it appeared to work fine when I used with an weaponbox/non-weaponbox entities.

OciXCrom:
@Arkshine The Windows build fails to generate after this change - https://www.amxmodx.org/downloads-new.php . Not sure if it's a problem from the commit or the generator.

Arkshine:
I know, thanks. I'm on it.

PR #549, opened 2018-09-04
Add Vector Zero Funcs. & Improve Vector.Inc File

ClaudiuHKS:
Add **bool: IsVectorZero ( Float: Vector [ 3 ] )** to check whether a vector equals to zero.
Add **SetVectorZero ( Float: Vector [ 3 ] )** to assign a vector to zero.

Remove redundant tabs at some lines.
Fix spelling @ **get_distance_f**.

This PR intends to complete AMXX's Vector library adding new stuff to it.
Also, the new two functions listed above will help plugins using this kind of stuff look more readable.

### DIFF
--- a/plugins/include/vector.inc
+++ b/plugins/include/vector.inc
@@ -28,7 +28,7 @@
  * 
  * @param origin1       The first vector
  * @param origin2       The second vector
- * 	
+ * 
  * @return              The distance between two input vectors
  */
 native get_distance(const origin1[3], const origin2[3]);
@@ -38,10 +38,10 @@ native get_distance(const origin1[3], const origin2[3]);
  * 
  * @param origin1       The first vector
  * @param origin2       The second vector
- * 	
+ * 
  * @return              The distance between two input vectors
  */
-native Float:get_distance_f(const Float:Origin1[3], const Float:Origin2[3]);
+native Float:get_distance_f(const Float:origin1[3], const Float:origin2[3]);
 
 /**
  * Calculates velocity in the direction player is looking.
@@ -49,7 +49,7 @@ native Float:get_distance_f(const Float:Origin1[3], const Float:Origin2[3]);
  * @param iIndex        Client index
  * @param iVelocity     Multiply vRetValue length by this much
  * @param vRetValue     Store the calculated velocity in this vector.
- * 	
+ * 
  * @noreturn
  * @error               If client is not connected or client index is not
  *                      within the range of 1 to MaxClients.
@@ -61,7 +61,7 @@ native velocity_by_aim(iIndex, iVelocity, Float:vRetValue[3]);
  * 
  * @param fVector       Input vector
  * @param vReturn       Output angle vector
- * 	
+ * 
  * @noreturn
  */
 native vector_to_angle(const Float:fVector[3], Float:vReturn[3]);
@@ -72,7 +72,7 @@ native vector_to_angle(const Float:fVector[3], Float:vReturn[3]);
  * @param vector        Input angle vector
  * @param FRU           One of the ANGLEVECTOR_* constants
  * @param ret           Output vector
- * 	
+ * 
  * @noreturn
  */
 native angle_vector(const Float:vector[3], FRU, Float:ret[3]);
@@ -81,7 +81,7 @@ native angle_vector(const Float:vector[3], FRU, Float:ret[3]);
  * Calculates the length of a vector.
  * 
  * @param vVector       Input vector
- * 	
+ * 
  * @return              Length of the input vector
  */
 native Float:vector_length(const Float:vVector[3]);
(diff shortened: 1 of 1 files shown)


HamletEagle:
This can be accomplished by using xs_vec_equal and xs_vec_set. This functions are a bit easier to use than xs_* but I'm not sure the difference is significant enough.

ClaudiuHKS:
**XS** is a file that has lots of data inside.
My functions are related to **Vector** only and they support both float and numeric parameters.

Mistrick:
Does pawn have any difference between set 0 or 0.0 to Float?

ClaudiuHKS:
warning 213: tag mismatch via http://www.amxmodx.org/webcompiler.cgi

Mistrick:
I mean in memory. Warning we can bypass and if there are no difference for what we need 2 funcs for int and float?

ClaudiuHKS:
I don't understand what you mean. Show me an example so I can understand.

Mistrick:
We know pawn has only cell type.
```
new Float:a = 0;
new Float:b = 0.0;
```
If in pawn VM there are no difference(example, all bytes are zero for first and second variant) we can use one func for int and float.

WPMGPRoSToTeMa:
@Mistrick there is no difference, we can write function that accepts both `_` and `Float`.

plugins/include/vector.inc
```diff
@@ -38,18 +38,18 @@ native get_distance(const origin1[3], const origin2[3]);
  * 
  * @param origin1       The first vector
  * @param origin2       The second vector
- * 	
+ * 
  * @return              The distance between two input vectors
  */
-native Float:get_distance_f(const Float:Origin1[3], const Float:Origin2[3]);
+native Float:get_distance_f(const Float:origin1[3], const Float:origin2[3]);
```
WPMGPRoSToTeMa: This breaks backward compatibility.

plugins/include/vector.inc
```diff
@@ -118,14 +118,38 @@ stock IVecFVec(const IVec[3], Float:FVec[3])
...
 }
+
+/**
+ * Checks if a vector is zero.
+ * 
+ * @param Vector        The input vector
+ * 
+ * @return              True if the vector is zero, false otherwise
+ */
+stock bool:IsVectorZero(const any:Vector[3])
```
WPMGPRoSToTeMa: `any` -> `{_, Float}`

plugins/include/vector.inc
```diff
@@ -118,14 +118,38 @@ stock IVecFVec(const IVec[3], Float:FVec[3])
...
+}
+
+/**
+ * Sets a vector to zero.
+ * 
+ * @param Vector        Vector
+ * 
+ * @noreturn
+ */
+stock SetVectorZero(any:Vector[3])
```
WPMGPRoSToTeMa: `any` -> `{_, Float}`

plugins/include/vector.inc
```diff
@@ -118,14 +118,38 @@ stock IVecFVec(const IVec[3], Float:FVec[3])
...
+/**
+ * Checks if a vector is zero.
+ * 
+ * @param Vector        The input vector
+ * 
+ * @return              True if the vector is zero, false otherwise
+ */
+stock bool:IsVectorZero(const any:Vector[3])
+{
+	return (Vector[0] == 0 && Vector[1] == 0 && Vector[2] == 0) ? true : false;
```
WPMGPRoSToTeMa: ` ? true : false` can be removed.

plugins/include/vector.inc
WPMGPRoSToTeMa: This breaks backward compatibility.

WPMGPRoSToTeMa:
Basically we can just add `ZERO_VECTOR` and then just use:
```Pawn
vec = ZERO_VECTOR;
xs_vec_equal(vec, ZERO_VECTOR);
```

plugins/include/vector.inc
WPMGPRoSToTeMa: @ClaudiuHKS with this it will accept only integer and float vectors.

plugins/include/vector.inc
ClaudiuHKS: OK fixing.

ClaudiuHKS:
If I add **public stock const Float: ZERO_VECTOR [ 3 ]** inside **AmxConst.INC**, it will act as **Float**.
It's impossible to declare it as **public stock const { _, Float }: ZERO_VECTOR [ 3 ]**.

Furthermore, I still don't want to use **xs_*** functions to check whether a **Vector** equals to zero.

So, do you still want me to add **public stock const Float: ZERO_VECTOR [ 3 ]** inside **AmxConst.INC**?

WPMGPRoSToTeMa:
@ClaudiuHKS I think integer vectors should be deprecated.

PR #552, opened 2018-09-06
Updated compiler copyright

voed:


OciXCrom:
Wouldn't it be better to grab the current year automatically with a function instead of updating it manually each year?

WPMGPRoSToTeMa:
@OciXCrom I think it should grab the year of the build or the year of the last changed source code file in compiler directory.

Issue #557, opened 2018-09-10
Feature Request: Add cooldown on thetime, timeleft and ff commands


Maxi605:
Basicly my request is to add a cooldown on how much you can use the thetime/timeleft/ff (Friendlyfire) commands, as it is right know you can flood them non stop.

Arkshine:
Just remembered that there is already an anti-flood plugin. Aren't you using it?

Maxi605:
I am, after spamming the command 3 or 4 times the anti-flood kicks in, but the chat (and audio) gets really flooded

WPMGPRoSToTeMa:
IMO `thetime` and `timeleft` should print and pronounce time only for the sender player.

Arkshine:
Both `thetime` and `timeleft` commands play the voice only on the player who typed the command.

WPMGPRoSToTeMa:
@Arkshine ah, it pronounces only to the sender, but prints for all.

Arkshine:
It might be useful to have a special cooldown to limit the message for all players. Like, if the limit is hit (meaning if a message for all players has been sent a moment ago), then we send only the message the sender.

WPMGPRoSToTeMa:
@Arkshine I don't know why it was designed to print time for all.

Arkshine:
Probably because it's an useful information for everyone, or can be handy depending on the situation. I'm fine with having new cvars to control such behavior.

OciXCrom:
Agreed. I submitted a pull request to add 2 new cvars in the plugin (#558).

Maxi605:
thetime and timeleft are very useful information for everyone, yes, but still, imagine if 10 out of 32 players start sending the same message in chat, it not only prints your line (username : thetime) but also 2 or 3 lines of the information of the command (Don't remember how many lines, i think is 2), i'm cool with having it as it is but with a cooldown to stop the flooding/abusing of the command (as Arkshine said).

PR #558, opened 2018-09-10
Add 2 new cvars

OciXCrom:
Requested in #557.

Added 2 new cvars in `timeleft.sma`:
- amx_time_print_to_all 0/1 -- controls whether or not the messages will be pritned to all players or just to the player who issued the command
- amx_time_show_cmd_in_chat 0/1 -- controls whether or not the commands will be shown in the chat or not

### DIFF
--- a/plugins/ns/timeleft.sma
+++ b/plugins/ns/timeleft.sma
@@ -19,6 +19,7 @@ new g_CountDown
 new g_Switch
 new Float:g_roundStartTime = 999999.9
 new bool:is_combat
+new g_amx_time_print_to_all, g_amx_time_show_cmd_in_chat
 
 public plugin_init() {
   register_plugin("TimeLeft",AMXX_VERSION_STR,"AMXX Dev Team")
@@ -28,6 +29,9 @@ public plugin_init() {
   register_clcmd("say timeleft","sayTimeLeft",0,"- displays timeleft")
   register_clcmd("say thetime","sayTheTime",0,"- displays current time")
   set_task(0.8,"timeRemain",8648458,"",0,"b")
+
+  g_amx_time_print_to_all = register_cvar("amx_time_print_to_all", "1")
+  g_amx_time_show_cmd_in_chat = register_cvar("amx_time_show_cmd_in_chat", "1")
   
   new szMapName[4]
   get_mapname(szMapName, charsmax(szMapName))
@@ -62,8 +66,8 @@ public sayTheTime(id){
   }
   new ctime[64]  
   get_time("%m/%d/%Y - %H:%M:%S",ctime,charsmax(ctime))  
-  client_print(0,print_chat, "The time:   %s",ctime )
-  return PLUGIN_CONTINUE
+  client_print(get_pcvar_bool(g_amx_time_print_to_all) ? 0 : id,print_chat, "The time:   %s",ctime )
+  return get_pcvar_bool(g_amx_time_show_cmd_in_chat) ? PLUGIN_CONTINUE : PLUGIN_HANDLED
 }
 
 public sayTimeLeft(id){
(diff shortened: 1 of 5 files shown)


plugins/ns/timeleft.sma
```diff
@@ -62,8 +66,8 @@ public sayTheTime(id){
   }
   new ctime[64]  
   get_time("%m/%d/%Y - %H:%M:%S",ctime,charsmax(ctime))  
-  client_print(0,print_chat, "The time:   %s",ctime )
-  return PLUGIN_CONTINUE
+  client_print(get_pcvar_bool(g_amx_time_print_to_all) ? 0 : id,print_chat, "The time:   %s",ctime )
+  return get_pcvar_bool(g_amx_time_show_cmd_in_chat) ? PLUGIN_CONTINUE : PLUGIN_HANDLED
```
HamletEagle: Could be just !get_pcvar_bool(g_amx_time_show_cmd_in_chat).

WPMGPRoSToTeMa: @OciXCrom it is ok, it should be explicit.

HamletEagle: I didn't say it's not ok("could be" not "should be").  But the line is too long and the shorter version is clear enough, seeing get_pcvar_BOOL clearly tells it will either return 0 or 1.

rsKliPPy: But you don't want to reutrn 0 or 1, you want `PLUGIN_CONTINUE` or `PLUGIN_HANDLED` (yes, I know they are the same underneath).

Arkshine: It's fine as it is. I agree `PLUGIN_` should be explicit. Just because it happens to be 0 and 1 doesn't mean you should abuse this behavior. If the long line is an issue, you can always cache the car's value before, totally fine.

HamletEagle: Of course there are more solutions, like caching as you said/even breaking the line. Also, I would not call "abusing" taking advantage of behavior that is set in stone.

Arkshine: `PLUGIN_` are symbols with their definitions and the documentation says you have to use them. If you don't explicit them, you imply their associated values that you are not supposed to use directly because irrelevant, not useful and confusing. If a beginner sees a code which returns some value from a condition, he may not understand why, whereas if you use `PLUGIN_*`, It's visually easier to understand and it's documented.
That's why if you do that, you're abusing with shortcuts, which is unhelpful.

HamletEagle: You are talking about explicitly returning numerical values instead of PLUGIN_* which I never said should be done. What I am saying is that I believe it is fine to return directly the result of an expression/native when doing that can save you from complicated/verbose/long lines of code.

WPMGPRoSToTeMa: @HamletEagle but there is a magic in converting between `bool` and `PLUGIN_*`.

HamletEagle: @WPMGPRoSToTeMa, the approach is not perfect, but lines that are too verbose are not perfect either. This could be solved in multiple other ways, as stated before: caching/breaking the line/using if-else instead of ternary,

Arkshine: You explicit the value underneath by returning a boolean. Those constants could have held others values, like -1, or 2. Actually, what about ``PLUGIN_HANDLED_MAIN`, are you going to return some boolean + 1? This just abusing the behavior for the fake sake of simplicity and you make harder to understand what's happening. As stated above, don't use magic.

Simplifying doesn't mean shorter and condensed code.

About the code, you should not use `LANG_PLAYER` with an index.

Also to make the code easier to understand, you can:
* Bind the cvars to variables
* Remove few terniary operator

Possible example:

```SourcePawn
public sayTimeLeft(id)
{
    new printTarget = id;

    if (CvarPrintToAll)
    {
        printTarget = 0;
        SetGlobalTransTarget(LANG_PLAYER);
    }

    if (CvarTimeLimit > 0.0)
    {
        if (CvarTimeVoice)
        {
            new voice[128];
            setTimeVoice(voice, charsmax(voice), 0, get_timeleft());
            client_cmd(id, "%s", voice);
        }

        client_print(printTarget, print_chat, "%l:  %d:%02d", "TIME_LEFT", (a / 60), (a % 60)):
    }
    else
    {
        client_print(printTarget, print_chat, "%l", "NO_T_LIMIT"):
    }

    return CvarShowCommandInChat ? PLUGIN_CONTINUE : PLUGIN_HANDLED:
}
```

OciXCrom: I meant "modernization" - using the new functions where possible.

PR #559, opened 2018-09-10
Fix admin chat displaying the wrong language

OciXCrom:
This is an ancient bug with a very simple fix. The problem is that when you write in the admin chat (`say_team @`), your prefix is translated in your language for all players instead of translating it in the language of the player who reads the message.

### DIFF
--- a/plugins/adminchat.sma
+++ b/plugins/adminchat.sma
@@ -218,7 +218,7 @@ public cmdSayAdmin(id)
 	
 	get_players(players, inum, "ch")
 	
-	for (new bool:is_sender_admin = !!is_user_admin(id), i = 0; i < inum; ++i)
+	for (new bool:is_sender_admin = is_user_admin(id) != 0, i = 0; i < inum; ++i)
 	{
 		pl = players[i]
 
--- a/plugins/adminchat.sma
+++ b/plugins/adminchat.sma
@@ -217,9 +217,9 @@ public cmdSayAdmin(id)
 	log_message("^"%s<%d><%s><>^" triggered ^"amx_chat^" (text ^"%s^")", name, userid, authid, message[1])
 	
 	if (is_user_admin(id)) // no diff here if admins have g_AdminChatFlag access or not, but we don't want to print "PLAYER"
-		format(message, charsmax(message), "(%L) %s :  %s", LANG_PLAYER, "ADMIN", name, message[1])
+		format(message, charsmax(message), "(%l) %s :  %s", "ADMIN", name, message[1])
 	else
-		format(message, charsmax(message), "(%L) %s :  %s", LANG_PLAYER, "PLAYER", name, message[1])
+		format(message, charsmax(message), "(%l) %s :  %s", "PLAYER", name, message[1])
 
 	get_players(players, inum, "ch")
 	
@@ -227,7 +227,10 @@ public cmdSayAdmin(id)
 	{
 		pl = players[i]
 		if (pl == id || get_user_flags(pl) & g_AdminChatFlag)
+		{
+			SetGlobalTransTarget(pl)
 			client_print(pl, print_chat, "%s", message)
+		}
 	}
 	
 	return PLUGIN_HANDLED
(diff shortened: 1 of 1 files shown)


plugins/adminchat.sma
```diff
@@ -217,9 +217,9 @@ public cmdSayAdmin(id)
 	log_message("^"%s<%d><%s><>^" triggered ^"amx_chat^" (text ^"%s^")", name, userid, authid, message[1])
 	
 	if (is_user_admin(id)) // no diff here if admins have g_AdminChatFlag access or not, but we don't want to print "PLAYER"
-		format(message, charsmax(message), "(%L) %s :  %s", id, "ADMIN", name, message[1])
+		format(message, charsmax(message), "(%L) %s :  %s", LANG_PLAYER, "ADMIN", name, message[1])
```
WPMGPRoSToTeMa: You can't do like that, you should re-format message for each player.

Arkshine: I think you can either use `SetGlobalTransTarget` + `%l` or using `%s` + `fmt("%L", id, "ADMIN")`

OciXCrom: I read somewhere that this is fixed in 1.9, guess it's wrong. I changed it to use `%l`.

WPMGPRoSToTeMa: @OciXCrom that is not how it works.

WPMGPRoSToTeMa: @OciXCrom it is not how it works.

Arkshine: Actually, I would cache `is_user_admin` and would do the check in the loop, and formatting should be made in `client_print`. It will be easier to translate `ADMIN` to the user's language.

OciXCrom: Done.

plugins/adminchat.sma
```diff
@@ -216,23 +216,18 @@ public cmdSayAdmin(id)
...
-	for (new i = 0; i < inum; ++i)
+	for (new bool:is_sender_admin = !!is_user_admin(id), i = 0; i < inum; ++i)
 	{
 		pl = players[i]
+
 		if (pl == id || get_user_flags(pl) & g_AdminChatFlag)
 		{
-			SetGlobalTransTarget(pl)
-			client_print(pl, print_chat, "%s", message)
+			client_print(pl, print_chat, "(%L) %s :  %s", pl, is_sender_admin ? "ADMIN" : "PLAYER", name, message[1])
```
Arkshine: Should be `%l`.

You use `%L` when you explicit the player's id.

You use `%l` when the native already passes the player's id such as `client_print(id, ...`.  You can do that because internally the player's id is set globally before the string is actually formatted. When the parser will check `%l`, it will use the saved value. Which means for native like formatting native `format`, you can use before `SetGlobalTransTarget` to set globally the player's id and then you can use `%l` without issue.

OciXCrom: I didn't know that. Thanks, I changed it.

plugins/adminchat.sma
```diff
@@ -216,20 +216,18 @@ public cmdSayAdmin(id)
...
 	
-	if (is_user_admin(id)) // no diff here if admins have g_AdminChatFlag access or not, but we don't want to print "PLAYER"
-		format(message, charsmax(message), "(%L) %s :  %s", id, "ADMIN", name, message[1])
-	else
-		format(message, charsmax(message), "(%L) %s :  %s", id, "PLAYER", name, message[1])
-
 	get_players(players, inum, "ch")
 	
-	for (new i = 0; i < inum; ++i)
+	for (new bool:is_sender_admin = !!is_user_admin(id), i = 0; i < inum; ++i)
```
WPMGPRoSToTeMa: I prefer `!= 0` instead of `!!`.

OciXCrom: Is there a difference?

Arkshine: It's just more friendly. It's easier to read "is not equal to" than "not not" :P.  You get it.

Arkshine:
Alright, looks good now. Thanks!

plugins/adminchat.sma
WPMGPRoSToTeMa: @OciXCrom yeah, but added/changed code should be as good as possible.

PR #565, opened 2018-09-11
Allow [i|scroll]message to include AMX Mod X vers.

ClaudiuHKS:
Allow [ _i_ | _scroll_ ]message to include AMX Mod X **version** as well.

### DIFF
--- a/plugins/imessage.sma
+++ b/plugins/imessage.sma
@@ -62,6 +62,7 @@ public infoMessage()
 	
 	get_cvar_string("hostname", hostname, charsmax(hostname));
 	replace(Message, charsmax(Message), "%hostname%", hostname);
+	replace(Message, charsmax(Message), "%version%", AMXX_VERSION_STR);
 	
 	set_hudmessage(values[0], values[1], values[2], X_POS, Y_POS, 0, 0.5, HOLD_TIME, 2.0, 2.0, -1);
 	


Arkshine:
You need to update `amxx.cfg` in the `/cstrike` and `/ns` folders as well.

Arkshine:
Yea, I did not merge it because I was uncertain. You get a valid point. Though not dramatic.

I was also unsure to show the version because you will now have something like "1.9.0.5200", and this would appear a little long & unnecessary. I'm okay to have %version%, but I don't think you need to specify the version in the original message.

ClaudiuHKS:
Both **1.9.0.5200** and **1.9.0** look OK to me.
How can I format **1.9.0.5200** to **1.9.0** _without_ **sscanf**, using a lot of functions and variables or even a **stock**?

Arkshine:
I don't know if it makes sense to provide the version in this message (as the purpose would just to tell the server uses AMXX and that's it), but you can either cut the string, like `version[3] = EOS` or you can use `AMXX_VERSION_MAJOR + '.' + AMXX_VERSION_MINOR`

ClaudiuHKS:
Thanks!

Arkshine:
You can actually use directly `AMXX_VERSION_MAJOR + "." + AMXX_VERSION_MINOR`.
You can concatenate strings in macro now with `+` and you can stringize with `#`.

PR #567, opened 2018-09-11
Modernize admincmd.sma

OciXCrom:
The changes include:

1. Properly named variables (`szName` instead of `name`, `g_` prefix for globals, etc).
2. Usage of new functions (`read_argv_int`, `get_players_ex`, `create_cvar`, etc).
3. Translations for command syntaxes.
4. Usage of `%N` instead of `%s<%d><%s><>` in log messages.
5. Usage of `%n` instead of `get_user_name` where possible.
6. Usage of `%l` instead of `%L` where possible.
7. Consistent brackets, indentation, spaces and semicolons.


plugins/admincmd.sma
```diff
@@ -165,18 +164,18 @@ public plugin_init()
 	register_clcmd("amx_showrcon", "cmdShowRcon", ADMIN_RCON, "<command line>")
 	register_clcmd("pauseAck", "cmdLBack")
 
-	rcon_password=get_cvar_pointer("rcon_password");
-	pausable=get_cvar_pointer("pausable");
-	timelimit=get_cvar_pointer( "mp_timelimit" );
-	p_amx_tempban_maxtime = register_cvar("amx_tempban_maxtime", "4320", FCVAR_PROTECTED);
+	g_pRconPassword=get_cvar_pointer("g_pRconPassword");
```
gtteamamxx: Are you sure you want to replace ("rcon_password") and other cvar names? :P

OciXCrom: You're viewing an older commit, that was a mistake I made which was fixed later.

gtteamamxx: Oh, my mistake ;)

OciXCrom: Actually I had it mistaken in another place in the code and I fixed it now, thanks for reminding me. The search and replace function failed me once again.

Arkshine:
1. I really disagree with this. It's unnecessary verbose and decreases readability. A 'name' implies a string, why would you want to add 'sz', it's redundant. A variable name should be simple and meaningful. If you name it properly, you don't need a prefix.

IgnacioFDM:
It's funny calling Hungarian notation "modernizing". I like g_ for globals. Not a fan of hungarian though.

Everything else, from the description (barely read the diff), seems good.

Mistrick:
In my mind hungarian notation make sense for global variables, for local variables you just write more code.

https://github.com/Mistrick/MapManager/blob/ed09e4d78ab5c6cb617b2e37c06a08df2cea7648/mapmanager.sma#L1598
https://github.com/Mistrick/MapManagerModular/blob/14d88d41cf922a8b4cb2a5625330bf33406fe05f/addons/amxmodx/scripting/map_manager_core.sma#L507
For now I think second variant is more readable.

OciXCrom:
Anyways, we should decide which style to use. I see the style is mixed up in most of the plugins and I want to make it consistent.

Arkshine:
> @Arkshine I personally find prefixed variables easier to understand what and how they're used for. The variable type helps a lot in some situations, especailly when you have two separate variables for string and integer. Take for example the ban commands, they use a `szMinutes` and a `iMinutes` variable - in my opinion it's better than `minutes_str` and `minutes_num`. The majority of people use this coding style, so I think the default AMXX plugins should adapt to it. This is only my opinion however, if you don't like it, should I revert the variable names?

I can't find the example you're talking about, show me some code. 
If you have to "separate", this means you name the variable wrongly or is not descriptive enough in the context. As simple as that. The code should be read as you would read English. Suffixes make it harder. People will always follow blindly what they see without thinking, like they're doing with the engine -> fakemeta conversion, right (in my very first plugin, I did not know how to name, I used `sz` on string because I saw others people doing that). I won't allow this, it's really ridiculous.

plugins/admincmd.sma
```diff
@@ -17,114 +17,79 @@
...
+	register_concmd("amx_plugins",		"cmdPlugins",	ADMIN_ADMIN,				"AMX_PLUGINS_SYNTAX",	.info_ml = true);
+	register_concmd("amx_modules",		"cmdModules",	ADMIN_ADMIN,				"AMX_MODULES_SYNTAX",	.info_ml = true);
+	register_concmd("amx_map",			"cmdMap",		ADMIN_MAP,					"AMX_MAP_SYNTAX",		.info_ml = true);
+	register_concmd("amx_extendmap",	"cmdExtendMap",	ADMIN_MAP,					"AMX_EXTENDMAP_SYNTAX",	.info_ml = true);
+	register_concmd("amx_cfg",			"cmdCfg",		ADMIN_CFG,					"AMX_CFG_SYNTAX",		.info_ml = true);
+	register_concmd("amx_nick",			"cmdNick",		ADMIN_SLAY,					"AMX_NICK_SYNTAX",		.info_ml = true);
+	register_concmd("amx_last",			"cmdLast",		ADMIN_BAN,					"AMX_LAST_SYNTAX",		.info_ml = true);
+	register_clcmd("amx_rcon",			"cmdRcon",		ADMIN_RCON,					"AMX_RCON_SYNTAX",		.info_ml = true);
+	register_clcmd("amx_showrcon",		"cmdShowRcon",	ADMIN_RCON,					"AMX_RCON_SYNTAX",		.info_ml = true);
+	register_clcmd("pauseAc",			"cmdLBack");
```
WPMGPRoSToTeMa: Please use spaces for alignment.

OciXCrom:
Lines `321` and `333` in my modified `admincmd.sma`. So, which style do you suggest using exactly? The default code has prefixes on some places and doesn't in other.

`g_global_variable_name` or `g_GlobalVariableName` or `g_Global_variable_name`?
`local_variable_name` or `localVariableName`?

plugins/admincmd.sma
OciXCrom: Done.

WPMGPRoSToTeMa: @OciXCrom looks perfect now, thanks!
image

Arkshine:
This is not a good example, because here you could use directly `read_argv_int`.
That's said, you check first the context, it's about retrieving values from a command's arguments.
I would probably use `new minuteArgument[32]` and `new minute`. Variant could exist. The point is just to try to name based on the context.

```SourcePawn
new szMinutes[32];
read_argv(2, szMinutes, charsmax(szMinutes));

new iMinutes = str_to_num(szMinutes);
```
```SourcePawn
new minutesArgument[32];
read_argv(2, minutesArgument, charsmax(minutesArgument));

new minutes = str_to_num(minutesArgument);
```
The latter is easier to read and less confusing.

OciXCrom:
@Arkshine I reverted the variable names.

Daniele386:
German translation is partially outdated, maybe you want to include possible changes from #620?

OciXCrom:
@Daniele386 I'll do it when the PR gets merged.

plugins/lang/admincmd.txt
```diff
@@ -91,10 +123,10 @@ PERM = für immer
 CLIENT_BANNED = Spieler "%s" gebannt
```
Daniele386: FOR_MIN = für %d Minuten

OciXCrom: There's no param for %d in the .sma file.

Daniele386: I thought at line 89? 
In the other languages you changed %s to %d?

OciXCrom: I didn't change any %s to %d.

Daniele386: Am I blind? 

I'm talking about lines 
- 89 (in your version line 121)

there is "%d"

and there are "%s":
- 12
- 164 (in your version line 196)
- 239 (in your version line 271)
- 314 (in your version line 346)
- 389 (in your version line 421)
- 464 (in your version line 496)
- 539 (in your version line 571)
- 614 (in your version line 646)
- 681 (in your version line 721)
- 764  (in your version line 796)
and so on...

But you haven't change the .sma in that point, have you?

Arkshine: You are not blind. `FOR_MIN` takes now an integer, so it should be `%d`.

OciXCrom: I see, I thought you meant line 89 in my version. I'll do the changes soon.

PR #573, opened 2018-09-12
Additions to hlsdk_const and cstrike_const

PartialCloning:
Added values to cstrike_const and hlsdk_const using regamedll_cs and reapi as reference.


Arkshine:
Looks good. 

Not sure if we should prefix `CS_` for those specific CS constants as well.

PM32:
I've checked this, values are ok :)

Arkshine:
Alright, let's go with that.

PR #575, opened 2018-09-12
Modernize and improve imessage.sma

OciXCrom:
The changes include:

1. Added new cvar `amx_imessage_only_dead` to control whether or not the messages will show only to dead players.
2. Added new cvar `amx_imessage_x_pos` to set the X position rather than having it hardcoded in the .sma file.
3. Added new cvar `amx_imessage_y_pos` to set the Y position rather than having it hardcoded in the .sma file.
4. Added new cvar `amx_imessage_holdtime` to set the holdtime rather than having it hardcoded in the .sma file.
5. Usage of `create_cvar` and `bind_pcvar_*` functions instead of the old ones.
6. Change `show_hudmessage` to `ShowSyncHudMsg` to prevent overlapping messages if the repeat time is shorter.
7. Consistent variable names, semicolons and brackets.

### DIFF
--- a/plugins/imessage.sma
+++ b/plugins/imessage.sma
@@ -25,7 +25,6 @@ enum _:MessageInfo
 	B
 }
 
-
 new Array:g_messages;
 new g_messagesNum;
 new g_hudObject;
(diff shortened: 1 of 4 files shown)


plugins/imessage.sma
```diff
@@ -14,110 +14,133 @@
 #include <amxmodx>
 #include <amxmisc>
 
-#define X_POS         -1.0
-#define Y_POS         0.20
-#define HOLD_TIME     12.0
+#define MAX_MSG_LEN   384
+#define TASK_MSG      12345
```
WPMGPRoSToTeMa: constify?

Arkshine:
Just a note about the next time you do a lot of changes in a plugin, please try to break down in multiple commits. As much as you can. It will be easier to review. Here, I would have made a commit for the style, the moving code, bind_cvar, etc.

OciXCrom:
@Arkshine Got it.

OciXCrom:
@IgnacioFDM That's not necessary. There's no need to force users to use semicolons. It's just a coding style. Half of the plugins have semicolons, so it's just right to add them where they're missing.

IgnacioFDM:
I mean, if the entire plugin is going to have semicolons, for consistency it's best to tell the compiler to keep it that way. Otherwise you end up with what you are trying to fix with this PR: Some lines have semicolons, some don't.

OciXCrom:
@IgnacioFDM I added `#pragma semicolon 1` before I submitted the PR to check if I missed semicolons somewhere, so no, there's not any missing. Future PRs should follow the same style and be careful when adding semicolons. The same thing is with the .inc files - there's no `#pragma` in them, but they all have semicolons.

plugins/imessage.sma
```diff
@@ -14,110 +14,133 @@
...
+
+	register_dictionary("imessage.txt");
+	register_dictionary("common.txt");
+
+	g_messages = ArrayCreate(MessageInfo);
+	g_hudObject = CreateHudSyncObj();
+
+	bind_pcvar_num(create_cvar("amx_imessage_only_dead", "0", _, "Set to 1 to show info messages only to dead clients", true, 0.0, true, 1.0), g_amx_imessage_only_dead);
+	bind_pcvar_float(create_cvar("amx_freq_imessage", "180", _, "Frequency in seconds of info messages", true, 0.0), g_amx_freq_imessage);
+	bind_pcvar_float(create_cvar("amx_imessage_x_pos", "-1.0", _, "X position for info messages", true, -1.0, true, 1.0), g_amx_imessage_x_pos);
```
WPMGPRoSToTeMa: Bound logic doesn't fit good here, because the valid values are `0.0 <= x <= 1.0 || x == -1.0`.

WPMGPRoSToTeMa: Bound logic doesn't fit good here, because the valid values are `0.0 <= x <= 1.0 || x == -1.0`.

WPMGPRoSToTeMa: @OciXCrom this can be an example for custom validation. And I think it would be better to add a param for custom filter in `create_cvar`.

OciXCrom: Yes, this may be very useful. The params in `create_cvar` definitely need some rework.

Issue #572, opened 2018-09-11
Admin plugin to increase the validity of the membership


all85100:
Admin plugin to increase the validity of the membership, in Users.ini set the validity period so that the member management more convenient.
I have someone else modified plug-ins, can learn from the next, so that can be modified faster! Plugins: Dl.lonzi.cn/cstrike/admin.sma, configuration files: Http://dl.lonzi.cn/cstrike/users.ini

Arkshine:
Can you do your changes on the `admin.sma` from 1.9 packages and posting a link again? Looks like you are using a very old version. It's really hard to see what you did and you are proposing.

all85100:
No, I mean in the admin plugin add a limit membership validity, my old plugin has this function is someone else to change, I mean you can see for reference, you can quickly increase this function, of course, you can not learn

不是,我是说在admin插件上增加一个限制会员有效期的功能,我的这个老插件有这个功能是别人改的,我意思是你看看可以借鉴一下,可以很快的增加这个功能,当然也可以不用借鉴

Arkshine:
Oh, you mean temporary admins. Like, for example: https://forums.alliedmods.net/showthread.php?p=112793 ?

HamletEagle:
Yes, he wants the option to set an expiration date for admins.

all85100:
https://forums.alliedmods.net/showthread.php?p=112793? 
This does not seem to me this good, but also added the configuration file, if you read my this plugin you will understand that this is very good, hoping to increase

OciXCrom:
It should be pretty easy to implement. An optional param needs to be added to users.ini which will hold the date and when reading the file, simply skip the line if the date has expired. Maybe also a cvar if the admin should automatically get removed from the file. I may give it a try later.

all85100:
OK,thinks

Issue #560, opened 2018-09-10
Miscstats plugin: Incorrect displaying of enemy remaining team name


all85100:
I like Amxmodx, some questions feedback, please fix Miscstats plugin: enemyremaining team name display error, when the enemy was killed one person, the remaining enemy team name wrong display set up their own team name, the remaining number and team name is not synchronized.
The old Miscstats plugin, after the death of the teammate, does not show the enemyremaining function, in the 1.9.0 also shows but does not correspond, I think this is a new feature, it should be, after the death of teammates show how many people and their team name, the enemy after the death shows how many people and the enemy team name, I come from China, English is not very good. If you don't understand me, please let me know.

WPMGPRoSToTeMa:
So, the team name is correct, but the player count is invalid?

all85100:
No, on the contrary, the team's death count is correct, the team's display name is wrong, but now its own team has also activated the Enemyremaining function after death, which was not shown previously, hoping as a new feature to fix him, After the death of the enemy or our teammates show the remaining number and the corresponding team name, and can be in the Stats.ini control show which side or all show,,, the necessity is not small, but also reasonable!

all85100:
Why didn't anyone fix this bug?

Arkshine:
Because someone needs to investigate. I can't do everything. Other people need to contribute.
Maybe @OciXCrom can try to give a look.

OciXCrom:
I honestly failed to understand what the actual issue is. Can you maybe post an image showing the problem?

Arkshine:
Looking at the beautiful code:
```SourcePawn
	new team = get_user_team(victim)
	if (EnemyRemaining && is_user_connected(victim))
	{
		if( TEAM_T <= team <= TEAM_CT )
		{
			new ppl[MAX_PLAYERS], pplnum, epplnum, a
			get_players(ppl, epplnum, "ae", team == TEAM_T ? "CT" : "TERRORIST")
			get_players(ppl, pplnum, "ae", team == TEAM_T ? "TERRORIST" : "CT")
			if( victim_alive )
			{
				for(a=0; a<pplnum; a++)
				{
					if( ppl[a] == victim )
					{
						ppl[a] = ppl[--pplnum]
						break
					}
				}
			}
			
			if (pplnum && epplnum)
			{
				new message[128], team_name[32]

				set_hudmessage(255, 255, 255, 0.02, 0.85, 2, 0.05, 0.1, 0.02, 3.0, -1)
[...]

PR #579, opened 2018-09-13
Modernize and improve scrollmsg.sma

OciXCrom:
The changes include:

1. Added cvar `amx_scrollmsg_color_red`.
2. Added cvar `amx_scrollmsg_color_green`.
3. Added cvar `amx_scrollmsg_color_blue`.
4. Added cvar `amx_scrollmsg_only_dead`.
5. Added cvar `amx_scrollmsg_speed`.
6. Added cvar `amx_scrollmsg_x_move_amount`.
7. Added cvar `amx_scrollmsg_x_start_pos`.
8. Added cvar `amx_scrollmsg_x_end_pos`.
9. Added cvar `amx_scrollmsg_y_pos`.
10. Usage of `create_cvar` and `bind_pcvar_*` functions instead of the old ones.
11. Change `show_hudmessage` to `ShowSyncHudMsg` to prevent overlapping messages if the repeat time is shorter.
12. Consistent variable names, semicolons and brackets.
13. Usage of `AutoConfigExec` - let me know if you want this to stay. I believe it's better this way rather than adding every cvar in `amxx.cfg`.

### DIFF
--- a/plugins/scrollmsg.sma
+++ b/plugins/scrollmsg.sma
@@ -38,15 +38,21 @@ public showMsg()
 	new a = g_startPos, i = 0
 
 	while (a < g_endPos)
+	{
 		g_displayMsg[i++] = g_scrollMsg[a++]
+	}
 
 	g_displayMsg[i] = 0
 
 	if (g_endPos < g_Length)
+	{
 		g_endPos++
+	}
 
 	if (g_xPos > 0.35)
+	{
 		g_xPos -= 0.0063
+	}
 	else
 	{
 		g_startPos++
@@ -101,7 +107,9 @@ public setMessage()
 		set_task(float(g_Frequency), "msgInit", 123, "", 0, "b")
 	}
 	else
+	{
 		server_print("%L", LANG_SERVER, "MSG_DISABLED")
+	}
 	
 	return PLUGIN_HANDLED
 }
--- a/plugins/scrollmsg.sma
+++ b/plugins/scrollmsg.sma
@@ -58,6 +58,7 @@ public plugin_init()
 	bind_pcvar_float(create_cvar( "amx_scrollmsg_y_pos",           "0.9",      _, "The Y position of the message",                true, -1.0, true, 1.0),  g_amx_scrollmsg_y_pos);
 
 	g_hud_object = CreateHudSyncObj();
+	AutoExecConfig();
 }
 
 public showMsg()
--- a/plugins/scrollmsg.sma
+++ b/plugins/scrollmsg.sma
@@ -17,99 +17,99 @@
 #define SPEED 0.3
 #define SCROLLMSG_SIZE	512
 
-new g_startPos
-new g_endPos
-new g_scrollMsg[SCROLLMSG_SIZE]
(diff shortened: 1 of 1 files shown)


ClaudiuHKS:
Very nice!
Just wanted to say that in my opinion, you should change **if(**, **while(**, **switch(** and **for(** to **if (**, **while (**, **switch (** and **for (**. Also, change **FLAG_A|FLAG_B|FLAG_C** to **FLAG_A | FLAG_B | FLAG_C**.
This isn't something you must do but in my opinion it makes the plugin more readable.

OciXCrom:
@ClaudiuHKS Thanks, I didn't notice them. I changed them.

all85100:
Great, it's all done. How do I download the new plugins that have changed the content?

all85100:
Usage of autoconfigexec, how did he change the plugin parameters to fixed? not using Amxx.cfg?

all85100:
@OciXCrom Can you fix this bug?
 https://github.com/alliedmodders/amxmodx/issues/560 
https://github.com/alliedmodders/amxmodx/issues/572
Have you tried ?

OciXCrom:
@all85100 You can't download them just yet. This is a pull request, it's not yet added in AMXX. Be patient.

plugins/scrollmsg.sma
```diff
@@ -14,94 +14,136 @@
...
-	register_srvcmd("amx_scrollmsg", "setMessage")
+	register_plugin("Scrolling Message", AMXX_VERSION_STR, "AMXX Dev Team");
+
+	register_dictionary("scrollmsg.txt");
+	register_dictionary("common.txt");
+
+	register_srvcmd("amx_scrollmsg", "setMessage", _, "<message> <frequency>");
+	bind_pcvar_string(get_cvar_pointer("hostname"), g_hostname, charsmax(g_hostname));
+
+	bind_pcvar_num(create_cvar(   "amx_scrollmsg_color_red",       "200",      _, "Red color amount",                             true, 0.0, true, 255.0), g_amx_scrollmsg_color_red);
```
voed: How about to implode it to one cvar `amx_scrollmsg_color 200 100 0`?

all85100:
This modified plugin, I tested, did not show the contents of amx_scrollmsg  "". Does it rely only on autoexecconfig generated files? The Amxx.cfg failed, Autoexecconfig generated files did not amx_
Scrollmsg parameters, My amxx.cfg file has amx_scrollmsg parameters and content, but it doesn't show.

It is best not to use AutoExecConfig, which is not very convenient for language translation and use.

plugins/scrollmsg.sma
OciXCrom: Not sure really, both seem fine to me. Let's hear more opinions.

OciXCrom:
@all85100 If a plugin uses AutoConfigExec() the cvars should be modified from the automatically generated .cfg file for that plugin. In this case amxx.cfg will have no effect because it's executed before the generated .cfg. It's not true that it can't be translated - it can. I don't see how amxx.cfg is better, especially when you need to add new cvars manually after each AMXX update.

PR #588, opened 2018-09-16
GeoIP: Add "reload" command

Duk0:


modules/geoip/geoip_main.cpp
```diff
@@ -156,12 +156,28 @@ void OnGeoipCommand()
 
 		MMDB_free_entry_data_list(entry_data_list);
 	}
+	else if (!strcmp(cmd, "reload"))
+	{
+		bool loaded = false;
+		
+		if (HandleDB.filename)
```
WPMGPRoSToTeMa: Consider:
```C++
bool loaded = HandleDB.filename != nullptr;

if (loaded)
```

Arkshine:
That would be welcomed to always provide some description, even if it seems obvious.

It seems to be a a good addition. It's true that the Geoip database file is loaded only at server start, and if you want to replace it, you would not have the choice than to stop the server.

Arkshine:
@WPMGPRoSToTeMa Adjusted things a little. Sounds good now?

modules/geoip/geoip_main.cpp
```diff
@@ -14,16 +14,19 @@
 #include "geoip_main.h"
 #include "geoip_natives.h"
 #include "geoip_util.h"
-#include <time.h>
+#include <ctime>
```
WPMGPRoSToTeMa: It is not required that cheader definitions are available in global namespace.
http://eel.is/c++draft/headers#4
http://eel.is/c++draft/headers#footnote-172
http://eel.is/c++draft/contents#footnote-166

Arkshine:
You always comment only when things are merged.

I believe this change has been made by my editor, probably because those headers are deprecated in c++.  In our context, I guess `time.h` should be kept.

modules/geoip/geoip_main.cpp
Arkshine: You always comment only when things are merged.

I believe this change has been made by my editor, probably because those headers are deprecated in c++. In our context, I guess time.h should be kept

PR #597, opened 2018-09-23
Move hamdata.ini to gamedata

Arkshine:
Time to move all the `hamdata.ini` data to gamedata files. 

* using `strtol` in the parser to allow number with a different format to be properly converted (such as `0x4`).
* The mod aliases are defined in `master.games.txt`


rsKliPPy:
Long overdue change, finally happening. \o/

Arkshine:
Let's :pray: the gamedata files contain the correct data. But it looks fine (just tested fastly cstrike).

PR #601, opened 2018-09-26
Fix a crash with set_tr/2 when -1 is passed in TR_pHit

Arkshine:
This reverts the changes in #433.

I think it has been imported from ReAMXX. Actually, I'm not sure why this check, but originally the intent could be to allow wordspawn, maybe?

The crash happens due to `id_to_edict` returning `nullptr` if index < 0 and because of the check `pHit` would be set with a `nullptr`.

Also, we should have likely `TR_pHit2` because as it is `TR_pHit` can't return or set wordspawn (0). It relies `FNullEnt` which includes wordspawn. If I'm not wrong, even `entity_get_edict2` is wrong then.

@WPMGPRoSToTeMa

### DIFF
--- a/modules/fakemeta/fm_tr.cpp
+++ b/modules/fakemeta/fm_tr.cpp
@@ -86,7 +86,7 @@ static cell AMX_NATIVE_CALL set_tr(AMX *amx, cell *params)
 	case TR_pHit:
 		{
 			e = TypeConversion.id_to_edict(*ptr);
-			if (*ptr != -1 && FNullEnt(e))
+			if (FNullEnt(e))
 				return 0; //TODO: return error
 			gfm_tr->pHit = e;
 			return 1;
--- a/modules/fakemeta/fm_tr.cpp
+++ b/modules/fakemeta/fm_tr.cpp
@@ -27,7 +27,6 @@ static cell AMX_NATIVE_CALL set_tr(AMX *amx, cell *params)
 	}
 
 	cell *ptr = MF_GetAmxAddr(amx, params[2]);
-	edict_t *e;
 
 	switch (type)
 	{
@@ -85,12 +84,13 @@ static cell AMX_NATIVE_CALL set_tr(AMX *amx, cell *params)
 		}
 	case TR_pHit:
 		{
-			e = TypeConversion.id_to_edict(*ptr);
-			if (FNullEnt(e))
-				return 0; //TODO: return error
-			gfm_tr->pHit = e;
+			const auto pEdict = TypeConversion.id_to_edict(*ptr);
+			if (pEdict == nullptr)
+			{
+				return 0;
+			}
+			gfm_tr->pHit = pEdict;
 			return 1;
-			break;
 		}
 	case TR_iHitgroup:
 		{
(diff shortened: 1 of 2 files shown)


PartialCloning:
Setting TR_pHit to -1 using set_tr2 does not crash the server for me and works as expected under the latest 1.9.0 build.

Arkshine:
I can reproduce and the crash makes sense looking at the code, as I explained above.

PartialCloning:
You could add "//TODO: return error"? The other parts that return 0 also return an error.

WPMGPRoSToTeMa:
`set_ptr` -> `set_tr` in the title.

> but originally the intent could be to allow wordspawn, maybe?

I think the same as you, but it should be better to ask the author of ReAMXX directly.

Anyway, why we can't just allow to set `pHit` to worldspawn?

Arkshine:
It will change plugins behavior, so for wordspawn, we should have `TR_pHit2`.

It's similar to `entity_get_edict` which returns 0 as invalid entity. If you suddenly return -1 as invalid and 0 for wordspawn, it will break things. That's why `entity_get_edict2` is added (though, it needs to be fixed, it uses `FNullEnt` to check invalid entity and it includes wordspawn).

WPMGPRoSToTeMa:
@Arkshine can you provide an example of changing plugins behavior for `get_tr`/`get_tr2`?

WPMGPRoSToTeMa:
@Arkshine my suggestion is to leave `get_tr`/`get_tr2` as it is for `TR_pHit`, but change the `set_tr`/`set_tr2` behavior.

Arkshine:
It's true for getting _pHit_ the engine will either return wordspawn or an entity, so to return -1 or 0, it will be the same the plugins. I guess it's fine for this one.

About setting, I'm not totally sure. Before, passing -1 or 0 would do nothing. After, only 0 would set wordspawn. I don't remember seeing plugin setting directly 0. Usually it's either an index or the result pHit. It looks safe.

Arkshine:
@WPMGPRoSToTeMa I've changed like your proposed. If good, we can go with that.

WPMGPRoSToTeMa:
@Arkshine we can temporarily add a log error for `0`/`FM_NULLENT` and if no one will report about that error during some period then we can safely add ability to set worldspawn and leave an error for `FM_NULLENT`. Also as I can see we need to update the documentation for `[get|set]_tr[2]`.

Arkshine:
If you put an error on -1, people will complain immediately. For example, the plugin I tested to fix the issues does `set_tr2(original_ptr, TR_pHit, get_tr2(new_trace, TR_pHit))`. 

Actually, I searched in around 500 plugins to see the `set_tr|2` usage, and the result is:
* value is either an entity or the result of TR_pHit (like above)
* seen 3 occurrences where 0 was used but it was used only to "reset", so it won't matter,  similar to:
```SourcePawn
set_tr(TR_vecEndPos,Float:{4096.0,4096.0,4096.0});
set_tr(TR_AllSolid,1);
set_tr(TR_pHit,0);
set_tr(TR_iHitgroup,0);
set_tr(TR_flFraction,1.0);
```
So, I think it's safe to go as it is.

As for the documentation, looks like no did review fakemeta include, I won't mind some help, I'm really horrible at it.

WPMGPRoSToTeMa:
@Arkshine ok.

Arkshine:
Documentation will be updated later. Ideally, someone should clean the whole `fakemeta.inc` include, there are some functions without proper documentation formatting.

PR #605, opened 2018-09-28
Fixed a typo in get_user_flags description

justgo97:


plugins/include/amxmodx.inc
```diff
@@ -1828,7 +1828,7 @@ native set_user_flags(index, flags = -1, id = 0);
  * @param index     Client index, 0 to set flags of server
  * @param id        Flag set id, ranging from 0 to 31
  *
- * @noreturn
+ * @return          Bitflag sum of client's admin flags, 0 if client is not connected
```
Arkshine: You can remove `, 0 if client is not connected` ; because 0 is only returned when there is an error (not connected), this return value can't be used.

Arkshine: You should not explicit return values which can't be used.

If you look at `get_user_weapons` source code, it errors on an invalid index, but it returns 0 if you are not in-game. The return value can be used.

Arkshine:
Sounds good now. Thanks!

plugins/include/amxmodx.inc
```diff
@@ -1828,7 +1828,7 @@ native set_user_flags(index, flags = -1, id = 0);
  * @param index     Client index, 0 to set flags of server
  * @param id        Flag set id, ranging from 0 to 31
  *
- * @noreturn
+ * @return          Bitflag sum of client's admin flags
```
WPMGPRoSToTeMa: Maybe just `Bitsum` instead of `Bitflag sum`? Or `Bit vector`?

justgo97: Well all other amxmodx.inc natives use Bitflag sum, so for consistency its better to either keep it as it is or change them all. also whats wrong with bitflag ? https://en.wiktionary.org/wiki/bitflag

WPMGPRoSToTeMa: @justgo97 hm, ok.

PR #606, opened 2018-09-28
Replaced magic number in is_user_admin

justgo97:


plugins/include/amxmisc.inc
```diff
@@ -26,7 +26,7 @@
 stock is_user_admin(id)
 {
 	new __flags = get_user_flags(id);
-	return (__flags > 0 && !(__flags & ADMIN_USER));
+	return (__flags > ADMIN_ALL && !(__flags & ADMIN_USER));
```
WPMGPRoSToTeMa: Should it logically be `!= ADMIN_ALL`?

justgo97: no difference ? 

__flags can't go less then 0 at least here.

Mistrick: What if 32-nd bit is set? This is undocumented but can be.

justgo97: admin flags are hardcoded in the default AmxModX admin system also i don't think that we can change this because of backward compatibility

rsKliPPy:
New flags can be added. You can actually do it yourself without breaking anything.

plugins/include/amxmisc.inc
Arkshine: Well `> 0` is just here to say "if the value is positive". 

It's also fine to use `ADMIN_ALL`, but it should be logically  `!= ADMIN_ALL` as stated WPMGPRoSToTeMa.

justgo97: updated

PR #607, opened 2018-09-29
Fix typo in cstrike_const.inc

HamletEagle:


Arkshine:
I believe it has been left for backward compatibility, so we can't remove this. Plugins are using this constant.

HamletEagle:
I assumed it was a recent typo. Seems like it was around since 2009, then yes, it can't be removed.

PR #610, opened 2018-09-30
Make UTF-8 safe others strings natives

Arkshine:
This makes the following string natives UTF-8 safe:
* `argparse`
* `parse`
* `strtok`
* `strtok2`
* `strbreak`
* `trim`

Basically, any natives using the C `isspace` function and that would throw an assert on Unicode character.

Note: this is not a full UTF-8 support, this is just for compatibility.

### DIFF
--- a/amxmodx/string.cpp
+++ b/amxmodx/string.cpp
@@ -1,4 +1,4 @@
-// vim: set ts=4 sw=4 tw=99 noet:
+﻿// vim: set ts=4 sw=4 tw=99 noet:
 //
 // AMX Mod X, based on AMX Mod by Aleksander Naszko ("OLO").
 // Copyright (C) The AMX Mod X Development Team.
@@ -210,6 +210,16 @@ void copy_amxmemory(cell* dest, cell* src, int len)
 		*dest++=*src++;
 }
 
+bool utf8isspace(const char* string)
+{
+	return utf8iscategory(string, 1, UTF8_CATEGORY_ISSPACE) != 0;
+}
+
+size_t utf8getspaces(const char* string)
+{
+	return utf8iscategory(string, SIZE_MAX, UTF8_CATEGORY_ISSPACE);
+}
+
 char* parse_arg(char** line, int& state)
 {
 	static char arg[3072];
@@ -218,7 +228,7 @@ char* parse_arg(char** line, int& state)
 	
 	while (**line)
 	{
-		if (isspace(**line))
+		if (utf8isspace(*line))
 		{
 			if (state == 1)
 				break;
@@ -919,6 +929,7 @@ static cell AMX_NATIVE_CALL amx_strtok(AMX *amx, cell *params)
 	int right_pos = 0;
 	unsigned int i = 0;
 	bool done_flag = false;
+	size_t spaces;
 	int len = 0;
 
 	//string[]
@@ -938,9 +949,9 @@ static cell AMX_NATIVE_CALL amx_strtok(AMX *amx, cell *params)
 	{
 		if (trim && !done_flag)
 		{
-			if (isspace(string[i]))
+			if ((spaces = utf8getspaces(string + i) > 0))
 			{
-				while (isspace(string[++i]));
+				i += spaces;
 				done_flag = true;
 			}
 		}
@@ -974,6 +985,7 @@ static cell AMX_NATIVE_CALL amx_strtok2(AMX *amx, cell *params)
 {
 	int left_pos = 0, right_pos = 0, len, pos = -1;
 	unsigned int i = 0;
+	size_t spaces;
 
 	char *string = get_amxstring(amx, params[1], 0, len);
 	char *left = new char[len + 1], *right = new char[len + 1];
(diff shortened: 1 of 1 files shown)


WPMGPRoSToTeMa:
Is this for 1.9 too?

Arkshine:
Yes, it's quite welcomed to not having crash when using common natives with utf-8.

Is there an isssue?

WPMGPRoSToTeMa:
@Arkshine ah, ok. I thought it can change the behavior because unicode space characters weren't considered, but if it was crashing then it is ok (like there was just no behavior).

Arkshine:
`UTF8_CATEGORY_ISSPACE` is for compatibility with C functions `is*`.

To properly check unicode space, I would need to use before first `UTF8_CATEGORY_SEPARATOR_SPACE` only.

amxmodx/string.cpp
```diff
@@ -989,9 +1001,9 @@ static cell AMX_NATIVE_CALL amx_strtok2(AMX *amx, cell *params)
 	int trim = params[7];
 
 	// ltrim left
-	if (trim & 1 && isspace(string[i]))
+	if (trim & 1 && (spaces = utf8getspaces(string, len)) > 0)
```
WPMGPRoSToTeMa: ` + i` missed?

Arkshine: `i` starts from 0 here. So, it should be fine.

amxmodx/string.cpp
```diff
@@ -938,9 +949,9 @@ static cell AMX_NATIVE_CALL amx_strtok(AMX *amx, cell *params)
 	{
 		if (trim && !done_flag)
 		{
-			if (isspace(string[i]))
+			if ((spaces = utf8getspaces(string + i, len) > 0))
```
WPMGPRoSToTeMa: Should `len` be `len - i`? (not only here)

Arkshine: `utf8iscategory` is a greedy function. It will try to match spaces as much as possible. Here we want to retrieve all the spaces all once until we hit a non-space character. We are not working with a specific maximum (apart the string itself of course), so it makes no difference.

Arkshine:
I've tested `trim`, `parse`, and runs the testsuite for `argparse` without issues.

amxmodx/string.cpp
WPMGPRoSToTeMa: @Arkshine so, it can be just `SIZE_MAX`?

Arkshine: You could, but I'm not sure why you would want to use some weird constant.
The documentation says _"Size of the input in bytes"_ ; so the string length is fine.

Arkshine:
It should be fine. `DEL` is part of the control category (`iscntrl`, `UTF8_CATEGORY_ISCNTRL`) and this will not be processed in the code above. 

If you look after, it calls `codepoint_read`:
https://github.com/alliedmodders/amxmodx/blob/2559fcf00aecb19fb6e6f8de846ae37cbcd6d572/third_party/utf8rewind/utf8rewind.c#L1405

And at the top of the function you see:
https://github.com/alliedmodders/amxmodx/blob/2559fcf00aecb19fb6e6f8de846ae37cbcd6d572/third_party/utf8rewind/internal/codepoint.c#L173-L179
It checks well `<=`.

amxmodx/string.cpp
Arkshine: I agree for `len - i`, the string is shifted. It would be the logical approach.
Since the string is guaranteed to be null-terminated we could use `SIZE_MAX`, but the usage here in this context is wrong.

WPMGPRoSToTeMa: @Arkshine we can remove the second arg from `utf8getspaces` and just pass `SIZE_MAX` in `utf8iscategory`, it isn't wrong.

WPMGPRoSToTeMa:
@Arkshine the behavior would be different on an architecture with signed `char`.

amxmodx/string.cpp
Arkshine: In this context, then yes, it's okay. It simplifies the code, and since the string is always null-terminated, it should be fine.

Arkshine:
Do we care about architecture other than x86-32?

WPMGPRoSToTeMa:
@Arkshine we don't, but the author of utf8rewind should care. It really looks like a bug, because in other functions the author uses `uint8_t*` wrapper around `char*`.

Arkshine:
Can you create an issue on the official repository? https://bitbucket.org/knight666/utf8rewind/issues?status=new&status=open
That would be really helpful and interesting.

WPMGPRoSToTeMa:
@Arkshine ok, I'll try.

Arkshine:
@WPMGPRoSToTeMa Looks like it's going to take a while waiting for an answer. I think it's fine to merge as it is. Right now, we don't care about other architecture, and string used in natives there are guaranteed to be null-terminated. Right?

WPMGPRoSToTeMa:
@Arkshine yeah, :shipit:.

Issue #614, opened 2018-10-03
undefined symbol: pow10


the-swank:
https://github.com/alliedmodders/amxmodx/blob/0b5584a334a0adb0baaaf5df620f329732f20922/compiler/libpc300/sc2.c#L603
https://github.com/alliedmodders/amxmodx/blob/0b5584a334a0adb0baaaf5df620f329732f20922/compiler/libpc300/sc2.c#L685

Freeman-AM:
Someone should close this @the-swank or @Arkshine 
It's now merged.

PR #615, opened 2018-10-03
Compiler: Fix undefined symbol: pow10

the-swank:
https://github.com/alliedmodders/amxmodx/issues/614


WPMGPRoSToTeMa:
Should we always define it?

Arkshine:
`__GNUC__` has been changed to `__APPLE__` in this commit: https://github.com/alliedmodders/amxmodx/commit/46123d359fa8d93a296902a0f830eafe8f31b344#diff-154039b13c663f3e047598b4e8eea478R589

Looks like it's fine to use directly `pow`.

PR #623, opened 2018-10-08
Workaround for incorrect heap deallocation on conditional operator

### LINKED ISSUE #621: Runtime error when using the conditional operator with fmt
# Help us help you
  - [x] I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/amxmodx/issues).
  - [x] I have tried my absolute best to reduce the problem-space and have provided the absolute smallest test-case possible.
  - [x] I can always reproduce the issue with the provided description below.

# Environment
  * Operating System version: Windows 10 64 Bits
  * Game/AppID (with version if applicable): Half-Life 7561
  * Current AMX Mod X version: 1.8.3 5169
  * Current Metamod version: v1.21p37
# Description
I'm trying to show an empty message in case some condition becomes false, the best think that came to my mind to improve readability is to use the ternary operator, but it shows me a runtime error...

# Problematic Code (or Steps to Reproduce)
Put this code in plugin_init, compile and test it
  ```PAWN
	server_print(id, "%s", true ? fmt("Hello") : "");
  ```

# Logs
[AMXX] Run time error 8: heap low

WPMGPRoSToTeMa:
Fixes #621.

Using conditional operator can produce different heap allocation size in different branches (one of them can also produce no allocation). The problem is that heap deallocation is executed unconditionally for both branches. It is even incorrect if both branches use the same allocation size. This workaround fixes this problem by adding an additional heap allocation for a branch with a minimal allocation size (or without an allocation). It also fixes a problem when allocation sizes are the same.

@rtxa test it please.

### DIFF
--- a/compiler/libpc300/sc3.c
+++ b/compiler/libpc300/sc3.c
@@ -1012,11 +1012,13 @@ static int hier13(value *lval)
   if (matchtoken('?')) {
     int flab1=getlabel();
     int flab2=getlabel();
+    int flab3=getlabel();
     value lval2 = {0};
     int array1,array2;
 
     int orig_heap=decl_heap;
     int diff1=0,diff2=0;
+    int maxdiff;
     if (lvalue) {
       rvalue(lval);
     } else if (lval->ident==iCONSTEXPR) {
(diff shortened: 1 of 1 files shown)


rtxa:
Thanks, I want to test it but, how can I compile this? Can you leave the compiler executable?

WPMGPRoSToTeMa:
@rtxa 
[amxxpc32.zip](https://github.com/alliedmodders/amxmodx/files/2458599/amxxpc32.zip)

rtxa:
Thank you, it works perfect.

Arkshine:
Is there a reason you are not using the same code from upstream?

WPMGPRoSToTeMa:
@Arkshine unfortunately I found upstream fix after creating this PR.

Arkshine:
Well, if possible, I would prefer to use the upstream code. It's around since a much longer time and likely less prone to side-effects. I've no idea if what you did is sane.

PR #628, opened 2018-10-12
Add set_hudmessage_ex native

### LINKED ISSUE #625: Why there aren't parameters for color2 in set_hudmessage?
Why this is hardcoded?
```PAWN
g_hudset.r2 = 255;
g_hudset.g2 = 255;
g_hudset.b2 = 250;
```
And this not?
```PAWN
g_hudset.r1 = static_cast<byte>(params[1]);
g_hudset.g1 = static_cast<byte>(params[2]);
g_hudset.b1 = static_cast<byte>(params[3]);`
```
Code from [here](https://github.com/alliedmodders/amxmodx/blob/c6c4e79b96c5f5b64767da7e8ba9fd3541984450/amxmodx/amxmodx.cpp#L480)

This is specially useful when you set a hud message with effect 2 (Scan Out), the definition is from [here](https://developer.valvesoftware.com/wiki/Game_text)

> Color2 <color255>
    The scanning color for the letter being scanned if the Text Effect keyvalue is set to Scan Out - usually a different shade of Color1.

rtxa:
Fix #625 
This basically add the missing parameters for color 2, that are used in effect 2 (Scan Out).


WPMGPRoSToTeMa:
Which `UTIL_HudMessage` will be called by `show_hudmessage`? One of them doesn't consider second colors.

rtxa:
@WPMGPRoSToTeMa why you say that?

WPMGPRoSToTeMa:
@rtxa 
> Which `UTIL_HudMessage` will be called by `show_hudmessage`? One of them doesn't consider second colors.

https://github.com/alliedmodders/amxmodx/blob/605d1a5b1eb2576f10805b61a77aac8d33684269/amxmodx/util.cpp#L225-L227
https://github.com/alliedmodders/amxmodx/blob/6e9947b64faabc5d1c0b3cfac626021e8ed48fd9/public/sdk/amxxmodule.cpp#L3009-L3011

> Btw, there are no second colors in dhud messages.

https://github.com/alliedmodders/amxmodx/blob/605d1a5b1eb2576f10805b61a77aac8d33684269/amxmodx/util.cpp#L252

rtxa:
@WPMGPRoSToTeMa thanks, anyway I haven't finish this yet, and I will modify the title later.

OciXCrom:
I don't think there's a need for a new native, especially one with differently placed parameters than the current one. Simply add the parameters in the existing `set_hudmessage` native after the `channel` argument.

rtxa:
@OciXCrom, I did not have any troubles in memorize the new parameters because I know I should type 3 more colors after the first three. I have taken a look to your idea, it is interesting, I don't know if it is practical because you can not see the settings directly and you have to go to one side of the code to look at them or what if you have a lot of settings, that would become a hell maybe, anyway, I think you should make another pullrequest for that.
@Arkshine take a look please.

OciXCrom:
That's exactly how settings should work. They should be in the beginning of the code, not hardcoded down the lines.

WPMGPRoSToTeMa:
@OciXCrom I think that's not true for simple cases.

francoromaniello:
Can this be added ? also i' make a pull request in rtxa branch to add alpha support

PR #631, opened 2018-10-14
Improve AdminSlot plugin

Arkshine:
Originally initiated by @wopox1337 but took over because incomplete and it needed to be to split in multiple commits.

Notable changes:
* Cvars behavior. This is not really a bug, but before if you play with the cvars  `sv_visiblemaxplayers` is not adjusted (or not immediately). E.g. both cvars is set, if you don't hide the slots anymore,  `sv_visiblemaxplayers` is not updated.  Now, when a cvar change is detected and if required, `sv_visiblemaxplayers` update is immediate. This removes the ugly task as well.
* When `sv_visiblemaxplayers` needs to be reset, this will use -1 (default value), and not max clients.
* The cvars have now a min/max bounds. This will reduce the chance of abuse.
* If `amx_reservation` or `amx_hideslots` is set to 0, and if `sv_visiblemaxplayers` is set with a positive value, then it will be reset to -1.
* Added description to cvar. Same as the command, ML should be used for cvar. Ideally, a new param should exist, but it's fine to use `LANG_SERVER` for now.

### DIFF
--- a/plugins/adminslots.sma
+++ b/plugins/adminslots.sma
@@ -49,7 +49,9 @@ public client_authorized(id)
 	if (access(id, ADMIN_RESERVATION) || (players <= limit))
 	{
 		if (get_pcvar_num(g_HidePtr))
+		{
 			setVisibleSlots(players, limit);
+		}
 		return;
 	}
 
@@ -69,9 +71,13 @@ setVisibleSlots(players, limit)
 	new num = players + 1;
 
 	if (players == MaxClients)
+	{
 		num = MaxClients;
+	}
 	else if (players < limit)
+	{
 		num = limit;
+	}
 
 	set_pcvar_num(g_sv_visiblemaxplayers, num);
 }
--- a/plugins/adminslots.sma
+++ b/plugins/adminslots.sma
@@ -52,7 +52,7 @@ public client_authorized(id)
 			setVisibleSlots(players, limit)
 		return
 	}
-	
+
  	server_cmd("kick #%d ^"%L^"", get_user_userid(id), id, "DROPPED_RES")
 }
 
@@ -72,6 +72,6 @@ setVisibleSlots(players, limit)
 		num = MaxClients
 	else if (players < limit)
 		num = limit
-	
+
 	set_pcvar_num(g_sv_visiblemaxplayers, num)
 }
--- a/plugins/adminslots.sma
+++ b/plugins/adminslots.sma
@@ -14,41 +14,51 @@
 #include <amxmodx>
 #include <amxmisc>
 
-new g_ResPtr;
-new g_HidePtr;
+new CvarReservation;
+new CvarHideSlots;
 new g_sv_visiblemaxplayers;
 
 public plugin_init()
 {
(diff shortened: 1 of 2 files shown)


plugins/adminslots.sma
```diff
@@ -14,64 +14,97 @@
...
+
+new CvarHandleMaxVisiblePlayers;
 
 public plugin_init()
 {
-	register_plugin("Slots Reservation", AMXX_VERSION_STR, "AMXX Dev Team")
-	register_dictionary("adminslots.txt")
-	register_dictionary("common.txt")
-	g_ResPtr = register_cvar("amx_reservation", "0", FCVAR_PROTECTED)
-	g_HidePtr = register_cvar("amx_hideslots", "0")
```
OciXCrom: `bind_pcvar_num`?

plugins/lang/adminslots.txt
```diff
@@ -1,5 +1,7 @@
 [en]
 DROPPED_RES = Dropped due to slot reservation
+CVAR_RESERVATION = Amount of slots to reserve
+CVAR_HIDESLOTS = If you set this to 1, you can hide slots on your server.^nIf server "full" of public slots and slots hidden, you must manually connect with connect console command.
```
OciXCrom: > If server "full" of public slots and slots hidden, you must manually connect with connect console command.

=>

> If the server's public and hidden slots are full, you must manually connect with the "connect" console command.

plugins/adminslots.sma
Arkshine: Cvars are hooked in another commit.

plugins/lang/adminslots.txt
Arkshine: I used the original text, but yes, I can change it.

wopox1337:
The next step should be to add an API for the plugin

OciXCrom:
Maybe also add a cvar to change the slot reservation flag?

plugins/lang/adminslots.txt
```diff
@@ -1,5 +1,7 @@
...
 
 [de]
 DROPPED_RES = Sorry, dieser Slot ist reserviert
```
Daniele386:
```suggestion
DROPPED_RES = Sorry, dieser Slot ist reserviert
CVAR_RESERVATION = Anzahl der zu reservierenden Slots
CVAR_HIDESLOTS = Wenn diese CVAR auf 1 gesetzt wird, können Slots auf Ihrem Server ausgeblendet werden.^nWenn die öffentlichen Slots auf dem Server "voll" sind und weitere freie Slots ausgeblendet sind, müssen Sie die Verbindung manuell über den Konsolenbefehl "connect" herstellen.
```

**--> updated german translation**

wopox1337:
need to be merged @Arkshine

wopox1337:
bump

wopox1337:
@Arkshine also merge this

Arkshine:
Ya.  @wopox1337 I remember testing it extensively (and actually did you too?), but don't hesitate to play with it if you see any issues.

PR #656, opened 2019-01-12
Double number of buffers.

In-line:
Dunno why it wasn't done before, so it's worth to try or at least know the reason why nobody did this.


WPMGPRoSToTeMa:
I think it would be better to create new functions in modules API that don't use amxx internal buffers, but instead read strings directly into your buffer.

PR #658, opened 2019-01-13
Removed garbage generated by VS

voed:
😒


Arkshine:
It should have been ignored with `.gitignore` but looks like it has been updated later.
Thanks!

PR #664, opened 2019-01-26
Add square distance functions to xs.inc

IgnacioFDM:
And remove whitespace

### DIFF
--- a/plugins/include/xs.inc
+++ b/plugins/include/xs.inc
@@ -11,7 +11,7 @@
 /**
  *  XS Library
  *  Version 0.1
- *  
+ *
  * MACROS THAT YOU CAN DEFINE BEFORE INCLUDING XS.INC:
  * XS_FLEQ_TOLERANCE:
  *  Tolerance that is used for XS_FLEQ float nearly-equal comparisions
@@ -34,7 +34,7 @@
  *  DEFAULT: 2048
  * XS_REPLACEBUF_SIZE
  *  DEFAULT: 3072
- * 
+ *
  * NOTES:
  *  On AMX, VexdUM is required for some math functions
  *  xs__ / XS__ (2 underscores) stuff is meant to be intern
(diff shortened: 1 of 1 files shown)


IgnacioFDM:
I haven't benchmarked this in pawn, but in C++ this kind of optimization usually leads to 2x speedup.

If you are trying to compare who is closest to some point, it just doesn't make sense to use distance instead of squared distance, since you only care whether the number is bigger or smaller.

IgnacioFDM:
But to be honest I think what's needed the most is to copy all xs_ stuff to C++.

Stuff like get_distance_f is like 5x faster than xs_vec_distance. Anything to do with arrays is so slow in pawn. 

Currently you have to choose between a quite tidy and extensive library under xs.inc, or the very few and messy (as in function name and parameters) vector.inc functions.

I'd say to copy everything math related from xs.c onto amxmodx, with the math_ prefix.

WPMGPRoSToTeMa:
I'm not sure that porting it to C++ will gain performance, because calling natives has some cost.

WPMGPRoSToTeMa:
> If you are trying to compare who is closest to some point, it just doesn't make sense to use distance instead of squared distance, since you only care whether the number is bigger or smaller.

For this case we can introduce new native function, so it would be only 1 native call instead of 2 and no sqrts.

IgnacioFDM:
>  so it would be only 1 native call instead of 2 

I don't see what you mean by 1 native call instead of 2, but anyway that's besides the point. That's just solving for a specific use case. This PR adds a basic vector function that is available in pretty much any game engine

plugins/include/xs.inc
```diff
@@ -627,13 +627,44 @@ stock Float:xs_vec_distance_2d(const Float:vec1[], const Float:vec2[])
 	               (vec1[1]-vec2[1]) * (vec1[1]-vec2[1]));
 }
 
+/**
+ * Computes the square of the distance between two vectors (points).
+ * This is faster than the distance.
```
Arkshine: What do you mean by that?

Arkshine:
Well, I guess it's fine.

PR #667, opened 2019-01-28
Menu buttons stayed english in Plugin's cvar/command menus

regener:
That worked, but if there is any better solution to get it working feel free to edit :D


Arkshine:
`%l` will look at the current player's id saved globally.

Here, you need to either call before  `SetGlobalTransTarget(id)` or you can use directly `%L, id` instead of `%l`

PR #669, opened 2019-01-29
Changed wpindex to attackerindex

SmileYzn:
Changed wpindex to attackerindex, used the description as reference!

### DIFF
--- a/plugins/include/csstats.inc
+++ b/plugins/include/csstats.inc
@@ -185,7 +185,7 @@ native get_user_vstats(index, victim, stats[STATSX_MAX_STATS], bodyhits[MAX_BODY
  *          6 - Damage
  *
  * @param index     Client index
- * @param wpnindex  Attacker client index, or 0 to retrieve the statistics from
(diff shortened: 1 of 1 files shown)


rsKliPPy:
This is actually a breaking change, but I doubt anyone actually ever wrote `.wpnindex = value` while calling the function.

SmileYzn:
The problem is usage of function istelf (documentation), i did not found anything that indicates the wpindex is client index in really. Only did a try error test.

Anyway someone have made a review of some descriptions in functions of csstats.inc that is a error in review.

in really, nothing to worry

voed:
I think better just add comment to documentation that it's not a weapon index, but client index

SmileYzn:
Someone can explain how this break the compatibility, or just give-me some documentation about it? Im back to amxx coding this week, and forgot some terms lol

rsKliPPy:
If you had a function
`native func(param_a, param_b);`
you could call it as
`func(.param_a = 69, .param_b = 420)`

That's why changing parameter names is a breaking change.

SmileYzn:
oh yes, i forgot about named params, thanks for explain

PR #677, opened 2019-02-05
Add get_pcvar_name native

KaidoRen:
I believe that this is a necessary native.

### DIFF
--- a/amxmodx/cvars.cpp
+++ b/amxmodx/cvars.cpp
@@ -374,6 +374,19 @@ static cell AMX_NATIVE_CALL get_pcvar_bounds(AMX *amx, cell *params)
 	return hasBound;
 }
 
+// get_pcvar_name(pcvar, string[], maxlen)
+static cell AMX_NATIVE_CALL get_pcvar_name(AMX *amx, cell *params)
+{
+	cvar_t *ptr = reinterpret_cast<cvar_t *>(params[1]);
+	if (!ptr)
+	{
+		LogError(amx, AMX_ERR_NATIVE, "Invalid CVAR pointer");
+		return 0;
+	}
+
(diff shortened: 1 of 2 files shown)


SmileYzn:
Since it will get information about registred pointer var, why not put more info than name in this native?
I think it will avoid future natives or changes on this.

SmileYzn:
I mean get all informations about pointer variable in one function.

like an menu_item_getinfo, since sometimes people needs to call 2 or 3 times different functions for this, you can only need to call one time for get more info that you needs.

KaidoRen:
> I mean get all informations about pointer variable in one function.
> 
> like an menu_item_getinfo, since sometimes people needs to call 2 or 3 times different functions for this, you can only need to call one time for get more info that you needs.

What does this have to do with this pull request?

KaidoRen:
Initially there were several functions for obtaining different information. I see no point now in creating one for all the information at once.

Arkshine:
I think I would pass the cvar name in the forward params instead because as native, you won't have a usage outside the forward.

@KaidoRen Can you show a concrete example of how a cvar name in the forward would be needed?

voed:
> I can get the name of the cvar and compare it with part of the name.

But you can do same with forward parameter

KaidoRen:
@voed 

> Exactly the same action may be needed in other places where cvar pointer is known, not necessarily in the handler hook_cvar_change. Therefore, I assure you that there is a benefit from the function.

OciXCrom:
Printing the cvar name in chat or console can also be useful. For example:

```
This feature is currently disabled, set the cvar %s to 1 in order to enable it.
```

SmileYzn:
Will be useful like a get_plugins_cvar, but without retrieve all cvars, only passed by pointer cvar handle

PR #685, opened 2019-02-11
Remove additional parameter

CrazY770:
There's no "brightness" parameter in TE_DLIGHT. Probably will cause server crash if it be passed.
https://github.com/alliedmodders/hlsdk/blob/master/dlls/islave.cpp#L367


Arkshine:
Thanks!

PR #687, opened 2019-02-27
Add new public var with mapname and new constant

KaidoRen:


plugins/include/amxconst.inc
```diff
@@ -29,6 +29,11 @@
  */
 #define MAX_STRING_LENGTH 16384
 
+/**
+ * The maximum buffer size required to store a map's name.
+ */
+#define MAX_MAPNAME_LENGTH 256
```
WPMGPRoSToTeMa: Why 256? [It is limited to be 64](https://github.com/dreamstalker/rehlds/blob/237ee464644b5c8d9485c1e4943019a24ade76e1/rehlds/engine/server.h#L114).

plugins/include/amxconst.inc
```diff
@@ -44,6 +49,11 @@
  */
 public stock const MaxClients; 
 
+/**
+ * Current map name
```
WPMGPRoSToTeMa: Please add to the description is it map file path (`maps/de_dust2.bsp`) or just a map name (`de_dust2`).

plugins/include/amxconst.inc
KaidoRen: With a margin :D

plugins/include/amxconst.inc
KaidoRen: @WPMGPRoSToTeMa if this were the path to the map, I would point out that this is the path.

amxmodx/CPlugin.cpp
```diff
@@ -154,6 +154,11 @@ int CPluginMngr::loadPluginsFromFile(const char* filename, bool warn)
 				*get_amxaddr(plugin->getAMX(), addr) = gpGlobals->maxClients;
 			}
 
+			if (amx_FindPubVar(plugin->getAMX(), "MapName", &addr) != AMX_ERR_NOTFOUND)
+			{
+				set_amxstring(plugin->getAMX(), addr, STRING(gpGlobals->mapname), PLATFORM_MAX_PATH);
```
Nextra: `PLATFORM_MAX_PATH` is not correct anymore.

WPMGPRoSToTeMa: It should be changed to `64 - 1` according to `MAX_MAPNAME_LENGTH`. Is there any named constant that can be used here?

KaidoRen: I did not find one. And I didn't want to create a new one.

WPMGPRoSToTeMa: @KaidoRen but anyway you should change it to `64 - 1`.

Issue #712, opened 2019-05-07
New public vars for plugin data


KaidoRen:
Hi everyone! I have a ready opportunity to do so:
```PAWN
#include <amxmodx>

public stock const PluginName[] = "test";
public stock const PluginVersion[] = "1.0";
public stock const PluginAuthor[] = "Kaido Ren";

public plugin_init()
{
    
}
```

The bottom line is that you don't need to call the register_plugin function.
Before you make a pull request, I would like to clarify whether there is a need for this?

voed:
Good idea. Also it would be great to add "url" variable for support/update link

eat1k:
I agree. It would be a really nice feature.

wopox1337:
also pls add "description"

fl0werD:
Maybe better use "plugin_info_t struct" like metamod or "public Plugin myinfo" like sourcemod?

KaidoRen:
@fl0werD Implement the structs in the compiler - I will do it like this

fl0werD:
@KaidoRen use array

KaidoRen:
This
```PAWN
public stock const PluginInfo[][] = {
    "test",
    "1.0",
    "Kaido Ren"
};
```
 is not equal to this (sm)
```PAWN
public Plugin myinfo =
{
	name = "test",
	author = "Kaido Ren",
	description = "",
	version = "1.0",
	url = ""
};
```

without explicit indication of arguments it looks incomprehensible and ridiculous.

fl0werD:
`plugin_info_t Plugin_info =
{
	META_INTERFACE_VERSION,			// ifvers
	"",				// name
	"",					// version
	"",				// date
	"fl0wer",				// author
	"",	// url
	"visible",				// logtag, all caps please
	PT_CHANGELEVEL,				// (when) loadable
	PT_ANYTIME,				// (when) unloadable
};`

Say it to metamod.

KaidoRen:
@fl0werD there at least a structure, but not a normal array.

fl0werD:
I can advice you to use default args in register_plugin.

WPMGPRoSToTeMa:
@fl0werD you can use [designated initializers](https://en.cppreference.com/w/cpp/language/aggregate_initialization#Designated_initializers) for `plugin_info_t` structure since C++20.

fl0werD:
@WPMGPRoSToTeMa I think it will added in amxx 3.0

fl0werD:
[...]

PR #714, opened 2019-05-09
Public vars for plugin data

KaidoRen:
```PAWN
#include <amxmodx>

public stock const PluginName[] = "test";
public stock const PluginVersion[] = "1.0";
public stock const PluginAuthor[] = "Kaido Ren";
public stock const PluginURL[] = "https://github.com/KaidoRen";
public stock const PluginDescription[] = "This is test plugin";

public plugin_init()
{
    
}
```
image

image

image

image

image


amxmodx/amxmodx.cpp
```diff
@@ -1381,14 +1396,16 @@ static cell AMX_NATIVE_CALL show_menu(AMX *amx, cell *params) /* 3 param */
 	return 1;
 }
 
-static cell AMX_NATIVE_CALL register_plugin(AMX *amx, cell *params) /* 3 param */
+static cell AMX_NATIVE_CALL register_plugin(AMX *amx, cell *params) /* 5 param */
 {
 	CPluginMngr::CPlugin* a = g_plugins.findPluginFast(amx);
```
wopox1337:
```
enum { arg_title, arg_version, arg_author, arg_url, arg_description }
```

Mistrick: it should be count from 1 or add first arg_count

amxmodx/amxmodx.cpp
```diff
@@ -1381,14 +1396,16 @@ static cell AMX_NATIVE_CALL show_menu(AMX *amx, cell *params) /* 3 param */
 	return 1;
 }
 
-static cell AMX_NATIVE_CALL register_plugin(AMX *amx, cell *params) /* 3 param */
+static cell AMX_NATIVE_CALL register_plugin(AMX *amx, cell *params) /* 5 param */
```
In-line: Please, check arg count and make it backward compatible, so old plugins dont need to be recompiled

amxmodx/srvcmd.cpp
```diff
@@ -63,6 +63,40 @@ void amx_command()
...
+		{
+			auto plugin = g_plugins.findPlugin(id);
+
+			if (plugin && plugin->isValid())
+			{
+				print_srvconsole("   Name: %s\n", plugin->getTitle());
+				print_srvconsole("   Version: %s\n", plugin->getVersion());
+				print_srvconsole("   Author: %s\n", plugin->getAuthor());
+				print_srvconsole("   URL: %s\n", plugin->getUrl());
+
```
In-line: I think it's better to not print URL if it's empty

amxmodx/srvcmd.cpp
```diff
@@ -63,6 +63,40 @@ void amx_command()
...
+				auto description = plugin->getDescription(); 
+				if (description[0] != '\0')
```
In-line: If AMXX is compiled in C++17, you can use

`if(auto blabla = ...; blablabla[0])`

amxmodx/srvcmd.cpp
```diff
@@ -63,6 +63,40 @@ void amx_command()
 
 		print_srvconsole("%d plugins, %d running\n", plugins, running);
 	}
+	else if (!strcmp(cmd, "plugin"))
+	{
+		auto id = atoi(CMD_ARGV(2));
```
In-line: Why get id before checking argument count? Are you sure this code is valid if id isn't supplied?

amxmodx/srvcmd.cpp
```diff
@@ -63,6 +63,40 @@ void amx_command()
...
+
+		if (CMD_ARGC() < 3)
+		{
+			print_srvconsole("Usage: amxx plugin < plugin_id >\nFor a list of plugins, use the \"amxx plugins\" command\n");
+		}
+		else
+		{
+			auto plugin = g_plugins.findPlugin(id);
+
+			if (plugin && plugin->isValid())
```
In-line: if with init

amxmodx/srvcmd.cpp
```diff
@@ -244,6 +278,7 @@ void amx_command()
 		print_srvconsole("   version                    - display amxx version info\n");
 		print_srvconsole("   gpl                        - print the license\n");
 		print_srvconsole("   plugins [ criteria ]       - list plugins currently loaded or ones matching given search criteria\n");
+		print_srvconsole("   plugin < plugin_id >       - information about a plugin\n");
```
In-line: Why < .. >?

plugins/include/amxmodx.inc
```diff
@@ -243,10 +243,12 @@ forward client_putinserver(id);
  * @param plugin_name   Name of the plugin
  * @param version       Version of the plugin
  * @param author        Author of the plugin
+ * @param url           URL of the plugin
+ * @param description   Description of plugin
```
In-line: Should be "the plugin"

amxmodx/amxmodx.cpp
```diff
@@ -23,15 +23,30 @@
 extern CFlagManager FlagMan;
 ke::Vector<CAdminData *> DynamicAdmins;
 
+const char *g_sInaccessibleXVars[] =
+{
+	"MaxClients",
+	"MapName",
+	"PluginName",
```
In-line: Why PluginName and not plugin_name? I would prefer to wrap this in macro.

`REGISTER_PLUGIN("blabla, blabalala", "bla")`, it is to noisy and long to declare plugin information like that

amxmodx/amxmodx.cpp
KaidoRen: new arguments are optional (see include)

amxmodx/srvcmd.cpp
KaidoRen: @In-line  ?

amxmodx/srvcmd.cpp
KaidoRen: @In-line what do you think should be here?

amxmodx/amxmodx.cpp
In-line: @KaidoRen This doesn't make it compatible with old plugins

amxmodx/srvcmd.cpp
```diff
@@ -40,7 +40,7 @@ void amx_command()
 				if ((*a).isValid() && !(*a).isPaused())
 					++running;
 
-				print_srvconsole(" [%3d] %-23.22s %-11.10s %-17.16s %-16.15s %-9.8s\n", plugins, (*a).getTitle(), (*a).getVersion(), (*a).getAuthor(), (*a).getName(), (*a).getStatus());
+				print_srvconsole(" [%3d] %-3i %-23.22s %-11.10s %-17.16s %-32.31s %-12.11s %-9.8s\n", plugins, (*a).getId(), (*a).getTitle(), (*a).getVersion(), (*a).getAuthor(), (*a).getUrl(), (*a).getName(), (*a).getStatus());
```
In-line: Why `(*a)` and not `a->`?

amxmodx/srvcmd.cpp
In-line: `plugin [ id ]`

amxmodx/srvcmd.cpp
```diff
@@ -63,6 +63,41 @@ void amx_command()
...
+		{
+			print_srvconsole("Usage: amxx plugin < plugin_id >\nFor a list of plugins, use the \"amxx plugins\" command\n");
+		}
+		else
+		{
+			auto id = atoi(CMD_ARGV(2));
+			
+			if (auto plugin = g_plugins.findPlugin(id); plugin && plugin->isValid())
+			{
+				print_srvconsole("   Name: %s\n", plugin->getTitle());
```
In-line: Title?

amxmodx/srvcmd.cpp
```diff
@@ -63,6 +63,41 @@ void amx_command()
...
+				{
+					print_srvconsole("   URL: %s\n", plugin->getUrl());
+				}
+
+				if (auto description = plugin->getDescription(); description[0])
+				{
+					print_srvconsole("   Description: %s\n", description);
+				}
+
+				print_srvconsole("   Filename: %s\n", plugin->getName());
```
In-line: Name?

amxmodx/srvcmd.cpp
KaidoRen: Yes. Yes. This is not I came up with :)

amxmodx/srvcmd.cpp
KaidoRen: Yes. This is not I came up with :)

amxmodx/amxmodx.cpp
KaidoRen: In issue ( #712 ) already something similar was discussed. I still tend to think that it is more convenient when you know which argument you are changing. And in the macro, as you know, even the name of the argument cannot be specified.

amxmodx/srvcmd.cpp
KaidoRen: Judging by the error, all the same AMX Mod X is not compiled in C ++17. I revert changes.

amxmodx/srvcmd.cpp
KaidoRen: Judging by the error, all the same AMX Mod X is not compiled in C ++17. I revert changes.

amxmodx/srvcmd.cpp
KaidoRen: Because there is no pointer type

amxmodx/amxmodx.cpp
```diff
@@ -1398,6 +1415,15 @@ static cell AMX_NATIVE_CALL register_plugin(AMX *amx, cell *params) /* 3 param *
 	a->setVersion(vers);
 	a->setAuthor(author);
 
+	if (params[arg_count] / sizeof(cell) > arg_author)
+	{
+		const char *url = get_amxstring(amx, params[arg_url], arg_url - 1, i);
+		const char *description = get_amxstring(amx, params[arg_description], arg_description - 1, i);
```
In-line: You can inline this variables

amxmodx/CPlugin.h
```diff
@@ -86,6 +90,8 @@ class CPluginMngr
 		inline void setTitle(const char* n) { title = n; }
 		inline void setAuthor(const char* n) { author =n; }
 		inline void setVersion(const char* n) { version = n; }
+		inline void setUrl(const char* n) { url = n; }
```
In-line: BTW. AMX tells you string length, so you potentially exploit it for faster owned string construction. But I think it isn't very important

amxmodx/amxmodx.cpp
```diff
@@ -23,15 +23,30 @@
...
 {
 	int len;
 	char* sName = get_amxstring(amx, params[1], 0, len);
 	cell ptr;
 
-	if (!strcmp(sName, "MaxClients") || !strcmp(sName, "MapName") || !strcmp(sName, "NULL_STRING") || !strcmp(sName, "NULL_VECTOR"))
+	for (auto var : g_sInaccessibleXVars)
 	{
-		return -1;
+		if (!strcmp(sName, var))
```
In-line: BTW You can add likely/unlikely hints. It's microoptimization, but in any case..

amxmodx/srvcmd.cpp
```diff
@@ -63,6 +63,44 @@ void amx_command()
...
 	}
+	else if (!strcmp(cmd, "plugin"))
+	{
+		if (CMD_ARGC() < 3)
+		{
+			print_srvconsole("Usage: amxx plugin [ id ]\nFor a list of plugins, use the \"amxx plugins\" command\n");
+		}
+		else
+		{
+			auto id = atoi(CMD_ARGV(2));
```
In-line: This can invoke UB. I think it's better to report to user if input is invalid

https://stackoverflow.com/questions/8871711/atoi-how-to-identify-the-difference-between-zero-and-error

amxmodx/srvcmd.cpp
```diff
@@ -63,6 +63,44 @@ void amx_command()
...
+			auto plugin = g_plugins.findPlugin(id);
+
+			if (plugin && plugin->isValid())
+			{
+				print_srvconsole("   Name: %s\n", plugin->getTitle());
+				print_srvconsole("   Version: %s\n", plugin->getVersion());
+				print_srvconsole("   Author: %s\n", plugin->getAuthor());
+
+				auto url = plugin->getDescription();
```
In-line: Oh, copy-paste. How many programms you ruined, how many bugs introduced.. How many hours of debugging I spent unable to stop you..

Please read this line carefully.

amxmodx/amxmodx.cpp
KaidoRen: @In-line what do you mean?

amxmodx/CPlugin.h
KaidoRen: @In-line I did everything the way it was done before me (title, version, author), that not to refactor someone else's code (I have no desire to do this)

amxmodx/amxmodx.cpp
KaidoRen: What's the point of this? :)

amxmodx/srvcmd.cpp
KaidoRen: lol :D

eat1k:
Any news?

KaidoRen:
@eat1k , I am waiting a approve from Inline. I do not know when this day will come. Probably never.

amxmodx/amxmodx.cpp
```diff
@@ -1381,14 +1396,16 @@ static cell AMX_NATIVE_CALL show_menu(AMX *amx, cell *params) /* 3 param */
...
 {
+	enum { arg_count, arg_title, arg_version, arg_author, arg_url, arg_description };
+
 	CPluginMngr::CPlugin* a = g_plugins.findPluginFast(amx);
 	int i;
 
-	char *title = get_amxstring(amx, params[1], 0, i);
-	char *vers = get_amxstring(amx, params[2], 1, i);
-	char *author = get_amxstring(amx, params[3], 2, i);
+	const char *title = get_amxstring(amx, params[arg_title], arg_title - 1, i);
```
In-line: Just auto?

amxmodx/amxmodx.cpp
```diff
@@ -23,15 +23,30 @@
...
+};
+
 static cell AMX_NATIVE_CALL get_xvar_id(AMX *amx, cell *params)
 {
 	int len;
 	char* sName = get_amxstring(amx, params[1], 0, len);
 	cell ptr;
 
-	if (!strcmp(sName, "MaxClients") || !strcmp(sName, "MapName") || !strcmp(sName, "NULL_STRING") || !strcmp(sName, "NULL_VECTOR"))
+	for (auto var : g_sInaccessibleXVars)
```
In-line: Inconsistent naming `var`/`sName`

KaidoRen:
Now we are waiting for the decision from @Arkshine

Arkshine:
Is there a reason you want to close this?

KaidoRen:
@Arkshine A hundred years later you appeared. Perfectly.

amxmodx/amxmodx.cpp
```diff
@@ -23,15 +23,30 @@
...
 {
 	int len;
 	char* sName = get_amxstring(amx, params[1], 0, len);
 	cell ptr;
 
-	if (!strcmp(sName, "MaxClients") || !strcmp(sName, "MapName") || !strcmp(sName, "NULL_STRING") || !strcmp(sName, "NULL_VECTOR"))
+	for (auto name : g_sInaccessibleXVars)
 	{
-		return -1;
+		if (likely(!strcmp(sName, name)))
```
Arkshine: I'm not totally sure it's going to be useful in this context. Also, this is _unlikely_ that forbidden xvars are going to be used here.

Arkshine:
Also, what about `get_plugin()`?

amxmodx/amxmodx.cpp
KaidoRen: @Arkshine and what's stopping me from creating a xvar with a name like PluginName?

Arkshine: I was talking about `likely` only, not the actual check for forbidden identifiers.

Arkshine: @WPMGPRoSToTeMa any thoughts about _likely_/_unlikely_ here?

KaidoRen: @Arkshine this is recommendation from @In-line (_likely_). Okay, I'll clean this up. ))

WPMGPRoSToTeMa: @Arkshine yeah, this is absolutely `unlikely`. But I think `unlikely` should be automatically deduced by the compiler in case of early return.

KaidoRen:
@Arkshine why do need the 12th parameter in the get_plugin function?

Arkshine:
Looks like this parameter would tell you if the plugin is in debug mode. But doc says it's unused. I don't think I have seen a plugin using it.
I think it's safe to replace it. @WPMGPRoSToTeMa what do you think?

WPMGPRoSToTeMa:
@Arkshine as I can see it is for amxmod compatibility. https://github.com/alliedmodders/amxmodx/commit/651c5d9f0104e15cf8e9d5222408d74f4f19d8b8
I think we can replace it.

KaidoRen:
@Arkshine wake up)0

KaidoRen:
@Arkshine dead again? Kk

d3m37r4:
@Arkshine This is right Thing, go back to consider this PR))

In-line:
Hey @KaidoRen contact me on telegram @Inline0

afwn90cj93201nixr2e1re:
no one, exactly no one: 
Arkshine: im not gonna make breakable changes and make some useless updates

Also arkshine: let's make useless column (id) whether [%d] exists, let's change position order and put url in table.

Did u ever `meta list`? Is there any useless stuff in MAIN table? I don't think so.
Url - is useless for many users, and it's takes too much space, it's should be displayed only under `meta info` analog.

Arkshine:
That's quite interesting you have something to say only after the PR is merged.
Helping around opened PRs to eliminate any potential issue would have been more productive than yelling at your convenience, don't you think?

You might not like this feature (personally, I'm not very fond of it), but looking at the thumb-ups on the first post should be a hint people like it. It's okay to ignore it, this is not a mandatory feature. However, any help to improve it is welcomed. Also as reminder, this is a dev branch, anything can change.

If you want to vent your frustration out, you might as well use your energy to motivate others people to participate here. I believe quite a lot of talented coders are active around. Anyone can review and help in a way to make it easier to do the final push. 

I understand you might be upset, but keep in mind, this is a community effort. If no one is willing to participate, it can't be helped -- everyone use their free time as they wish.

afwn90cj93201nixr2e1re:
I didn't see that pr before, that's why, i really don't like sourcemode style.
But this PR is ok.
My post was sarcasmic.

Btw i agree with u, this future is useless, but u merged it only coz first post have thumbs up by useless persons, so, that's not nice.

afwn90cj93201nixr2e1re:
https://github.com/alliedmodders/amxmodx/commit/1bc342a6926a7ea96f7c72482cb1a67744a997be#diff-487c6cf9264c2a33d9839f854ad014ecL1519

Why this was removed?
@Arkshine

Issue #728, opened 2019-05-31
OnConfigsExecuted blocked with server_cmd("/maps")


serfreeman1337:
# Description
Map configs execution and **OnConfigsExecuted** forward can be blocked by any plugin uses 
**server_cmd** with **/maps**.

https://github.com/alliedmodders/amxmodx/blob/975d8778005842efb0a2729f09017c359176f58d/amxmodx/CoreConfig.cpp#L324

AMXX Version: **1.9.0.5241**

# Problematic Code (or Steps to Reproduce)
  ```PAWN
#include <amxmodx>

public plugin_init() {
    log_amx("---> plugin_init")
    
    create_cvar("some_cvar", "some_cvar_value")
    create_cvar("yet_another_cvar", "yet_another_cvar_value")
    
    AutoExecConfig()
    
    server_cmd("/maps /maps /maps")
}

public OnAutoConfigsBuffered() {
    log_amx("---> OnAutoConfigsBuffered")
}

public OnConfigsExecuted() {
    log_amx("---> OnConfigsExecuted")
}
  ```

# Logs
```
L 05/29/2019 - 16:31:33: -------- Mapchange to de_mirage --------
L 05/29/2019 - 16:31:34: [debug.amxx] ---> plugin_init
L 05/29/2019 - 16:31:34: [debug.amxx] ---> OnAutoConfigsBuffered
```

UnrealKaraulov:
@serfreeman1337 fixed in https://github.com/alliedmodders/amxmodx/pull/1022

PR #729, opened 2019-05-31
Fix OnConfigsExecuted blocked with server_cmd("/maps") #728

afwn90cj93201nixr2e1re:
It's just an example.
Fix it.

### DIFF
--- a/amxmodx/CoreConfig.cpp
+++ b/amxmodx/CoreConfig.cpp
@@ -317,12 +317,12 @@ void CoreConfig::CheckLegacyBufferedCommand(char *command)
 		return;
 	}
 
-	if (!m_LegacyMainConfigExecuted && strstr(command, MainConfigFile))
+	if (!m_LegacyMainConfigExecuted && strstr(command, MainConfigPath))
 	{
 		m_LegacyMainConfigExecuted = true;
 	}
 
-	if (!m_LegacyMapConfigsExecuted && strstr(command, MapConfigDir))
+	if (!m_LegacyMapConfigsExecuted && strstr(command, MapConfigPath))
 	{
 		m_LegacyMapConfigsExecuted = true;
 	}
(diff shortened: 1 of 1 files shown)


afwn90cj93201nixr2e1re:
https://github.com/alliedmodders/amxmodx/issues/728

afwn90cj93201nixr2e1re:
@Arkshine 
аркишин
проверяй.

Freeman-AM:
You should name the PR "lol" or "cat ftw" or whatever pass in your mind...

afwn90cj93201nixr2e1re:
PR name just for more attention.

asherkin:
Please familiarise yourself with the [Contributing Guidelines](https://github.com/alliedmodders/amxmodx/blob/master/.github/CONTRIBUTING.md).

This PR can be re-opened if it is given a proper title and description, please don't open another.

afwn90cj93201nixr2e1re:
@asherkin https://github.com/alliedmodders/amxmodx/pull/729#issuecomment-497755786 special for you.

afwn90cj93201nixr2e1re:
@asherkin

dvander:
It looks like this was never merged, and fixes a known issue - can we merge it?

Arkshine:
Fix should be acceptable. It's indeed more reliable with the full path.

I can be wrong, but it's not a perfect fix. `MapConfigPath` is filled when `ExecuteAutoConfigs()` is executed. However, it's executed 6.1 seconds later. This means `strstr(command, MapConfigPath)` will match an empty string until filled and you could trigger likely the same issue with `server_cmd("")` -- but with should fix the issue with `/maps`.

afwn90cj93201nixr2e1re:
It is really bad fix because as i said i did it to get an attention because the whole system where one of reserved path used only for amxmodx - is kinda dumb, so...
You need to think how to reorganize it. I would recommend you just to close eyes on it, because it is pretty useless and it will break something else, as i remember, probably that was the reason why i stopped forcing merging this PR into the master branch. Actually here's issue not only with '/maps' but with something else as well, again, as far as i remember.

UnrealKaraulov:
https://github.com/alliedmodders/amxmodx/pull/1022

PR #751, opened 2019-08-07
Change to Altha-2 to Alpha-3 code space

wopox1337:
https://en.wikipedia.org/wiki/ISO_639
Added ability to use [hbs] or [eng] or [rus] for lang


amxmodx/CLang.cpp
```diff
@@ -145,8 +145,8 @@ CLangMngr::CLang::CLang(const char *lang)
 {
 	m_LookUpTable.clear();
 	m_entries = 0;
-	strncpy(m_LanguageName, lang, 2);
-	m_LanguageName[2] = 0;
+	strncpy(m_LanguageName, lang, LANGNAME_LEN - 1);
```
ClaudiuHKS: Looks like it'd be better to use `strcpy` instead :: https://coliru.stacked-crooked.com/a/7db05714e747cf4c

amxmodx/CLang.h
```diff
@@ -20,6 +20,8 @@
 #define ERR_BADKEY	1	// Lang key not found
 #define ERR_BADLANG 2	// Invalid lang
 
+#define LANGNAME_LEN 4
```
ClaudiuHKS: IMO, would look nicer named `LANGNAME_SIZE` or `LANG_NAME_SIZE`.

Arkshine:
Note that you will need to provide an alternative to `[get_lang](https://www.amxmodx.org/api/lang/get_lang)`.

PR #755, opened 2019-09-06
Update compile.sh

wopox1337:
- Add ability to compile multiple files thru args

### DIFF
--- a/plugins/compile.sh
+++ b/plugins/compile.sh
@@ -1,24 +1,20 @@
-#!/bin/bash
-
-# AMX Mod X
-#
-# by the AMX Mod X Development Team
-#  originally developed by OLO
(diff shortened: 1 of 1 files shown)


wopox1337:
Why old method needs temp.txt output file instead stdout stream?

plugins/compile.sh
```diff
@@ -1,24 +1,20 @@
-#!/bin/bash
-
-# AMX Mod X
-#
-# by the AMX Mod X Development Team
-#  originally developed by OLO
-#
-# This file is part of AMX Mod X.
```
Arkshine: Please don't remove the header.

wopox1337: done

PR #760, opened 2019-09-30
(1.9-dev) Fixes to read_args and read_argv* natives

etojuice:
* Fixed a crash when trying to read unspecified arguments of a command sent by "amxclient_cmd" with read_argv* natives
* Fixed read_args returning command itself instead of nothing when command's (which was sent by amxclient_cmd) arguments were not specified


etojuice:
These changes should be applied to 1.10 as well.

Arkshine:
Thanks for the PR.

About the branch, you should always propose a patch on `master`. This is the up-to-date branch.
Then, once merged, we can cherry-pick the concerned commit to `1.9-dev`, if needed.

etojuice:
Okay.

PR #761, opened 2019-09-30
Fixes to read_args and read_argv* natives

etojuice:
* Fixed crash with read_argv* natives and commands sent with amxclient_cmd native
* Fixed read_args returning argv[0] instead of empty string when both amxclient_cmd's arguments were not specified

### DIFF
--- a/amxmodx/amxmodx.cpp
+++ b/amxmodx/amxmodx.cpp
@@ -2544,7 +2544,7 @@ static cell AMX_NATIVE_CALL read_argv_int(AMX *amx, cell *params) /* 1 param */
 		return 0;
 	}
 
-	const char *value = g_fakecmd.notify ? ((argc >= 1 && argc < 3 && g_fakecmd.argv[argc] != nullptr) ? g_fakecmd.argv[argc]  : "") : CMD_ARGV(argc);
+	const char *value = g_fakecmd.notify ? ((argc >= 1 && argc < 3 && g_fakecmd.argv[argc] != nullptr) ? g_fakecmd.argv[argc] : "") : CMD_ARGV(argc);
 
 	return atoi(value);
 }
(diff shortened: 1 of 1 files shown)


amxmodx/amxmodx.cpp
```diff
@@ -2558,15 +2558,15 @@ static cell AMX_NATIVE_CALL read_argv_float(AMX *amx, cell *params) /* 1 param *
...
-	const char *value = g_fakecmd.notify ? (argc >= 1 && argc < 3 ? g_fakecmd.argv[argc] : "") : CMD_ARGV(argc);
+	const char *value = g_fakecmd.notify ? ((argc >= 1 && argc < 3 && g_fakecmd.argv[argc] != nullptr) ? g_fakecmd.argv[argc]  : "") : CMD_ARGV(argc);
 	float flValue = atof(value);
 
 	return amx_ftoc(flValue);
 }
 
 static cell AMX_NATIVE_CALL read_args(AMX *amx, cell *params) /* 2 param */
 {
-	const char* sValue = g_fakecmd.notify ? (g_fakecmd.argc > 1 ? g_fakecmd.args : g_fakecmd.argv[0]) : CMD_ARGS();
```
Arkshine: Unless I'm wrong, this is the correct behavior. `read_args` should include the command + arguments.
In case of using `amxclient_cmd` with empty arguments, it should return at least the command.

amxmodx/amxmodx.cpp
```diff
@@ -2544,7 +2544,7 @@ static cell AMX_NATIVE_CALL read_argv_int(AMX *amx, cell *params) /* 1 param */
 		return 0;
 	}
 
-	const char *value = g_fakecmd.notify ? (argc >= 1 && argc < 3 ? g_fakecmd.argv[argc] : "") : CMD_ARGV(argc);
+	const char *value = g_fakecmd.notify ? ((argc >= 1 && argc < 3 && g_fakecmd.argv[argc] != nullptr) ? g_fakecmd.argv[argc]  : "") : CMD_ARGV(argc);
```
WPMGPRoSToTeMa: nit: There are two spaces before colon: `g_fakecmd.argv[argc]  :`.

amxmodx/amxmodx.cpp
```diff
@@ -2558,15 +2558,15 @@ static cell AMX_NATIVE_CALL read_argv_float(AMX *amx, cell *params) /* 1 param *
...
```
WPMGPRoSToTeMa: nit: There are two spaces before colon: `g_fakecmd.argv[argc]  :`.

amxmodx/amxmodx.cpp
WPMGPRoSToTeMa: `read_args` should include only arguments. Remember the popular pattern for handling `say`:
https://github.com/alliedmodders/amxmodx/blob/8f88dc5ac8df471a39bc6904edff4e44a8d46564/plugins/adminchat.sma#L92-L94

Arkshine: I checked/tested and it seems you're right. I was pretty sure if no arguments, command would be returned. I vaguely remembered seeing that in ReHLDS, guess not.

PR #772, opened 2019-10-20
Fix internal player's team id returning -1 in a specific but valid situation

### LINKED ISSUE #757: get_user_team returns wrong team ID
Case 1:
When server starts (after it was stopped), and some player join in the server and spawn, if he die and change his team while he is still dead, the return value of get_user_team(id) native will be -1 until he spawn again.

Case 2: If player disconnects from the server, and server changes the map, and then player join again, if he die and change his team the return value of get_user_team(id) native will be 0 until he spawn again.  

NOTE: This happends only when player join the server for the first time since server was started or map was changed. "Case 2" won't happend if player was on the server while map was changed.

I uploaded video so you can understand better what I am talking about. 
Video is showing only case 1.

Video: https://youtu.be/otMWW-S0Yxo

Arkshine:
Fixes #757 
^ More technical info there.

This fixes the situation where a player is dead and changes team. At this point, if the new team name is not yet associated with a valid index, retrieving such index will return -1 until usually the player spawns.

The proposed fix relies on `CBasePlayer::m_iTeam` and is only available to mods where gamedata files are available (CS, CZ, DoD, TFC, OPFor and HLDM).

### DIFF
--- a/amxmodx/CGameConfigs.h
+++ b/amxmodx/CGameConfigs.h
@@ -180,15 +180,15 @@ class CGameConfigManager : public IGameConfigManager
 		member = type.fieldOffset;                                                  \
 	}
 
-#define GET_OFFSET_NO_ERROR(classname, member)										\
-	static int member = -1;															\
-	if (member == -1)																\
+#define GET_OFFSET_NO_ERROR(classname, member)                                      \
+	static int member = -1;                                                         \
+	if (member == -1)                                                               \
 	{                                                                               \
 		TypeDescription type;                                                       \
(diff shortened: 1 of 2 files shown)


WPMGPRoSToTeMa:
I'd like `get_user_team` to return team number from `m_iTeam` directly, so it would be 100% correct in any further case. But unfortunately this may lead to incorrect team name.

Arkshine:
I had this thought we might just use `m_iTeam` but yes, I'm afraid it will change some behavior. For now,  we should keep the fix specific.

Arkshine:
:ship: :pray:

PR #809, opened 2019-11-30
Update pluginmenu.sma

oaus:
Fix for:

> Player 1 is not in game.
[AMXX] Displaying debug trace (plugin "pluginmenu.amxx", version "1.9.0.5263")
[AMXX] Run time error 10: native error (native "menu_display")
[AMXX] [0] pluginmenu.sma::DisplayPluginMenu (line 159)
[AMXX] [1] pluginmenu.sma::CvarMenuSelection (line 474)

### DIFF
--- a/plugins/pluginmenu.sma
+++ b/plugins/pluginmenu.sma
@@ -469,7 +469,7 @@ public CvarMenuSelection(id, menu, item)
 	{
 		menu_destroy(menu);
 		
-		if (ExplicitPlugin[id]==-1)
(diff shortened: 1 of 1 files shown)


OciXCrom:
User connection should be checked at least in the beginning of the `DisplayPluginMenu` function.

Where you put it right now it doesn't make much sense because there are many other if-else checks in the code that shouldn't be executed either if the user is not connected.

OciXCrom:
You should also follow the plugin's coding style, i.e. use curly brackets even when there's only one line of code.

PR #822, opened 2020-01-07
admincmd.sma exploits fix

etojuice:
This patch applies fixes to exploits found in handlers of amx_cvar and amx_map commands


etojuice:
@Arkshine Since the changes were accepted and merged I will reveal the command sequence.

The command sequence to change rcon_password on windows servers pre-patch was:
`amx_map ";rcon_password 1 /../de_dust2"`
Must have access to amx_map command.

PR #823, opened 2020-01-08
adminvote.sma exploit fix

etojuice:
### Restrict having ".." character sequence in amx_votemap command arguments
Fixes exploit on Windows servers that allows executing potentially dangerous console commands


OciXCrom:
Not sure what the actual exploit is, but there's probably no need for a case-insensitive check.

etojuice:
> 
> 
> Not sure what the actual exploit is, but there's probably no need for a case-insensitive check.

Of course.

PR #840, opened 2020-02-27
Add missing is_user_authorized() native

Giferns:
It is implemented in amxmodx.cpp, but missing in [amxmodx.inc](https://github.com/alliedmodders/amxmodx/blob/master/plugins/include/amxmodx.inc)

https://github.com/alliedmodders/amxmodx/blob/d3ef881f31790a71a6070771ae66c812fd4c4a2c/amxmodx/amxmodx.cpp#L941-L949

### DIFF
--- a/plugins/include/amxmodx.inc
+++ b/plugins/include/amxmodx.inc
@@ -922,6 +922,19 @@ native is_user_bot(index);
 native is_user_hltv(index);
 
 /**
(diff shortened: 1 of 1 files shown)


afwn90cj93201nixr2e1re:
from master to master:?)

Arkshine:
Quite funny we notice this function missing after +16 years.

Thanks.

Issue #850, opened 2020-04-03
[DoD]  Set score & set/get deaths natives broken


diamond-optic:
Mod: Day of Defeat
HLDS: 1.1.2.6 (Jul 24 2019 8308)  [and also reHLDS 3.7.0.694 (Mar 22 2020 2171)]
Metamod: 1.21p38
OS: Windows Server 2012

Issue:

**dod_set_user_score** >> Only changes the players score displayed on the scoreboard. It reverts back to the 'legit' score when dying or when gaining another score point (capping a flag etc)

**dod_get_pl_deaths** >> Returns the players score instead of deaths

**dod_set_pl_deaths** >> Really sets the players score, but it changes the displayed deaths on the scoreboard to the input value. It also sets the displayed score on the scoreboard to usually a really high or really low (negative value) number. The displayed values on the scoreboard are also once again reset to the 'legit' values when you die or cap a flag etc

also with the dod_set_pl_deaths is it possible to somehow make it not display the player as dead on the scoreboard? I know the include file makes note that the player will be labeled as dead, but maybe theres a workaround to this?

Plugin I tested with: https://pastebin.com/yQqLc1D4

diamond-optic:
in amxmodx/modules/dod/dodfun/CMisc.h

#define STEAM_PDOFFSET_SCORE		475  + LINUXOFFSET // score
#define STEAM_PDOFFSET_DEATHS		476  + LINUXOFFSET // deaths

in my testing im finding that the offsets are now 476 for score and 477 for deaths

diamond-optic:
in regards to the teamname.. in the NPD.cpp, if i change line 209 from:

char *szTeamName = MF_GetAmxString(amx, params[1], 0, &iLen);
to:
char *szTeamName = MF_GetAmxString(amx, params[2], 0, &iLen);

I can set the team name, and then getting the team name returns that string.. but it doesnt get seem to get set to anything unless you manually set it to something first. Checking through some old logs from a couple years ago this was working correctly at some point. Maybe the game was, and no longer is, setting that value on its own anymore?

Arkshine:
About the offset. I believe the gamedata are correct: https://github.com/alliedmodders/amxmodx/blob/master/gamedata/common.games/entities.games/dod/offsets-cbaseplayer.txt#L941-L957 ;
`m_iObjScore`: 1924 / 4 (size int type) = 476
`m_iDeaths`: 1908 / 4  = 477
[...]

PR #851, opened 2020-04-05
Fix broken DoD Fun module natives

diamond-optic:
- Updates the hardcoded offsets
- Fixes the wrong parameter being used for the team name by dod_set_pl_teamname

Issue Discussion: #850


fysiks1:
When did these offsets change?

PR #862, opened 2020-05-14
Turkish Translation Fixes/Addings

Snakeren:
Fixed lots of grammar, punctuation and orthographic mistakes made by previous translator. The translation may seem wrong for foreigns but i tried my best to make it understandable for Turkish speakers. Also English does not fully fit in Turkish so i changed some translations a bit as being different than default English.

Example:
ADMIN_KICK_1 = ADMIN: kick %s 

I could do that as ADMIN: %s atti but i need to add accusative case to %s and it can change according to last latter of a word. ($nake'yi or Eren'i for example) 

I also could do that as ADMIN: atti %s (previous translate was like this) but this would be wrong and hard to understand by tr speakers at the first time

So i made the change thinking it's the best like that:

"ADMIN, %s kisisini atti." That actually means "ADMIN, kicked the person %s" I also changed : as , because of punctuation rules. If there's a problem, i can change.


seypaa:
i can confirm this changes. But "ban" means in Tr "yasaklanmak" :(

WPMGPRoSToTeMa:
> i can confirm this changes. But "ban" means in Tr "yasaklanmak" :(

Can you submit a PR fixing that?

Snakeren:
> i can confirm this changes. But "ban" means in Tr "yasaklanmak" :(

Yes but since %99 of the players use the word "banlamak" instead of "yasaklamak", i prefered to use like this. I can change if it's a problem.

PR #671, opened 2019-01-29
Add player validation for menu_display

voed:
Invalid player triggered crash because get/set_pdata is unsafe

Issue #645, opened 2018-12-15
CRASH on menu_display


dystopm:
This suddenly happened while changing map. Never happened before (always reading core dumps) on 1.8.2 amxx version.

> amxx version
> AMX Mod X 1.9.0.5234 (http://www.amxmodx.org)
> Authors:
>         David "BAILOPAN" Anderson, Pavol "PM OnoTo" Marko
>         Felix "SniperBeamer" Geyer, Jonny "Got His Gun" Bergstrom
>         Lukasz "SidLuke" Wlasinski, Christian "Basic-Master" Hammacher
>         Borja "faluco" Ferrer, Scott "DS" Ehlert
> Compiled: Nov  3 2018 14:55:29
> Built from: https://github.com/alliedmodders/amxmodx/commit/9eb1f42
> Build ID: 5234:9eb1f42
> Core mode: JIT+ASM32

Something that's more than interesting is this... no other function called, just menu_display, a full backtrace wasn't able to show more. Plugins ported to work on 1.9 were just recompiled and not modified, so I'm giving the credits to the 1.9 version itself since this never happened before (more than 4 years using 1.8.2)

> (gdb) bt full
> #0  0xb2c9cca1 in menu_display(tagAMX*, int*) () from /home/server/cstrike/addons/amxmodx/dlls/amxmodx_mm_i386.so
> No symbol table info available.

At that time, no one connected to the server also. It was 32/32, last log:

> L 12/15/2018 - 17:06:28: Started map "zm_area64_new[zpchile]" (CRC "896559481")

Core dump created at 17:06:32, 4 seconds later, no player connect log. Guess yourself.

If there's anything I should add here, please let me know here.

Arkshine:
Is it something happening all the time? Or just one time suddenly?
Looks like you're using the latest AMXX version so you should have the menu fix.
Without context & log, not going to easy to figure out something.

The only way it could crash at the beginning of `menu_display` is if the menu handle (C pointer) is not null but invalid for some reasons. Though, I don't see how it's even possible here.

If you don't mind I can provide binaries with debug, so the next time the backtrace should hopefully show more info.

PM32:
Just one time suddenly. About context, map started and it just crashed, so I grabbed core dump and figured out this.

But, it's more than weird because menu_display is not used randomly, just on menues called by client commands, and in case of displaying a menu on an invalid player, native would throw an error, which isn't this case.

[...]

PR #863, opened 2020-05-30
Fix menu crashes

afwn90cj93201nixr2e1re:


afwn90cj93201nixr2e1re:
sometimes dropclient and other stuff which was merged into amxmodx not called, and check's 
`if(!pPlayer->ingame)` not working properly, so, we should return everything to clientdrop and remove deprecation from client_disconnect, coz disconnected is really piece of shit.

afwn90cj93201nixr2e1re:
i really don't wanna explain why this bug caused by replacing default meta disconnect with dropclient and other stuff like this, coz it's gonna takes too much and i should explain too much stuff about it, btw u free to investigate it.

afwn90cj93201nixr2e1re:
@Arkshine

afwn90cj93201nixr2e1re:
"in show_menu(tagAMX*,"
and
"in menu_display(tagAMX*,"
to google.

u should reopen this #645 and close with this pr.

afwn90cj93201nixr2e1re:
@Arkshine

Arkshine:
Thanks for the help.

However, if you want to be taken seriously, may I suggest:
* Do not spam – there is an edit button
* Mind your language & stop bashing – this is not helpful
* Explain properly – people are not clairvoyant and this won't help the review process

Thanks for your understanding.

Note: If you want to talk about a specific feature, please open an issue and start the discussion from there respectfully.

afwn90cj93201nixr2e1re:
did u?
изображение


i think no.

afwn90cj93201nixr2e1re:
изображение



изображение



also i wanna say that that is not clear solution, we should return to using meta client disconnect and reset stuff directly in them instead new way.

voed:
Newmenus already fixed in #671. Old-style already has ingame check.

afwn90cj93201nixr2e1re:
Не связаны.

voed:
Then we need to fix `ingame` value because there should be much more unsafe code if ingame behavior is broken

afwn90cj93201nixr2e1re:
> sometimes dropclient and other stuff which was merged into amxmodx not called, and check's
if(!pPlayer->ingame) not working properly, so, we should return everything to clientdrop and remove deprecation from client_disconnect, coz disconnected is really piece of shit.

> this bug caused by replacing default meta disconnect with dropclient and other stuff like this

> we should return to using meta client disconnect and reset stuff directly in them instead new way

 > Then we need to fix ingame value because there should be much more unsafe code if ingame behavior is broken

@Arkshine

etojuice:
Is this problem really cstrike-exclusive?

afwn90cj93201nixr2e1re:
rehlds exclusive.
dropclient exclusive.

etojuice:
> 
> 
> rehlds exclusive.
> dropclient exclusive.

Then you should perform the check outside `if (g_bmod_cstrike)` condition, right?

afwn90cj93201nixr2e1re:
Жуйс. Это псевдофикс. Нет, не за пределами, все это должно было изначально хэндлится ingame check'ом, но из-за этой хрени с dropclient'ом, которая описана 20 раз выше that's ingame check doesn't work as expected.

Why we should check is pdata valid out of scope pdata using? Idk.

afwn90cj93201nixr2e1re:
I don't think that u should merge this fix.
I wanna say that u should remove deprecation from client_disconnect and return reset stuff to client_disconnect.


u can also ask tema for that.

WPMGPRoSToTeMa:
Basically I don't understand what this bug is about. Can you write a single comment with explanations and examples?

Arkshine:
Not sure from where you did this screenshot. It's not even the current code:

image

https://github.com/alliedmodders/amxmodx/blob/master/amxmodx/meta_api.cpp#L906-L935 or did I miss something?

Arkshine:
The check on this PR makes sense, but same as WPMGPRoSToTeMa, I really don't understand what you're saying just above.

This is the original code in this callback. It still does the same thing.

image

Or do I miss something?

afwn90cj93201nixr2e1re:
i dont think that this is exactly same.

@WPMGPRoSToTeMa  should investigate when disconnect not called, coz sometimes also is_user_connected check's doesnt work properly. But if i downgrading to 1.8.2 or 1.8.3 they r work's as expected, there is also too much topics with rehlds + latest amxmodx with that stuff, that's really strange.
My mission is end, i just write something about it to get attention.

afwn90cj93201nixr2e1re:
Continue discussion in https://github.com/alliedmodders/amxmodx/pull/264#issuecomment-658530597, coz seems like i found some problems.

PR #868, opened 2020-06-06
Added small features at compile.sh

CrXane:
See https://forums.alliedmods.net/showthread.php?p=2705106

### DIFF
--- a/plugins/compile.sh
+++ b/plugins/compile.sh
@@ -20,7 +20,7 @@
 # Quit script
 quit_script () {
 	if (( $# != 0 )); then
-		echo $@; fi
+		echo "$@"; fi
 	
 	echo "Exiting..."
 	exit 1
@@ -38,17 +38,20 @@ network_status () {
 
 # Checks if folder exists
 exists () {
-	test -e $1
+	test -e $script_path/$1
 }
 
 # Returns OS' name
 get_os_name () {
+	script_path="$(dirname $(readlink -f $0))"
+	cd $script_path
 	os="$(uname -s)"
 	
 	case "$os" in
-		"Linux"  ) os="linux" ;;
-		"Darwin" ) os="mac" ;;
-	*) esac
+		"Linux"  ) os="linux"   ;;
+		"Darwin" ) os="mac"     ;;
+	*) quit_script "Unsupported OS" ;;
+	esac
 }
 
 # Checks if app is installed
@@ -58,12 +61,12 @@ check_for_app () {
 	else quit_script "Script requires $1 package"; fi
 }
 
-# Checks if variable is number
+# unused ; Checks if variable is number
 is_number () {
 	if ! [[ $1 =~ ^-?[0-9]+([.][0-9]+)?$ ]]; then 
-		quit_script "Invalid number"
-	fi
-}  
+		return 1
+	else return 0; fi
+}
 
 ##################### 
 # ARGUMENT CHECKING #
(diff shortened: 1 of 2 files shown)


plugins/compile.sh
```diff
@@ -7,18 +7,33 @@
...
-for sourcefile in *.sma
-do
-        amxxfile="`echo $sourcefile | sed -e 's/\.sma$/.amxx/'`"
-        echo -n "Compiling $sourcefile ..."
-        ./amxxpc $sourcefile -ocompiled/$amxxfile >> temp.txt
-        echo "done"
-done
+if [ $# != 0 ]
+	then
+		for src in "$@"
```
WPMGPRoSToTeMa, line 21: Should `src` be validated that it ends with `.sma`?

CrXane: Fixed it.

CrXane:
```
./compile.sh --help

  Usage:
--help    (-h) : This menu
--version (-v) : AMX MOD X Compiler version
--latest  (-l) : Latest AMX MOD X version

To compile all source files in the directory
  execute the script without any parameters
To compile specific or multiple source files
  execute the script with parameters set as source files' name
  e.g ./compile.sh admin.sma admincmd.sma adminchat.sma

./compile.sh --version

AMX Mod X Compiler 1.9.0.5270
Copyright (c) 1997-2006 ITB CompuPhase
Copyright (c) 2004-2013 AMX Mod X Team

./compile.sh --latest

Latest AMX MOD X version
1.9 Build: 5270
1.10 Build: 5401
```

CrXane:
Script is functional but doesn't look professional and understandable. I will make edits in upcomming days.

plugins/compile.sh
```diff
@@ -7,18 +7,268 @@
...
+	then
+		quit_script
+	fi
+}
+
+check_for_updates () {
+	check_if_gameserver
+	get_game
+	update_game $id
+	echo -e "\nSucessfully updated"
```
WPMGPRoSToTeMa, line 205: Typo.
```suggestion
	echo -e "\nSuccessfully updated"
```

plugins/compile.sh
```diff
@@ -7,18 +7,268 @@
...
+				;;
+			"--update" | "-u" )
+				check_for_updates
+				;;	
+			*)
+		esac
+		
+		for src in "$@"
+		do
+			if [[ $src == *.sma ]]
```
WPMGPRoSToTeMa, line 252: Should we warn in case of wrong extension?

CrXane: Yes, I thought about it the other day and also adding counter to show how many did plugins did compile and how many didn't, if amxxpc returns any information.

plugins/compile.sh
```diff
@@ -7,18 +7,268 @@
...
+
+get_base () {
+	tar=(which tar)
+	if (( tar == -1 ))
+	then
+		echo "Please install \"tar\" and run the script again"
+		
+		if [ "$os" == "linux" ]
+		then
+			echo "  type: sudo apt-get install tar"
```
Ecklory, line 42: At least one apt based distribution lacks tar?
Having tar is standard for apt and is in depends.

plugins/compile.sh
```diff
@@ -7,18 +7,268 @@
...
+	if (( ver == 1 ))
+	then
+		echo "      configs   / hamdata.ini"
+	fi
+	echo "      data      / all default files"
+	echo "      dlls      / all default files"
+	echo "      modules   / all default files"
+	echo "      plugins   / all default files"
+	echo "      scripting / all default files"
+	echo -en "\nContinue ? [Y/n] : "
```
Ecklory, line 192: Why require a different register?

CrXane: Could you explain what you mean about register?

CrXane:
@Arkshine

plugins/compile.sh
Ecklory: @CrXane 
Yes. This can be useful for debugging, sometimes.
p.s. Write our nicknames on different lines, @WPMGPRoSToTeMa goat )))

CrXane: @Ecklory I thought about adding 1.8.3 & 1.8.2 but breaks it's going to kill the optimization.

Ecklory: @CrXane 
In general, there are several options.
1. hlds_run (restart the server anyway).
2. A few more scripts.
3. Rewrite from bash to some golang.

CrXane: @Ecklory
I've updated add added a couple more features. Check the first post.
I'm using bash because it's preinstalled on Unix systems.

Ecklory: > I'm using bash because it's preinstalled on Unix systems.

The executable file on the golang requires absolutely nothing and works almost everywhere.

CrXane: @Ecklory
Are you sure?
I'm running Ubuntu 20.04 Desktop and I don't have Go preinstalled.
golang-go requries 324 MB of additional storage and gccgo-go 191 MB.

Ecklory: @CrXane 
I'm talking about the fact that hardly anyone will edit the bash script. Since we want functionality, there will only be crutches on bash.

CrXane: @Ecklory
I've made my best to simplify the script and to make it readable on the first glance. The only downside with bash is that the functions have to be created before calling them, which looks like you're looking the code upside-down.

fysiks1:
Seems like this should be two separate pull requests since there are two significant changes that are unrelated to each other.  Also, the pull request should probably have a description that actually mentions what is being done so you can actually get an idea of what happened when looking through a commit log.

CrXane:
Closing PR and opening a new one.

Issue #876, opened 2020-07-07
Refactor dir reading sort


afwn90cj93201nixr2e1re:

afwn90cj93201nixr2e1re:
```
g_plugins.CALMFromFile(path); addons/amxmodx/configs/plugins-additional.ini
g_plugins.CALMFromFile(path); addons/amxmodx/configs/plugins-auw.ini
g_plugins.CALMFromFile(path); addons/amxmodx/configs/plugins-zzzzzzzzzzasdasdasd.ini
g_plugins.CALMFromFile(path); addons/amxmodx/configs/plugins-reasdasd.ini
g_plugins.CALMFromFile(path); addons/amxmodx/configs/plugins-advanced_ultimate_weapons.ini
ParseAndOrAdd(files, ep->d_name); cvars.ini
ParseAndOrAdd(files, ep->d_name); users.ini
ParseAndOrAdd(files, ep->d_name); core.ini
ParseAndOrAdd(files, ep->d_name); plugins.ini
ParseAndOrAdd(files, ep->d_name); ultimate_weapons.ini
ParseAndOrAdd(files, ep->d_name); plugins
ParseAndOrAdd(files, ep->d_name); miscstats.ini
ParseAndOrAdd(files, ep->d_name); clcmds.ini
ParseAndOrAdd(files, ep->d_name); zl
ParseAndOrAdd(files, ep->d_name); .
ParseAndOrAdd(files, ep->d_name); plugins-advanced_ultimate_weapons.ini
ParseAndOrAdd(files, ep->d_name); maps
ParseAndOrAdd(files, ep->d_name); plugins-reasdasd.ini
ParseAndOrAdd(files, ep->d_name); sql.cfg
ParseAndOrAdd(files, ep->d_name); ..
ParseAndOrAdd(files, ep->d_name); cmds.ini
ParseAndOrAdd(files, ep->d_name); nicknames.ini
ParseAndOrAdd(files, ep->d_name); plugins-zzzzzzzzzzasdasdasd.ini
ParseAndOrAdd(files, ep->d_name); cmdaccess.ini
ParseAndOrAdd(files, ep->d_name); custom_configs
ParseAndOrAdd(files, ep->d_name); amxx.cfg
ParseAndOrAdd(files, ep->d_name); asdasd
ParseAndOrAdd(files, ep->d_name); maps.ini
ParseAndOrAdd(files, ep->d_name); custommenuitems.cfg
[...]

PR #877, opened 2020-07-07
Implement plugins loading order (alphabetic)

afwn90cj93201nixr2e1re:
Fixes: #876

### DIFF
--- a/amxmodx/amxmodx.h
+++ b/amxmodx/amxmodx.h
@@ -13,6 +13,7 @@
 #if defined PLATFORM_POSIX
 #include <unistd.h>
 #include <stdlib.h>
+#include <dirent.h>
 #include "sclinux.h"
 #endif
 #include <ctype.h>				//tolower, etc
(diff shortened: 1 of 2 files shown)


afwn90cj93201nixr2e1re:
@Arkshine, @dvander 

Core, enchachment labels and merge it asap...

afwn90cj93201nixr2e1re:
изображение

dvander:
Relying on file names to enforce ordering is not a good idea. It is better to write plugins that are resilient to load order.

afwn90cj93201nixr2e1re:
Em, what r u talking about? And then why we need maps/plugins-mapname.ini lists? Why we need plugins- prefixes at all then?

dvander:
You shouldn't rely on plugin load order. SourceMod loads all *.smx files it sees and does not guarantee load order.

I've been away from AMX Mod X long enough that I shouldn't be making this decision, though, so I leave the review to @Arkshine and others.

afwn90cj93201nixr2e1re:
I know how sourcemod works, it's sometimes designed more better than amxmodx, that's cause why we need order in plugins-* loading.

afwn90cj93201nixr2e1re:
And also we can make some metamodule, if this pr not gonna be accepted, but imo it's should be implemented in core tho.

afwn90cj93201nixr2e1re:
I just made amxmodx module.

Issue #875, opened 2020-07-03
ClientUserInfoChanged_Post for bot's authorization should be refactored


afwn90cj93201nixr2e1re:

afwn90cj93201nixr2e1re:
![изображение](https://user-images.githubusercontent.com/9306996/86494657-9a855f80-bd65-11ea-8dc5-fc5a713fb789.png)



Bot's doesnt have any private data on first iteration, they have pdata only on second call.
	printf("ClientPutInServer_Post %p | %p\n", pEntity, pEntity->pvPrivateData);
![изображение](https://user-images.githubusercontent.com/9306996/86494691-ba1c8800-bd65-11ea-98fc-ec37168079f1.png)


But player's already have initialized pdata on putinserver.


So we should change this line 
https://github.com/alliedmodders/amxmodx/blob/master/amxmodx/meta_api.cpp#L1015-L1016
to


`	} else if (pEntity && pEntity->pvPrivateData && pPlayer->IsBot()) {`

afwn90cj93201nixr2e1re:
@Arkshine

PR #883, opened 2020-07-13
Update meta_api.cpp

afwn90cj93201nixr2e1re:
Fixes: #875

### DIFF
--- a/amxmodx/meta_api.cpp
+++ b/amxmodx/meta_api.cpp
@@ -1012,7 +1012,7 @@ void C_ClientUserInfoChanged_Post(edict_t *pEntity, char *infobuffer)
 	if (pPlayer->ingame)
 	{
 		pPlayer->name =name;			//	Make sure player have name up to date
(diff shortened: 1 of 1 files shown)


afwn90cj93201nixr2e1re:
Also i think that we should'nt call useless stuff on non-valid players, but idk. 

https://github.com/alliedmodders/amxmodx/blob/a534e840f985dae76818cd375fa88155f97e35e1/amxmodx/meta_api.cpp#L1007-L1009
maybe we should move that into condition? 	`if (pPlayer->ingame)`

afwn90cj93201nixr2e1re:
@Arkshine

afwn90cj93201nixr2e1re:
@dvander this one
on first bot connection (client_putinserver) pdata is not created yet, but on second - created, as on default user, so, bot's client_connect == client_putinserver, while player client_connect != client_putinserver, so, if you are gonna try to set/get some offsets on bot - crash.

Issue #886, opened 2020-07-14
Add basic MPROP_SELECTED_COLOR implementation


afwn90cj93201nixr2e1re:

afwn90cj93201nixr2e1re:
rn we should make too much manipulation's to get/set selected item color inside menu, also with using callbacks, so, i just suggest to implement simple color inserting for active item (last selected, item for example, or idk, maybe some new return type on menu callback, whatever, so for preventing menu_get_item info/menu_set_item_name and other useless stuff like that, rn we should fully reimplement logic inside callback for preventing color changes on non accessible items.


```
menu_setprop(g_Menu, MPROP_ACTIVE_ITEM_COLOR, "\r");
```

and then on callback's just return someething like that

```
return ITEM_ENABLED | ITEM_ACTIVE;
return ITEM_DISABLED;
return ITEM_IGNORE | ITEM_ACTIVE; // if there's no access then we should ignore active flag.
```

afwn90cj93201nixr2e1re:
@Arkshine what do u think? Any ideas? Coz i really don't know how implement it properly, also the main problem that menu is hardcoded inside core, not a separate module. So, i can't just write my own implementation.

afwn90cj93201nixr2e1re:
So, we can do it with simple item color overriding, something like

MenuHandler(id, menu, item) {
    menu_setprop(menu, MPROP_ITEM_COLOR, "\r");
}

then we can't undo changes on next item click, so, it's better to have both options MPROP_ITEM_COLOR and MPROP_SELECTED_ITEM_COLOR (for auto switching).

PR #889, opened 2020-07-14
Implement default enabled color, provide info to callback about defau…

afwn90cj93201nixr2e1re:
Fixes: #888
Fixes: #886

### DIFF
--- a/amxmodx/newmenus.cpp
+++ b/amxmodx/newmenus.cpp
@@ -820,7 +820,6 @@ static cell AMX_NATIVE_CALL menu_additem(AMX *amx, cell *params)
 
 	if (params[arg_count] / sizeof(cell) >= arg_enabled_color)
 	{
-		char *enabled_color;
 		pItem->enabled_color = get_amxstring(amx, params[arg_enabled_color], 2, len);
 	}
 
--- a/amxmodx/newmenus.cpp
+++ b/amxmodx/newmenus.cpp
@@ -91,7 +91,7 @@ bool CloseNewMenus(CPlayer *pPlayer)
 	return true;
 }
 
-Menu::Menu(const char *title, AMX *amx, int fid, bool use_ml) : m_Title(title), m_ItemColor("\\r"), 
+Menu::Menu(const char *title, AMX *amx, int fid, bool use_ml) : m_Title(title), m_ItemNumColor("\\r"), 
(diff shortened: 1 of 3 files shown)


afwn90cj93201nixr2e1re:
@Arkshine can you merge it?

amxmodx/newmenus.cpp
```diff
@@ -806,14 +809,20 @@ static cell AMX_NATIVE_CALL menu_additem(AMX *amx, cell *params)
...
+	access = params[arg_access];
 
 	menuitem *pItem = pMenu->AddItem(name, cmd, access);
 
-	pItem->handler = params[5];
+	pItem->handler = params[arg_callback];
+
+	if (params[arg_count] / sizeof(cell) >= arg_enabled_color)
+	{
+		char *enabled_color;
```
Arkshine, line 823: Unused.

afwn90cj93201nixr2e1re:
> enabled_color seems quite confusing. If anything, we are talking about the default_active_color.

im pretty agree with you, at all, i also thought about that, i was betwen default_color and active_color, so, you know why i selected active_color? Coz there's condition, that's the only one reason:
https://github.com/alliedmodders/amxmodx/blob/cac06ff4cc2250dab2fcfeb45feaa5fdc944bca9/amxmodx/newmenus.cpp#L498-L502

as you can see it's under condition.
default_active_color - i don't get point, but i can describe why it's can't be 'default', coz by default it was \w, but if someone gonna change it for example with set_itemcolor and blablabla, then it's not default anymore, u agree? So, that's why i chose that.

amxmodx/newmenus.cpp
afwn90cj93201nixr2e1re: Anything else?

afwn90cj93201nixr2e1re:
I also keep in case that menu item can be '\yA\rblabla', if you're really need renaming just say that again and then  im gonna rename it.

afwn90cj93201nixr2e1re:
Done, anything else? @Arkshine

afwn90cj93201nixr2e1re:
Also If it's possible merge this one, but make them 16 instead 8, just for future using
https://github.com/alliedmodders/amxmodx/pull/656

And then without a twinge of conscience, I will make both default_disabled_color and combine it into one native, now I have not done this just in order not to approach the art (4) limit.

@Arkshine

PR #912, opened 2020-11-19
possible memory leak on deattach

afwn90cj93201nixr2e1re:


afwn90cj93201nixr2e1re:
Some of them already exist in plugins loading, but aren't in deattach/server change.

amxmodx/meta_api.cpp
```diff
@@ -1730,6 +1738,14 @@ C_DLLEXPORT	int	Meta_Detach(PLUG_LOADTIME now, PL_UNLOAD_REASON	reason)
 	g_plugins.clear();
 	g_langMngr.Clear();
 
+	ArrayHandles.clear();
+	TrieHandles.clear();
+	TrieIterHandles.clear();
+	TrieSnapshotHandles.clear();
+	DataPackHandles.clear();
+	TextParsersHandles.clear();
+	GameConfigHandle.clear();
```
dvander, line 1747: Next time these have to be extended, I'd move them into a big helper function to avoid repeating everything.

afwn90cj93201nixr2e1re: I thought about same thing, yes but there's few strange things in code, which r makes duplication more preferably.

PR #918, opened 2020-12-02
Update get_user_userid documentation

etojuice:
It says the native will return 0 if the userid is not available or the client index is invalid, but in fact it returns -1.

https://github.com/alliedmodders/amxmodx/blob/master/amxmodx/amxmodx.cpp#L942
https://github.com/alliedmodders/amxmodx/blob/1.9-dev/amxmodx/amxmodx.cpp#L927

### DIFF
--- a/plugins/include/amxmodx.inc
+++ b/plugins/include/amxmodx.inc
@@ -1278,7 +1278,7 @@ native get_user_authid(index, authid[], len);
  *
  * @param index     Client index
  *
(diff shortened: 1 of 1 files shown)


afwn90cj93201nixr2e1re:
test

test


test

asdasd

asdasd



asdasd

afwn90cj93201nixr2e1re:
> 
> 
> test
> 
> test
> 
> test
> 
> asdasd
> 
> asdasd
> 
> asdasd

PR #928, opened 2021-01-11
Add missing kRenderFx constant

OciXCrom:
I accidentally discovered that there's another kRenderFx value that can be used - 21.

> set_pev(iEnt, pev_renderfx, 21)

Not sure about how it should be named and I'm open to ideas if mine doesn't make sense.
It simply makes the entity black and its texture is barely visible. Changing color or amount doesn't seem to have any effect.
Here's what this looks like when used on an entity.

### DIFF
--- a/plugins/include/amxconst.inc
+++ b/plugins/include/amxconst.inc
@@ -351,7 +351,7 @@ enum
 	kRenderFxExplode,           /* Scale up really big! */
 	kRenderFxGlowShell,         /* Glowing Shell */
 	kRenderFxClampMinScale,     /* Keep this sprite from getting very small (SPRITES only!) */
-	kRenderFxBlackout
+	kRenderFxLightMultiplier,   /* CTM !!!CZERO added to tell the studiorender that the value in iuser2 is a lightmultiplier */
 };
 
 /**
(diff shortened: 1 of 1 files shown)


SmilexGamer:
How did we miss this for years... Lol
Anyways, great catch. I think it could be called "kRenderFxNoTexture".

Arkshine:
https://github.com/ValveSoftware/halflife/blob/c7240b965743a53a29491dd49320c88eecf6257b/common/const.h#L715

OciXCrom:
@SmilexGamer yes, there's text on the back and it's barely visible. Here's a better image - https://i.imgur.com/oVRnJfs.png
@Arkshine thanks. I changed the name to match the one there.

SmilexGamer:
> @SmilexGamer yes, there's text on the back and it's barely visible. Here's a better image - https://i.imgur.com/oVRnJfs.png
> @Arkshine thanks. I changed the name to match the one there.

I think I'm blind, I'm not really seeing it. It doesn't matter, though. Guess it's not a "Blackout" or a "NoTexture" fx, as @Arkshine pointed out.

OciXCrom:
@SmilexGamer it probably depends on your monitor's calibration. Now that I saw the image from my phone, I see what you see.

PR #934, opened 2021-02-22
Use precache_generic() for sounds that are played client-side

OciXCrom:
Since files precached with precache_generic() don't add up to the 512 limit, it makes sense to use this precache method for sounds that are played client-side with "client_cmd spk" in order to save precious resource space.


afwn90cj93201nixr2e1re (approved):

afwn90cj93201nixr2e1re:
Also correct one.
@dvander

PR #939, opened 2021-03-11
Add info about callback function

OciXCrom:
Add missing information about the callback function structure for `register_*cmd`.


Nord1cWarr1or:
@dvander

PR #952, opened 2021-03-21
Add missing set_task_* callback info

OciXCrom:
Add function callback information for `set_task` and `set_task_ex`.


x-Eagle-x:
> if data is passed
should be
> if any data has been passed

PR #956, opened 2021-03-25
Implement client_speak()

OciXCrom:
A simple and very commonly used function.


xLeviNx:
This would be extremely useful

PR #958, opened 2021-04-17
Added natives arraycopy & arraycompare

ArKaNeMaN:
```pawn
/**
 * Copies all elements of array to s destination array.
 *
 * @param src     Array to copy 
 * @param dest    Destination array
 * @param size    Size of array
 *
 * @noreturn
 */
native arraycopy(const any:src[], any:dest[], const size);

/**
 * Compares two arrays.
 *
 * @param what    First array 
 * @param with    Second array
 * @param size    Size of array
 *
 * @return Returns true if arrays are equaly, false otherwise.
 */
native bool:arraycompare(const any:what[], const any:with[], const size);
```


WPMGPRoSToTeMa:
What are use-cases for `arraycopy`? You already can copy one array to another:
```Pawn
new a[3], b[3];
a = b;
```

ArKaNeMaN:
> You already can copy one array to another

Oops... Really...
But `arraycopy` may use when need to copy part of array. But need to add parameter `start` for it...

eyal282:
I don't think old syntax PRs get accepted

PR #966, opened 2021-06-19
Fix for loop index out of bounds error

fysiks1:
Fixes an index out of bounds error when refreshMenus() and subsequently findAdminsWithMenu() is called when there is 32 players in the server. Indexing the playersList array in the for loop condition causes it to try to index with a value of 32 on the very last condition check which is not valid for this array.

### DIFF
--- a/plugins/cstrike/restmenu.sma
+++ b/plugins/cstrike/restmenu.sma
@@ -495,8 +495,10 @@ findAdminsWithMenu(playersList[MAX_PLAYERS], &playersCount, const commandLevel =
 
 	get_players(playersList, playersCount, "ch");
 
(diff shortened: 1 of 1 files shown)


fysiks1:
I imagine that this could be eligible for 1.9-dev.

WPMGPRoSToTeMa (approved):

Arkshine:
Thanks!

fysiks1:
Any chance this can be applied to 1.9-dev since it's a bug fix?

dvander:
Sure, just submit a PR against the 1.9 branch.

WPMGPRoSToTeMa:
> 
> 
> Sure, just submit a PR against the 1.9 branch.

Note that it was already cherry-picked into `1.9-dev`.

PR #972, opened 2021-07-28
Add native `arraycompare`.

ArKaNeMaN:
```pawn
/**
 * Compares two arrays.
 *
 * @param what    First array 
 * @param with    Second array
 * @param size    Size of array
 *
 * @return Returns true if arrays are equaly, false otherwise.
 */
native bool:arraycompare(const any:what[], const any:with[], const size);
```

Example:
```pawn
new const Float:SOME_ARRAY[3] = {1.23, 1.11, 3.21};
new const Float:SOME_ARRAY2[3] = {1.23, 1.11, 78.3};

new Float:SomeArray[3];
SomeArray = SOME_ARRAY;

arraycompare(SOME_ARRAY, SomeArray, 3); // => true
arraycompare(SOME_ARRAY2, SomeArray, 3); // => false
arraycompare(SOME_ARRAY2, SomeArray, 2); // => true
```

### DIFF
--- a/amxmodx/amxmodx.cpp
+++ b/amxmodx/amxmodx.cpp
@@ -4453,6 +4453,20 @@ static cell AMX_NATIVE_CALL arrayset(AMX *amx, cell *params)
 	return 1;
 }
 
+// native bool:arraycompare(const any:what[], const any:with[], const size);
+static cell AMX_NATIVE_CALL arraycompare(AMX *amx, cell *params)
+{
+	cell *what = get_amxaddr(amx, params[1]);
+	cell *with = get_amxaddr(amx, params[2]);
+	int size = params[3];
+
+	while (size--)
+		if (*with++ != *what++)
+			return 0;
+
+	return 1;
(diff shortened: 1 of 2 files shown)


luxxxoor:
this does not look safe if one array is smaller than the count size

i think a better approach would be to create a native that looks like that:
(array1, array1Count, array2, array2Count, maxNumberOfElementsToCompare)

maxNumberOfElementsToCompare should only be valid for natural numbers exept 0
the default value of maxNumberOfElementsToCompare should be 0 or less

if no maxNumberOfElementsToCompare is set and array1Count != array2Count then return false
also.. you should add documentation for all the cases

ArKaNeMaN:
I think that this should be handled by scripters.
What prevents you from passing an invalid value to `arrayNCount`?

luxxxoor:
If you think the scripter should handle that, you should mention it in the documentation

it's not nice when you expect something from the api and get something else

PR #977, opened 2021-08-01
Change default chat time to 0.0 (Fixes #975)

### LINKED ISSUE #975: Non-zero defaut chat time in nextmap.sma breaks functionality for some games
In amxmodx-1.8.2, the default value for `chattime` is 0.0 when the mp_chattime cvar doesn't exist.  However, in 64327b8ddb8db2b2c27d52e09ce03ec0737ccf3a the default value was changed to 10.0.  This breaks map change functionality in games that don't have mp_chattime because we can't know how long the built-in delay is between intermission (event 30) and the normal map change functionality.  For example, in Opposing Forces, the server will change the map before `delayedChange()` is called due to it being delayed for 10 seconds.

fysiks1:
This restores the map change functionality implemented in amxmodx-1.8.2 when the mp_chattime cvar does not exist by setting the default `chattime` value back to 0.0.  This fixes #975.

### DIFF
--- a/plugins/nextmap.sma
+++ b/plugins/nextmap.sma
@@ -103,7 +103,7 @@ public delayedChange(param[])
 public changeMap()
 {
 	new string[32]
-	new Float:chattime = g_mp_chattime ? get_pcvar_float(g_mp_chattime) : 10.0;	// mp_chattime defaults to 10 in other mods
+	new Float:chattime = g_mp_chattime ? get_pcvar_float(g_mp_chattime) : 0.0;
 	
 	if (g_mp_chattime) {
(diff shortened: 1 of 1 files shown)


djearthquake:
g_mp_chattime = get_cvar_pointer("mp_chattime") ? get_cvar_pointer("mp_chattime") : register_cvar("mp_chattime", "20") If 'mp_chattime' is created when nonexistent, like on 'Opposing Force', it resolves this. Changing float times on a null pointer won't create a "delayedChange" task. That is why we hear timeleft count down. We play for 2 more minutes. Next the changelevel is based off mapcycle instead of map we voted for. "delayedChange" I recommend clamping as fail-safe. floatclamp(chattime,1.0,122.0)

Issue #980, opened 2021-08-16
Add set_hudmessage2


RauliTop:
I need to change rgb white color when hud flicks, it is hard-coded here (255, 255, 250): https://github.com/alliedmodders/amxmodx/blob/master/amxmodx/util.cpp#L225

And here: https://github.com/alliedmodders/amxmodx/blob/master/amxmodx/amxmodx.cpp#L547 

So, it's necessary to add a new native to achieve that or add new params after channel at set_hudmessage.

Something like this: https://forums.alliedmods.net/showthread.php?t=69263

But I need SyncHud. That's why I can't use that stock

francoromaniello:
Done in #981

PR #981, opened 2021-08-16
[FEATURE REQUEST] set_hudmessage add missing parameters

francoromaniello:
Related to pull request #628 and #980 add set_hudmessage_ex with the missing params but for branch 1.9-dev

### DIFF
--- a/plugins/include/amxmodx.inc
+++ b/plugins/include/amxmodx.inc
@@ -679,10 +679,53 @@ native disable_logevent(handle);
  * @param fadeintime    Time it takes the message to fully appear (fade-in)
  * @param fadeouttime   Time it takes the message to fully disappear (fade-out)
  * @param channel       Channel to use on the client
+ * @param red           Red 2 component of hudmessage color
+ * @param green         Green 2 component of hudmessage color
+ * @param blue          Blue 2 component of hudmessage color
+ * @param alpha2        Alpha 2 component of hudmessage color		
+ * @param alpha1        Alpha 1 component of hudmessage color
  *
  * @noreturn
  */
-native set_hudmessage(red = 200, green = 100, blue = 0, Float:x = -1.0, Float:y = 0.35, effects = 0, Float:fxtime = 6.0, Float:holdtime = 12.0, Float:fadeintime = 0.1, Float:fadeouttime = 0.2, channel = -1);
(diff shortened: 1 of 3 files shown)


plugins/include/amxmodx.inc
```diff
@@ -679,10 +679,53 @@ native disable_logevent(handle);
...
+ * @param effects       Display effect
+ * @param fxtime        Duration of the effect
+ * @param holdtime      Time the message stays on screen
+ * @param fadeintime    Time it takes the message to fully appear (fade-in)
+ * @param fadeouttime   Time it takes the message to fully disappear (fade-out)
+ * @param channel       Channel to use on the client
+ *
+ * @noreturn
+ */
+native set_hudmessage_ex(red = 200, green = 100, blue = 0, alpha = 0, red2 = 255, green2 = 255, blue2 = 250, alpha2 = 0, Float:x = -1.0, Float:y = 0.35, effects = 0, Float:fxtime = 6.0, Float:holdtime = 12.0, Float:fadeintime = 0.1, Float:fadeouttime = 0.2, channel = -1);
```
dvander, line 723: why is this new native needed?

francoromaniello: set_hudmesage have alpha1, r2, g2, b2, alpha2 hardcoded, this native make the possibility of use all the params available for huds.

image
image

metita: Needed because we will have more flexibility on how hudmessages are displayed, so we can create more cool effects on them. (wondering why that value is hardcoded)

dvander: You can extend the old native. Add parameters like "alpha = 255, color[4] = {0, 0, 0, 0}" (or whatever the defaults should be for backward compatibility).

In the native, you can figure out the number of parameters passed by doing:

    cell num_params = params[0] / sizeof(cell);

Then, if num_params >= 13 (the new count), you can safely extract the additional values.

francoromaniello: Done

dvander:
Thanks! LGTM

RauliTop:
@dvander
This should be added on 1.10-dev too

rtxa:
@dvander Glad they are merging this, but bad at the same time because I did a pull request for the same thing a long time ago but still didn't get any attention. I guess my only criticism is that this repository requires more active mainteners to avoid these situations.

dvander:
Sorry that got missed. Feel free to ping me if you want a review on something. I can't guarantee I'll see or get to everything though, especially for things that I don't know anything about.

rtxa:
@dvander Okay, thanks!

PR #982, opened 2021-08-16
[FIX] alpha2 order in set_hudmessage

francoromaniello:
In my last pull #981 in the color 2 array the order is incorrect, alpha is 3 not 0.
My mistake, sorry about that.


RauliTop:
@dvander

PR #983, opened 2021-08-22
Update parson and fix json_object_set_value

francoromaniello:
The json module needs to update the parson dependency, the repository of parson have commits for memory leaks and more.
I made a git submodule to the repository to make it more easy to update.

EDIT: Also fixed json_object_set_value generate a crash if invalid value is passed 

image
Image related to the crash of json_object_set_value 

EDIT 2: it's required another pull to 1.9-dev ? is the correct branch master to make a pull ?


metita:
@dvander

JSON original repo has been receiving several commits since its release as a AMXX Module but hasn't received any update since then, this worth a watch i guess.

dvander:
We can't include submodules to third-party repos since it exposes various risks. Two options: we either fork it into the AlliedModders team github, and keep the two in sync - or we copy and paste the files in. Since we already did the copy and paste here, and it's quite a small repository, that would be the easier route.

metita:
I think we should just fork it (not sure why it was not like that before?) considering that in the long term it should be more easy to maintain and start working from there.

PR #993, opened 2021-09-19
Added new JSON natives

FEDERICOMB96:
Related to https://github.com/alliedmodders/amxmodx/pull/983


ShootingKing-AM:
Can you give more info? like., What the PR implements ? Can those be done in anyway in a Plugin? If yes, then why this PR etc. etc... ?

UnrealKaraulov:
I think one of reasons... this is more faster than in PAWN

Issue #995, opened 2021-09-24
[Build] checkout-deps.sh pip - python version mismatch


ShootingKing-AM:
# Issue
Recent builds are failing on [Travis-CI](https://travis-ci.org/github/alliedmodders/amxmodx/builds/768378506) and the build fails locally when no <kbd>pip</kbd> is installed on python version *less than* <kbd>Python3.6</kbd>

# Why
Relavant logs from Travis (same are found locally)

```bash
The detected Python installation does not have PIP

Installing the latest version of PIP available (VIA "get-pip.py")
--2021-04-26 06:36:11--  https://bootstrap.pypa.io/get-pip.py
Resolving bootstrap.pypa.io (bootstrap.pypa.io)... 151.101.0.175, 151.101.64.175, 151.101.128.175, ...
Connecting to bootstrap.pypa.io (bootstrap.pypa.io)|151.101.0.175|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 1936711 (1.8M) [text/x-python]
Saving to: ‘./get-pip.py’

./get-pip.py        100%[===================>]   1.85M  --.-KB/s    in 0.08s   
2021-04-26 06:36:12 (23.3 MB/s) - ‘./get-pip.py’ saved [1936711/1936711]

ERROR: This script does not work on Python 2.7 The minimum supported Python version is 3.6. Please use https://bootstrap.pypa.io/pip/2.7/get-pip.py instead.
Installation of PIP has failed

The command "$CHECKOUT_DIR/support/checkout-deps.sh && cd $CHECKOUT_DIR" failed and exited with 1 during .

Your build has been stopped.
```

Issue is because of, https://github.com/alliedmodders/amxmodx/blob/8fefd48af24f18b74948769ec66669b3a6c5f5ec/support/checkout-deps.sh#L95-L131

Relevant code from, https://bootstrap.pypa.io/get-pip.py, will only work for python min version 3.6.
```python
this_python = sys.version_info[:2]
min_version = (3, 6)
if this_python < min_version:
    message_parts = [
        "This script does not work on Python {}.{}".format(*this_python),
        "The minimum supported Python version is {}.{}.".format(*min_version),
        "Please use https://bootstrap.pypa.io/pip/{}.{}/get-pip.py instead.".format(*this_python),
    ]
    print("ERROR: " + " ".join(message_parts))
    sys.exit(1)
```

# Solutions
[...]

PR #1000, opened 2021-09-28
[Build] Migrate to Python3

ShootingKing-AM:
Issue #995 
Min Python requirement would be <kbd>v3.6</kbd>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,9 @@ language: cpp
 sudo: false
 compiler:
   - clang
+install:
+  - pyenv install 3.6.3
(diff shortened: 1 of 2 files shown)


ShootingKing-AM:
Why dosent travis-ci build for all PRs ?

dvander:
We should probably move to Github actions. Peace-Maker was kind enough to do this for SM & MM:S, so there's a bit of a template to go off: https://github.com/alliedmodders/metamod-source/pull/83

ShootingKing-AM:
This is just a quick fix, for now maybe :D

PR #1019, opened 2021-12-01
Added new debug feature: Performance Monitor.

UnrealKaraulov:
**Description:**
_This is new feature for plugins debugging._
Search big performance issues for any plugin that in debug mode. No server slow down!
Output is plugin and function name, and server execution lag in ms. Output target: current amxlog file.


UnrealKaraulov:
Now pull request without duplicated code, and now handled more cases.

UnrealKaraulov:
@Arkshine sorry for ping, but you can comment my small pr?

I think it very useful feature, and can be easy added to AMXMODX without any regressions.

UnrealKaraulov:
**Description:**

Search big performance issues for any plugin that in debug mode. No server slow down. 
Output is plugin and function name, and server execution lag in ms. Output target: current amxlog file.

amxmodx/amx.cpp
```diff
@@ -4173,3 +4172,45 @@ int AMXAPI amx_GetStringOld(char *dest,const cell *source,int use_wchar)
...
+#include <CPlugin.h>
+
+int AMXAPI amx_ExecPerf(AMX* amx, cell* retval, int index)
+{
+    if (amx && retval)
+    {
+        CPluginMngr::CPlugin* perf_Plug = g_plugins.findPluginFast(amx);
+        if (perf_Plug && perf_Plug->isDebug())
+        {
+            char perf_funcname[512];
```
dvander, line 4188: This should be:

    char perf_funcname[sNAMEMAX + 1];

amxmodx/amx.cpp
```diff
@@ -4173,3 +4172,45 @@ int AMXAPI amx_GetStringOld(char *dest,const cell *source,int use_wchar)
...
+            using std::chrono::duration;
+            using std::chrono::milliseconds;
+
+            auto t1 = high_resolution_clock::now();
+            int err = amx_Exec(amx, retval, index);
+            auto t2 = high_resolution_clock::now();
+
+            auto ms_int = duration_cast<milliseconds>(t2 - t1);
+            auto ms_integer = (int)ms_int.count();
+            if (ms_integer > 1)
```
dvander, line 4208: Please make this configurable, via cvar or command-line or something. 1ms is probably okay since that's quite a big chunk of the frame budget, but it'd be best to have the server operator have some control over it (and maybe express the units in microseconds instead?)

dvander (commented):
This is a good idea and I'd be happy to take this, with a few minor changes.

amxmodx/amx.cpp
UnrealKaraulov: Thanks

amxmodx/amx.cpp
```diff
@@ -4173,3 +4172,45 @@ int AMXAPI amx_GetStringOld(char *dest,const cell *source,int use_wchar)
...
 }
+
+
+#include <chrono>
+#include <amxmodx.h>
+#include <CPlugin.h>
+
+int AMXAPI amx_ExecPerf(AMX* amx, cell* retval, int index)
+{
+    if (amx && retval)
```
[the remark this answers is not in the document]
UnrealKaraulov: Okay!

amxmodx/amx.cpp
UnrealKaraulov: At current time bigger than 1 ms (2+ms) it very big time for one function, yes I think need make it configurable

amxmodx/amx.cpp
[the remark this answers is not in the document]
UnrealKaraulov: Fixed

amxmodx/amx.cpp
UnrealKaraulov: Fixed

amxmodx/amx.cpp
```diff
@@ -4185,9 +4185,9 @@ int AMXAPI amx_ExecPerf(AMX* amx, cell* retval, int index)
         CPluginMngr::CPlugin* perf_Plug = g_plugins.findPluginFast(amx);
         if (perf_Plug && perf_Plug->isDebug())
         {
-            char perf_funcname[512];
+            char perf_funcname[sNAMEMAX + 1];
             amx_GetPublic(perf_Plug->getAMX(), index, perf_funcname);
-            if (strlen(perf_funcname) == 0)
+            if (perf_funcname[0] == '\0')
```
dvander: Sorry one more thing here, I just realized - amx_GetPublic can return an error, in which case the string won't be initialized. So you could fix this either by:

    char perf_funcname[sNAMEMAX + 1] = {};

Or:

    int err = amx_GetPublic(perf_Plug->getAMX(), index, perf_funcname);
    if (err != AMX_ERR_NONE)
        snprintf(perf_funcname, sizeof(perf_funcname), "Unknown_ID%d", index);

amxmodx/amx.cpp
UnrealKaraulov: @dvander why better, add new cvar to amxx.cfg file?

dvander: Pretty much any method would work, I think. The exact method doesn't matter as long as it's easy for server operators to change.

amxmodx/amx.cpp
UnrealKaraulov: Yes, really. Thanks.

UnrealKaraulov: I think now it is fixed?

amxmodx/amx.cpp
UnrealKaraulov: I can do it from PC, at current time I used phone for browsing GitHub

UnrealKaraulov: @dvander you can review new update ? 

https://github.com/alliedmodders/amxmodx/pull/1019/commits/5cb6fa0374289ae6bbac75348a456872b127151d

Now added cvar.

UnrealKaraulov:
* Added cvar amx_perflog (0 for disable this feature)
* Enabled for every plugin if amx_debug is 2. (Even if plugin without debug support)

https://github.com/alliedmodders/amxmodx/pull/1019/commits/5cb6fa0374289ae6bbac75348a456872b127151d

UnrealKaraulov:
@dvander, maybe something is missing? ( any suggestions ? )

amxmodx/meta_api.cpp
```diff
@@ -135,9 +135,14 @@ cvar_t init_amxmodx_debug = {"amx_debug", "1", FCVAR_SPONLY};
 cvar_t init_amxmodx_mldebug = {"amx_mldebug", "", FCVAR_SPONLY};
 cvar_t init_amxmodx_language = {"amx_language", "en", FCVAR_SERVER};
 cvar_t init_amxmodx_cl_langs = {"amx_client_languages", "1", FCVAR_SERVER};
+cvar_t init_amxmodx_perflog = { "amx_perflog", "1.0", FCVAR_SPONLY };
```
ShootingKing-AM, line 138: `amx_preflog_ms` would be more descriptive, imo.

ShootingKing-AM:
On which platforms did you test this ?

amxmodx/amx.cpp
```diff
@@ -4173,3 +4173,43 @@ int AMXAPI amx_GetStringOld(char *dest,const cell *source,int use_wchar)
...
+        using std::chrono::duration_cast;
+        using std::chrono::duration;
+        using std::chrono::microseconds;
+
+        auto t1 = high_resolution_clock::now();
+        int err = amx_Exec(amx, retval, index);
+        auto t2 = high_resolution_clock::now();
+
+        auto ms_int = duration_cast<microseconds>(t2 - t1);
+        auto ms_float = (float)(ms_int.count() / 1000.0f);
```
ShootingKing-AM, line 4207: Why dont you use `static` variables here ? amx_Exec will be called many a times. 

`t2` and `ms_int` variables are only being used once, may be we can skip its mem allocation and do the following,
```c++
 auto ms_float = (float)(( duration_cast<microseconds>( high_resolution_clock::now() - t1) ).count() / 1000.0f);
```

UnrealKaraulov: Why you think what static faster than local, this is not PAWN language, but okay I can do it.

ShootingKing-AM: Again, dont do it just because of my suggestion. If you think local vars are better, just state so, no issues.

amxmodx/amx.cpp
```diff
@@ -4173,3 +4173,43 @@ int AMXAPI amx_GetStringOld(char *dest,const cell *source,int use_wchar)
...
+        const char* perf_plugname = perf_Plug->getName();
+        if (!perf_plugname || perf_plugname[0] == '\0')
+            perf_plugname = "Unknown_plugin";
+
+        using std::chrono::high_resolution_clock;
+        using std::chrono::duration_cast;
+        using std::chrono::duration;
+        using std::chrono::microseconds;
+
+        auto t1 = high_resolution_clock::now();
```
WPMGPRoSToTeMa, line 4202: You should use `steady_clock` instead. `high_resolution_clock` may be a subject to time adjustments.
See notes in https://en.cppreference.com/w/cpp/chrono/high_resolution_clock.

amxmodx/amx.cpp
```diff
@@ -4173,3 +4173,43 @@ int AMXAPI amx_GetStringOld(char *dest,const cell *source,int use_wchar)
   dest[len]='\0';     /* store terminator */
   return AMX_ERR_NONE;
 }
+
+
+#include <chrono>
```
WPMGPRoSToTeMa, line 4178: Is it allowed to use STL instead of AMTL in amxmodx?

amxmodx/amx.cpp
WPMGPRoSToTeMa: As I remember `amx_Exec` may be executed recursively, so you can't use `static` variables here.

amxmodx/meta_api.cpp
WPMGPRoSToTeMa: @ShootingKing-AM if we end up with microseconds then it should be `amx_perflog_us`.

amxmodx/modules.cpp
```diff
@@ -187,7 +187,7 @@ int load_amxscript_internal(AMX *amx, void **program, const char *filename, char
 	bool will_be_debugged = false;
 	tagAMX_DBG *pDbg = NULL;
 
-	if ((int)CVAR_GET_FLOAT("amx_debug") >= 2 || debug)
+	if (amxmodx_debug->value >= 2.0f || debug)
```
WPMGPRoSToTeMa, line 190: Could you please elaborate why you changed this?

WPMGPRoSToTeMa (changes requested):

amxmodx/modules.cpp
UnrealKaraulov: @WPMGPRoSToTeMa ->value is floating point , and it executed faster? ( I created cvar for using in amx_ExecPerf and this cvar can be used here, directly without execute CVAR_GET_FLOAT )

amxmodx/amx.cpp
UnrealKaraulov: I don't know

amxmodx/meta_api.cpp
UnrealKaraulov: Yes I can do it

amxmodx/amx.cpp
UnrealKaraulov: I can use another clock?

amxmodx/amx.cpp
UnrealKaraulov: Yes I see. Thanks

amxmodx/amx.cpp
UnrealKaraulov: In this case I can't use static variables even if this faster

amxmodx/amx.cpp
UnrealKaraulov: @WPMGPRoSToTeMa  I can use "clock" function from https://www.cplusplus.com/reference/ctime/clock/ in this case ? Or this function has small resolution ?

UnrealKaraulov: @WPMGPRoSToTeMa  

I use chrono because I think it has better precision, but if default function 'clock()' has normal precision, i can use it.

WPMGPRoSToTeMa: I'm fine with using `chrono`, but let's wait for @dvander's opinion on this. As I remember `clock` has milliseconds precision on Windows, so it's not applicable here.

amxmodx/amx.cpp
UnrealKaraulov: @WPMGPRoSToTeMa maybe you know another better way?

amxmodx/meta_api.cpp
UnrealKaraulov: @WPMGPRoSToTeMa but this is milliseconds, with possible to use with microseconds precision.
I think amx_preflog_ms better than us?

amxmodx/amx.cpp
UnrealKaraulov: @WPMGPRoSToTeMa okay, but at this time I update it to steady_clock https://github.com/alliedmodders/amxmodx/pull/1019/commits/58efc74c9e3e18ac123a24308394e5757bbfb73c

amxmodx/amx.cpp
UnrealKaraulov: Yes, it can be recursive, calling one function from another.

amxmodx/meta_api.cpp
WPMGPRoSToTeMa: @UnrealKaraulov ah, it's float, I see. Yes, then of course we should name it `amx_perflog_ms`.

amxmodx/amx.cpp
UnrealKaraulov: Wait for dvander comment

UnrealKaraulov:
> On which platforms did you test this ?

On Windows but I think <chrono> is portabled.  (and no alternatives at this time)

ShootingKing-AM:
> > On which platforms did you test this ?
> 
> On Windows but I think < chrono > is portabled and can be used at any other OS.

Authors are usually tasked with testing their code cross-platform. Maybe you can or someone can test the code on linux. Assuming it will work is not really a good practice, imo.

@dvander can allow the this workflow to run now so that we can check if there are some compile time errors cross-platform and cross-toolkit.

dvander:
Stuff in std:: is by definition portable.

amxmodx/modules.cpp
dvander: Bypassing the string lookup is fine, but it should be `(int)amxmodx_debug->value == 2`. The value is an integer so using fuzzy comparisons is confusing.

dvander (commented):
Looks good, just a few more minor comments.

amxmodx/modules.cpp
UnrealKaraulov: Okay I do it. 
This is can't decrease performance?(convert from float to int and CMP two ints, or just CMP two float values)

amxmodx/amx.cpp
```diff
@@ -539,7 +539,7 @@ static int amx_BrowseRelocate(AMX *amx)
 
   amx->sysreq_d=0;      /* preset */
   #if (defined __GNUC__ || defined ASM32 || defined JIT) 
-    amx_Exec(amx, (cell*)(void*)&opcode_list, 0);
+    amx_ExecPerf(amx, (cell*)(void*)&opcode_list, 0);
```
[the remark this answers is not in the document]
UnrealKaraulov: Thanks, removed it.

amxmodx/amx.cpp
```diff
@@ -4173,3 +4173,42 @@ int AMXAPI amx_GetStringOld(char *dest,const cell *source,int use_wchar)
   dest[len]='\0';     /* store terminator */
   return AMX_ERR_NONE;
 }
+
+
+#include <chrono>
+#include <amxmodx.h>
+#include <CPlugin.h>
```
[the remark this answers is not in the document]
UnrealKaraulov: Yes but then I need use `#undef min max` because it conflict with AMXMODX min/max definitions.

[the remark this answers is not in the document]
dvander: That's fine. #define min/max is very old school, it's an old/bad practice.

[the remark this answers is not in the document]
UnrealKaraulov: Now moved to top.

dvander:
Thanks, merged.

UnrealKaraulov:
@dvander, thanks

amxmodx/amx.cpp
UnrealKaraulov: @WPMGPRoSToTeMa yes possible in future need add clocks to AMTL and then use it in AMXMODX. I did not attach importance to your message until I ran with this issue: https://github.com/alliedmodders/amxmodx/pull/1026

justgo97:
How to disable performance monitor logs?

UnrealKaraulov:
> How to disable performance monitor logs?

impossible without this pull request : https://github.com/alliedmodders/amxmodx/pull/1028

PR #1021, opened 2021-12-03
Fix some misprints and mistakes in code, part 2

UnrealKaraulov:
Fix many useless break\s in code. Fix buffer overflow in non cstrike mods (MAX_WEAPONS misprints) and small other fixes.

### DIFF
--- a/compiler/libpc300/sc6.c
+++ b/compiler/libpc300/sc6.c
@@ -70,7 +70,7 @@ static ucell hex2long(char *s,char **n)
     s++;
   } /* if */
 
-  assert((*s>='0' && *s<='9') || (*s>='a' && *s<='f') || (*s>='a' && *s<='f'));
+  assert((*s>='0' && *s<='9') || (*s>='a' && *s<='f') || (*s>='A' && *s<='F'));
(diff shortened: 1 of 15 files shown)


UnrealKaraulov:
@dvander but I think it can be broke dod_weapon_type function, x > 31 . 

https://github.com/alliedmodders/amxmodx/blob/383507daaf67939c9782366513cc5fb351849fb2/modules/dod/dodx/NBase.cpp#L268

UnrealKaraulov:
@dvander possible only in this function need use MAX_WEAPONS instead of DODMAX_WEAPONS. (because 1<<31 is max possible value)

dvander:
It seems like the intent was `if ((weaponsbit & (1 << x)) != 0)`

UnrealKaraulov:
@dvander in any case, 31 bit is max, see http://cpp.sh/4fucs

dvander:
This code is like 20 years old. You're not going to find anyone who still remembers it.

UnrealKaraulov:
I  don't know it, maybe :) 

Okay then just need wait for someone report about issues in this function :)

PR #1022, opened 2021-12-03
Fix OnConfigsExecuted issue. (restmenu.sma and adminhelp.sma is broken)

UnrealKaraulov:
OnConfigsExecuted not executed and broke all plugins where used this forward. Even default like **restmenu.sma** and **adminhelp.sma**!

### DIFF
--- a/amxmodx/CoreConfig.cpp
+++ b/amxmodx/CoreConfig.cpp
@@ -316,14 +316,10 @@ void CoreConfig::CheckLegacyBufferedCommand(char *command)
 		return;
 	}
 
-	if (!m_LegacyMainConfigExecuted && strstr(command, MainConfigFile))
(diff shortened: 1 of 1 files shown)


UnrealKaraulov:
@dvander 
**Description:**
If any config matches the conditions, then forward OnConfigsExecuted will never be executed and this will lead to errors in all plugins where OnConfigsExecuted is used.

UnrealKaraulov:
For example execute function from any plugin at map start:

`server_cmd("amxx.cfg")` **and forward  OnConfigsExecuted will never be executed**

UnrealKaraulov:
@dvander It doesn't matter which command, the main thing is that **amxx.cfg** or **/maps** are present there.

UnrealKaraulov:
https://github.com/alliedmodders/amxmodx/issues/728 https://github.com/alliedmodders/amxmodx/pull/729

UnrealKaraulov:
A more visual representation can be seen here: https://github.com/alliedmodders/amxmodx/issues/728

Issue #1025, opened 2021-12-06
Debug feature: Performance Monitor Issue


UnrealKaraulov:
https://github.com/alliedmodders/amxmodx/commit/8ee3da24cb811ca5c425bc4aec954eee22ca371d

If someone get problems with using this feature or have any suggestions, please use this issue for it.

UnrealKaraulov:
Closed

UnrealKaraulov:
@dvander I see on older linux os next message: 

`L 12/06/2021 - 08:33:46: [META] ERROR: dll: Failed query plugin '<amxmodx_mm_i386.so>'; Couldn't open file '/home/server/game/cstrike/addons/amxmodx/dlls/amxmodx_mm_i386.so': /home/server/game/cstrike/addons/amxmodx/dlls/amxmodx_mm_i386.so: undefined symbol: _ZNSt6chrono3_V212steady_clock3nowEv`

Possible need add something to buildbot ?

dvander:
It's missing linkage to libstdc++.

We find the static library here: https://cs.alliedmods.net/amxmodx/source/AMBuildScript#316
And currently only amxxpc uses it here: https://cs.alliedmods.net/amxmodx/source/compiler/amxxpc/AMBuilder#14

amxmodx_i486.so needs a similar link step, I think.

UnrealKaraulov:
@dvander I do build manual from https://wiki.alliedmods.net/Building_AMX_Mod_X

But get next errors: 
![VirtualBox_UBUNTU LATEST_06_12_2021_10_05_50](https://user-images.githubusercontent.com/88730285/144802411-11f58078-f9fb-4250-9891-380bc88c1de6.png)

UnrealKaraulov:
All **is** need rename to **==** ?

dvander:
Yeah, newer versions of Python will complain about this. It's a bug in the script, we should have used "==" for string comparison.

UnrealKaraulov:
@dvander i tried to add std support in https://github.com/alliedmodders/amxmodx/pull/1026

dvander:
I merged the build fix. I think it should work. You can test by using "ldd -d -r <.so file>" on the command-line, and it should tell you if any symbols are missing.

UnrealKaraulov:
> I merged the build fix. I think it should work. You can test by using "ldd -d -r <.so file>" on the command-line, and it should tell you if any symbols are missing.

It can be added to GitHub actions?

[...]

PR #1026, opened 2021-12-06
Allow to use std:: in amxmodx to fix 'Performance monitor' from previous PR

UnrealKaraulov:
@dvander it can solve this https://github.com/alliedmodders/amxmodx/issues/1025#issuecomment-986457300 issue ?


UnrealKaraulov:
@dvander 

Tested on ubuntu-latest and it works! 

I don't know about older os versions.

PR #1027, opened 2021-12-07
Just print error, but don't crash plugins.

UnrealKaraulov:
**Please, if no debug info, don't crash plugins, just write error and skip debug info.**

Description:

Many plugins like extended ban systems (fresh bans, etc) or antihacks (aim detectors, etc) compiled without debug for protect source code from hackers. 
At this moment, if activate `amx_debug 2` all plugins without debug info stop working. 

After the merge, we will just write information about the missing of debug option and the plugin continues work without any issues. (Also it can be useful for debug this plugins with Performance Monitor even without debug info)


dvander:
I'm confused, what crash is this fixing?

UnrealKaraulov:
> I'm confused, what crash is this fixing?

Sorry for bad english, this prevent plugin stopped if no debug info with amx_debug 2

UnrealKaraulov:
@dvander now it just print in log 'no debug info, skip it' without error (plugin still works instead of paused)
VirtualBox_UBUNTU LATEST_07_12_2021_07_59_17

amxmodx/modules.cpp
```diff
@@ -215,8 +215,13 @@ int load_amxscript_internal(AMX *amx, void **program, const char *filename, char
...
 			amx->flags |= AMX_FLAG_DEBUG;
 		} else {
-			ke::SafeStrcpy(error, maxLength, "Plugin not compiled with debug option");
-			return (amx->error = AMX_ERR_INIT);
+			const char * pluginname = filename;
+			// Easy extract pluginname from file path.
+			while(strstr(pluginname,"/")) pluginname = strstr(pluginname,"/") + 1;
+			while(strstr(pluginname,"\\")) pluginname = strstr(pluginname,"\\") + 1;
+			// Just skip debug info if not found.
+			AMXXLOG_Log("[AMXX] Plugin not compiled with debug option. Ignore debug. ( plugin \"%s\" )", pluginname);
```
dvander, line 223: Try to follow local C++ style, "*" near name and spaces between keywords/tokens.

			const char *pluginname = filename;
			// Easy extract pluginname from file path.
			while (strstr(pluginname, "/"))
				pluginname = strstr(pluginname, "/") + 1;
			while (strstr(pluginname, "\\"))
				pluginname = strstr(pluginname, "\\") + 1;

And remove extra spaces in logging:

			AMXXLOG_Log("[AMXX] Plugin not compiled with debug option. Not debugging plugin \"%s\"", pluginname);

UnrealKaraulov:
@dvander 

**Before this fix.**
If we enabled amx_debug 2, all plugins compiled without debug option - stopped with error.
**After this fix.**
If we enabled amx_debug 2, all plugins compiled without debug option - continue works, but without debug flag.


I hope now I wrote it easier and clearer :)

amxmodx/modules.cpp
UnrealKaraulov: Thanks, fixed in https://github.com/alliedmodders/amxmodx/pull/1027/commits/2b4111ebfd7a06edbec7437d0f2e90560870605f

UnrealKaraulov: @dvander also added `#ifdef ` check in https://github.com/alliedmodders/amxmodx/pull/1027/commits/d4f896960179c3f990873a030f40b7f6babb7c38

UnrealKaraulov:
@dvander i think latest commit in this PR is really latest. 

Now if user forced set debug in plugins.ini - default behavior.
If user set amx_debug 2, then skip debug if no debug info found.

UnrealKaraulov:
@dvander also I see problem, amx_debug cvar loaded not a first map.

UnrealKaraulov:
Any questions?

wopox1337:
@dvander @Arkshine ping

UnrealKaraulov:
@dvander @Arkshine америкосы

PR #1035, opened 2022-01-20
Update string.inc

rust2:
fix some spacing


rust2:
@dvander could u check?

PR #1043, opened 2022-04-05
Update newmenus.inc

rust2:


dvander:
But who doesn't love menu menus?

rust2:
yes

Issue #855, opened 2020-04-22
Bug in nextmap plugin: mp_chattime gets increased on map change


rtxa:
## Issue
There's an issue with the nextmap plugin, when intermission is triggered and you do a map change, e.g. using amx_map, mp_chattime gets increased by 2.

When this is done several times, the mp_chattime will have reach a high value, making the time for a map change very long.

The cause of this is found in the function event `changeMap()`, most specifically in this line:
https://github.com/alliedmodders/amxmodx/blob/2559fcf00aecb19fb6e6f8de846ae37cbcd6d572/plugins/nextmap.sma#L112

The function `delayedChange()` never gets executed because it's called by a task with the value of `[mp_chattime original value + two seconds]`, and if you do the map change before of that, the bug is triggered.

https://github.com/alliedmodders/amxmodx/blob/2559fcf00aecb19fb6e6f8de846ae37cbcd6d572/plugins/nextmap.sma#L95-L101

## Solution
One solution that has come to my mind is to the reset mp_chattime to his original value in plugin_end()

```pawn
new g_changeMapCalled;

public delayedChange(param[])
{
	engine_changelevel(param)
}

public plugin_end()
{
	if (g_mp_chattime && g_changeMapCalled) {
		set_pcvar_float(g_mp_chattime, get_pcvar_float(g_mp_chattime) - 2.0)
	}
}

public changeMap()
{
	new string[32]
	new Float:chattime = g_mp_chattime ? get_pcvar_float(g_mp_chattime) : 10.0;	// mp_chattime defaults to 10 in other mods
	
	if (g_mp_chattime) {
		set_pcvar_float(g_mp_chattime, chattime + 2.0)		// make sure mp_chattime is long
		g_changeMapCalled = true;
	}
	new len = getNextMapName(string, charsmax(string)) + 1
	set_task(chattime, "delayedChange", 0, string, len)	// change with 1.5 sec. delay
}
```

This way, you will make sure the mp_chattime is reseted always to his original value, making it more reliable.

Please, tell me what you think, I tested the solution for about a week and I didn't see the issue coming back since then.
If you have a better solution or an improvement feel free to post it, thanks.

stefanofabi:
executing the mp_chattime command with rcon returns the value incremented by 2?

rtxa:
> 
> 
> executing the mp_chattime command with rcon returns the value incremented by 2?

[...]

PR #1044, opened 2022-04-09
nextmap.sma - Fix mp_chattime getting increased on quick map change

rtxa:
Related issue: #855 

# Introduction

This should fix an ancient bug where the server takes too long to change the map after it's has finished (scoreboard screen is displayed).

The scoreboard screen will stay for too long, becoming annoying to the players to the point of even leaving the server.

# When this happens

Basically this happens when you change the map in the middle of the intermission screen. The cvar `mp_chattime` which controls how long the intermission screen stays, gets increased by 2 seconds. If you do this several times, the intermission screen time would start to get bigger, to the point of becoming noticeable to the players.

This most likely happens when the player doesn't like the next map and decides in the middle of the intermission screen to change it to a one he likes.

# Steps to reproduce it

1. Load some map.
2. Check out the value of `mp_chattime` first. Default value should be 10 seconds.
3. Wait the map to finish. You can speed up the process by setting `mp_timelimit` to `0.1` which means the map will finish in 10 seconds.
4. The scoreboard screen is being displayed. Change to another map with `amx_map` before the server changes it.
5. Check again `mp_chattime` value. This time you'll see that it has been increased by 2 seconds.
 
# How I fix it

The next-map plugin needs to override the next-map system from Half-Life to work. To achieve this, it will look first how long the intermission screen stays by reading the value of `mp_chattime`. After that, it will add two extra seconds to the `mp_chattime` to have enough time to change the map before the original next-map system does it.

Basically a task is being called when the intermission screen is on to return to the original value of `mp_chattime`. But the code doesn't take into account a player changing the map in the middle of the intermission screen. The `mp_chattime` original value never gets restored.

Instead of relying in that someone will not change the map before the task executes, this fix will restore the original value in `plugin_end()`, making it safe from the player's actions. This forward is called just before the map ends. I have tested this fix for a long time and I didn't find any issues so far.

# The End

Well, with everything said, I hope to see this get merged. Let me know if you need me to make a further explanation on something. Thanks for reading.

### DIFF
--- a/plugins/nextmap.sma
+++ b/plugins/nextmap.sma
@@ -22,6 +22,7 @@ new g_nextMap[32]
 new g_mapCycle[32]
 new g_pos
 new g_currentMap[32]
+new g_changeMapCalled;
 
 // pcvars
 new g_mp_friendlyfire, g_mp_chattime
@@ -94,10 +95,14 @@ public sayFFStatus()
 
 public delayedChange(param[])
 {
-	if (g_mp_chattime) {
+	engine_changelevel(param)
+}
+
+public plugin_end()
+{
+	if (g_mp_chattime && g_changeMapCalled) {
 		set_pcvar_float(g_mp_chattime, get_pcvar_float(g_mp_chattime) - 2.0)
 	}
-	engine_changelevel(param)
 }
 
 public changeMap()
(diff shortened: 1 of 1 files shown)


xLeviNx:
Wasn't the two solutions that would work listed in the thread you tagged? #855

rtxa:
> Wasn't the two solutions that would work listed in the thread you tagged? #855
Yeah, the solution I suggested is there, but the issue has a lot of verbosity, so prefer to read the solution in here instead. Also you can see changes easily which are just a couple lines.

rtxa:
Hi @dvander, Would you review this? It's been a while since I did the PR and I have not receive any feedback yet.

PR #1052, opened 2022-06-16
Update trie_natives.cpp

rust2:
"Invalid map handle provided (%d)" error is very unclear when it appears in complex code.


rust2:
Oh... Tries here are hash *maps* so that's why they were names as *map*

PR #1056, opened 2022-07-10
fix gcc 12.1 compilation

Flummi:


dvander:
Thanks for the PR - please separate out the sqlite changes in a separate PR, because they're unrelated (even if the goal is gcc-related).

Flummi:
@dvander it's now seperated. :) https://github.com/alliedmodders/amxmodx/pull/1058

PR #1058, opened 2022-07-17
update sqlite to version 3.39.0

Flummi:
related: https://github.com/alliedmodders/amxmodx/pull/1056


SmileYzn:
3.39.2 is out with some fixes

djearthquake:
It will compile if sqlite is NOT optimized referencing pNew and returning local variable, maybe at some point. Everything else compiles fine optimized least with the version I cloned.

PR #1063, opened 2022-09-10
Update admin.sma

pepepepito0147:
small optimization -> g_users_ini

replaced

- format -> formatex
- console_print -> engclient_print

### DIFF
--- a/plugins/admin.sma
+++ b/plugins/admin.sma
@@ -26,11 +26,13 @@ new AdminCount;
 
 new PLUGINNAME[] = "AMX Mod X"
 
+new g_users_ini[] = "users.ini"
+
 #define ADMIN_LOOKUP	(1<<0)
 #define ADMIN_NORMAL	(1<<1)
-#define ADMIN_STEAM		(1<<2)
+#define ADMIN_STEAM	(1<<2)
 #define ADMIN_IPADDR	(1<<3)
-#define ADMIN_NAME		(1<<4)
+#define ADMIN_NAME	(1<<4)
 
 new bool:g_CaseSensitiveName[MAX_PLAYERS + 1];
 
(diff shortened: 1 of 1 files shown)


luxxxoor:
you have wrong indentation, also why static instead of new ? you keep memory used instead of freeing it

pepepepito0147:
> you have wrong indentation, also why static instead of new ? you keep memory used instead of freeing it

1. wrong indentation? #1073 
2. static or new? https://forums.alliedmods.net/showpost.php?p=344418&postcount=1

luxxxoor:
you should only use static when its not feasible to use new

for example:

1.a function that is called multiple times in a second
2.you make a heavy operation and you don't want to repeat it every time

yes, static is faster than new, but you keep memory occupied, why would you need to keep memory occupied for a function that is call only once in a minute ?

just think why c/c++ or similar languages dont use global variables and instead prefer to use dynamic allocation

pepepepito0147:
`is relative`, the same has been previously applied here:

- https://github.com/alliedmodders/amxmodx/blob/master/plugins/admin.sma#L147
- https://github.com/alliedmodders/amxmodx/blob/master/plugins/admin.sma#L562..L566

rtxa:
Well, memory is cheap today and static is useful in expensive operations, but still I see this PR trivial or really minimal priority because the changes will not bring any benefit because this code isn't doing anything expensive or not being called so many times. In any case, the use of constants instead of magic numbers is a great change and isn't mentioned in the PR.

Also, I see that swapping engclient_print() to console_print() is not a good change because the latter has support for UTF-8 character which is useful in a multilingual context while I'm not that sure about engclient_print(). So by trying to bring a small optimization, you're maybe breaking something. I have notice too that the plugin isn't fully translated.

Other reason which the PR is not good is whitespaces changes and code formatting changes which obsfucates what you're really trying to fix/improve and less chances to get your PR merged. Those should be in another PR or in the final commit.

pepepepito0147:
> well this is just the tip of the ***iceberg***, to update all other plugins!

PR #1075, opened 2023-03-06
Update CMisc.h to stop real players showing as 127.0.0.1(That IP has full admin on listenserver.cfg) AND appveyor so the MacOS build goes through.

djearthquake:
Resolve real players being mistaken for bots.


djearthquake:
Using other patch fork.

PR #1076, opened 2023-03-19
Fix CFrameActionMngr (RequestFrame) not being cleared

### LINKED ISSUE #1039: Fix RequestFrame
Hello. After some investigations regarding server crashes I discovered that the RequestFrame handles are not freed on map change. For example, if I use it inside the client_disconnected public forward it will cause memory leaks if the forward is called after pfnChangeLevel. It seems that the reason behind that is the fact the RequestFrame forwards are not freed at mapchange, like they are for most of the functions that return handles, such as register_forward. Manual deletion is also not possible, as the native doesn't return the handle that was created. Could this please be fixed in AMXX 1.10 ?

Thanks in advance! :)

Th3-822:
The fix looks like a workaround because there is no clear method in [amtl/am-deque.h](https://github.com/alliedmodders/amtl/blob/ad9d5c33534dcfd902e2aeb0744df1a227b1c6d3/amtl/am-deque.h) and i don't know if i can or how to call it directly

if i were better at C stuff i could make a pr to add a clear method on am-deque first

Fixes #1039, also don't forget to apply it to 1.9-dev too

### DIFF
--- a/amxmodx/CFrameAction.h
+++ b/amxmodx/CFrameAction.h
@@ -53,6 +53,15 @@ public:
 		}
 	}
 
+	void clear()
+	{
+		int count = m_requestedFrames.length();
+		while (count--)
+		{
+			m_requestedFrames.popFront();
+		}
+	}
+
 private:
 	ke::Deque<ke::AutoPtr<CFrameAction>> m_requestedFrames;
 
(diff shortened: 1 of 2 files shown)


xLeviNx:
This would be ideal to integrate. Nice work

GLoOoccK:
@dvander

lozatto (approved):

Th3-822:
can we get this merged for 1.9-dev too? or it needs another pr?

psychonic:
> can we get this merged for 1.9-dev too? or it needs another pr?

Done

PR #1083, opened 2023-08-05
Add new special format specifiers.

LazZzerrr:
- `%t` : expands to a string containing the player's SteamID.


GLoOoccK:
@dvander

PR #1111, opened 2024-08-24
Add `ignorewarn` pragma to pawn compiler

ArKaNeMaN:
Usage:
```pawn
#pragma ignorewarn 217
// Will ignore all 217 (loose indentation) warnings
```

### DIFF
--- a/compiler/libpc300/sc2.c
+++ b/compiler/libpc300/sc2.c
@@ -1158,6 +1158,12 @@ static int command(void)
           preproc_expr(&val,NULL);
           if (val>0)
             sc_tabsize=(int)val;
+
+        } else if (strcmp(str, "ignorewarn") == 0) {
+          cell val;
+          preproc_expr(&val, NULL);
+          pc_enablewarning(val, 0);
+
         } else if (strcmp(str,"align")==0) {
           sc_alignnext=TRUE;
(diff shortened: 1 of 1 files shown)


xLeviNx:
Good practice would be to not ignore these and fix indentation.

ArKaNeMaN:
> Good practice would be to not ignore these and fix indentation.

Yes, but a have some macros which visually dont have wrong indentation, but compiler shows 217 warnings.
I dont want to refuse to use this macros, so I want to hide this warnings.

Ganter1234:
#pragma tabsize 0

ArKaNeMaN:
@Ganter1234, it did not help)

Giferns:
This might also help in the situations with
`warning 206: redundant test: constant expression is non-zero`

Th3-822:
> #pragma tabsize 0

this one doesn't work because there is a check for > 0 before applying it

dystopm:
> > Good practice would be to not ignore these and fix indentation.
> 
> Yes, but I have some macros which visually dont have wrong indentation, but compiler shows 217 warnings. I dont want to refuse to use this macros, so I want to hide this warnings.

Show those macros, I had that problem and I solved them by using ( ) enclosing variable assignations.

PR #1133, opened 2025-11-14
Race Condition for Language Target When Formatting with `%l`

### LINKED ISSUE #1116: Race Condition for Language Target When Formatting with `%l`
The `%l` format parameter relies on persistence of the language target set by `SetGlobalTransTarget()`.  However, when this was implemented, it reused an existing global variable.  This global variable gets set by several other functions that use it only for temporary purposes (e.g. `server_print()`).  Because of this, it seems that the `%l` format deserves its own global variable so that client-related print functions can properly use this persisted language target.

Previous discussion in the forums [here](https://forums.alliedmods.net/showthread.php?t=350153).

Can be reproduced with the following plugin based on the thread linked above:

```
#include <amxmodx>

public plugin_init() 
{
	register_dictionary("common.txt")
	register_concmd("testlang", "cmdTest")
}

public cmdTest(id)
{
	SetGlobalTransTarget(id)
	custom_print(id, "%l", "CL_NOT_FOUND");
	log_amx("this is gonna force language change to default");
	custom_print(id, "%l", "CL_NOT_FOUND");

	return PLUGIN_HANDLED
}

custom_print(id, msg[], any:...)
{
	new formattedMsg[MAX_FMT_LENGTH];
	vformat(formattedMsg, charsmax(formattedMsg), msg, 3);
	console_print(id, "%s ^3%s", "[prefix]", formattedMsg);
}
```

Join the server and set your personal language to a language that does not match the server default language (e.g. if your server default is "en" then run `setinfo lang fr`).  Run `testlang` in console.

GLoOoccK:
Fixes https://github.com/alliedmodders/amxmodx/issues/1116

### DIFF
--- a/amxmodx/CLang.cpp
+++ b/amxmodx/CLang.cpp
@@ -626,3 +626,8 @@ void CLangMngr::SetDefLang(int id)
 {
 	m_CurGlobId = id;
 }
+
+void CLangMngr::SetFmtLang(int id)
+{
+	m_CurFmtId = id;
+}
--- a/amxmodx/CLang.h
+++ b/amxmodx/CLang.h
@@ -130,6 +130,7 @@ private:
 
 	// Current global client-id for functions like client_print with first parameter 0
 	int m_CurGlobId;
+	int m_CurFmtId;
 public:
 	// Merge a definitions file
 	int MergeDefinitionFile(const char *file);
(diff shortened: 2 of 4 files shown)


psychonic (approved):
Thanks!

PR #1135, opened 2025-11-16
Backport: Convert package.pl to Python.

MisterCalvin:
The [win32-1.9](https://builds.alliedmods.net/amxx/builders/win32-1.9) builder completes the build but times out in the upload phase because it still runs `package.pl`, which hangs while downloading the now-retired GeoLite2 archive. I believe @dvander will need to update the builder on his side to execute `package.py` instead of `package.pl` (just like the [win32-1.10](https://builds.alliedmods.net/amxx/builders/win32-1.10) builder does)

Cherry-picked from 27f451a868c3154a0ab0d4d70e0a650074fd485d


MisterCalvin:
I realized after opening this that [linux-1.9](https://builds.alliedmods.net/amxx/builders/linux-1.9/builds/60/steps/upload/logs/stdio) executes the same `package.pl` and finishes the upload stage without issue, so I'm not sure about this now

psychonic:
I think your first assessment was probably correct. I can take a closer look at this later today or tomorrow.

For what it's worth, the buildbot scripts are public. (There's just no automation to automatically update them on the build master). This is the relevant piece, https://github.com/alliedmodders/build-configs/blob/master/masters/amxmodx/builders.py#L30-L38

MisterCalvin:
Oh nice, thanks psy. Don't know how I missed that

PR #1138, opened 2025-12-31
Implement new meta natives for module development: MF_GetScriptInfo, MF_GetAmxVersion, MF_SetFailState

DarthMan:
Description:

This PR introduces three new meta natives intended to expose more internal data and control to C++ module developers. These additions allow modules to query plugin metadata, check the core AMXX version, and handle plugin failures more discreetly.

Changes:

- MF_GetScriptInfo: Allows a module to retrieve the Title, Author, and Version of a running plugin by providing the AMX* pointer. Arguments can be passed as nullptr if specific fields are not needed. This native was added as currently there is no way to get a plugin name, version and author when writing a module, unlike is the case with SourceMod.

- MF_GetAmxVersion: Returns the current AMX Mod X version string.

- MF_SetFailState: Sets a plugin into a failed state (similar to the Pawn set_fail_state native). Key difference: This native does not automatically print or log the error message, allowing the module developer to handle logging customly or silently if required.

### DIFF
--- a/amxmodx/modules.cpp
+++ b/amxmodx/modules.cpp
@@ -1137,6 +1137,43 @@ const char *MNF_GetAmxScriptName(int id)
 	return nullptr;
 }
 
+bool MNF_GetAmxScriptInfo(const AMX* amx, const char** pszTitle = nullptr, const char** pszAuthor = nullptr, const char** pszVersion = nullptr)
+{
+	CPluginMngr::CPlugin* pl = g_plugins.findPluginFast(const_cast<AMX*>(amx));
+
+	if (pl->getId() != -1)
+	{
+		if (pszTitle != nullptr)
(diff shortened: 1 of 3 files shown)


amxmodx/modules.cpp
```diff
@@ -1137,6 +1137,43 @@ const char *MNF_GetAmxScriptName(int id)
...
+			*pszVersion = pl->getVersion();
+		}
+
+		return pl->isValid();
+	}
+
+	return false;
+}
+
+void MNF_SetAmxFailState(const AMX* amx, const char* str)
```
ClaudiuHKS, line 1167: Here, I'd like to recommend a further param. which would specify whether the failstate will be loud or quiet.
If loud, some errors would log.

PR #1145, opened 2026-06-16
Fix DoD X module hitgroup detection

ClaudiuHKS:
- Fix **DoD X** AMXX module hitgroup detection.

Why?

* There are thousands of cases in which you headshot an enemy and `client_death()` forward executes with `hitGroup != HIT_HEAD`, even if you obviously see their steel pots/ tin hats jumping around and they get the full red intense HS-related screen fade.

Both `this->m_LastHitGroup = pTrace->iHitgroup;` and `gmsgHealth` are executed into the same frame in `CBasePlayer::TraceAttack()`, with `this->m_LastHitGroup = pTrace->iHitgroup;` having a higher priority (executed 1st).


psychonic:
Thanks. I think that offsets like that would be more appropriate for gamedata rather than convars

ClaudiuHKS:
> Thanks. I think that offsets like that would be more appropriate for gamedata rather than convars

You're right, even if **DoDX** ext. already registers some CVARs, better update a small txt/ini than a cfg or the entire module itself. I will take a look and see how **AMXX** ofs. gamedata files work and I'll try my best switching from CVARs to gamedata files (using same Git branch).

PR #1154, opened 2026-07-24
csx: keep get_user_astats/get_user_vstats working against bots

metita:
When `csstats_rankbots` is 0, CSX skips all stat recording for any fight involving a bot (`saveKill`/`saveHit` bail early via `ignoreBots`). That also skips the per-life attacker/victim arrays, so `get_user_astats` and `get_user_vstats` return 0 when a player fights or is killed by a bot, and statsx HUDs lose that info. You end up forced to enable bot ranking just to get the HUDs back.

This records the attacker/victim arrays (the ones those two natives read) regardless of `csstats_rankbots`, and keeps everything that feeds the persistent rank (the `life` totals) and the session/round stats gated as before. So bots still never enter the rank, but the combat HUDs work.

`saveShot` is left unchanged since it doesn't feed `get_user_astats`/`get_user_vstats`.

### DIFF
--- a/modules/cstrike/csx/CMisc.cpp
+++ b/modules/cstrike/csx/CMisc.cpp
@@ -148,6 +148,29 @@ void CPlayer::saveKill(CPlayer* pVictim, int wweapon, int hhs, int ttk)
 	if ( !isModuleActive() )
 		return;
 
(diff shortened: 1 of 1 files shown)


RauliTop:
Tested, works fine.

PR #1156, opened 2026-07-28
Fix debug info limits in amxxpc and speed up the compiler

metita:
Two related sets of changes to the compiler and to the runtime that reads what the compiler writes. Both are about plugins built with `-debug`: the first half fixes the debug info itself, the second half makes `amxxpc` faster (compiling with `-debug` is where the old code hurt most, since the debug string list is the largest list the compiler keeps).

## Debug info: size limit and bounds checking

`AMX_DBG_HDR` counted every table with an `int16_t`. A plugin with more than 32767 line entries wraps the counter, and `append_dbginfo()` then writes a table the runtime cannot read back. That is reachable with a normal large plugin compiled with `-debug`, not just with generated code.

The counts are now `int32_t`, and the counting loops go through a macro that raises `error 102` (table overflow) instead of wrapping. The header layout changed, so the magic goes from `0xf1ef` to `0xf1f0` — see Compatibility below.

On the loading side, `dbg_LoadInfo()` had no idea where the buffer ended. The name scans were plain `for (ptr = ptr + sizeof(...); *ptr != '\0'; ptr++)` loops, and the table counts from the header were trusted as-is, so truncated or corrupt debug data read past the allocation. It now takes the buffer size, checks each entry and each count against the end pointer, and returns `AMX_ERR_FORMAT` instead. `modules.cpp` validates `hdr->size` against the file size before deriving that buffer, so a bad offset is rejected with a log line rather than pointing the loader at arbitrary memory.

`ClipFileName()` copied into a `static char[256]` with `strcpy()`, which a longer path overflows. It clips in place now, which also drops the shared static buffer.

`append_dbginfo()` also wrote a `hdr.size` larger than the data it actually emitted: the symbol pass counted `strlen()` of the whole rest of the debug string instead of just the symbol name. The loader allocates and copies `hdr.size` bytes out of the plugin buffer, so that overcount is read past the end of the file. It is a few KB on a small plugin and hundreds of KB on a large one. `size` is exact now.

The counting pass in `append_dbginfo()` walked the debug string list three times, once per table; it is a single `switch` over one walk now. The counters also stopped being incremented through the address of a packed struct member, which is undefined behaviour and warns on GCC/Clang.

## Compiler speed

Nothing here changes generated code. The changes are all lookup and iteration:

- `insert_string()` walked the whole list to append, so building the debug string list, the autolist and the path list was quadratic in the number of entries. Each list root now keeps a tail pointer.
- `get_dbgstring(index)` and `get_autolist(index)` walk from the head on every call, so the loops that iterate them with an increasing index were quadratic too. Added `get_dbgstring_next()` / `get_autolist_next()` and used them at the call sites.
- `find_subst()` indexed macros by first character only, which degenerates into a linear scan once many macros share a letter (`_` in particular). It uses an FNV-1a hash over the name prefix now. This runs for every identifier of every line, so it shows up on any plugin with a large include set.
- `findglb()` used the linear `find_symbol()` walk over `glbtab` while the hash table was already there and maintained. It uses the hash table now, through a variant that keeps `find_symbol()`'s filter: `FindInHashTable()` also returns parented `iCONSTEXPR` symbols (enum fields), which `findglb()` must not return, so the shared implementation takes a flag instead of loosening the existing behaviour.
- `finddepend()` scanned both symbol tables in full for every symbol, but almost no symbol has a child. Symbols now carry a `has_child` flag set wherever a `parent` is assigned. The flag is never cleared, so a stale `TRUE` only costs the scan that used to happen anyway; `FALSE` is the reliable direction.
- `stgopt()` tried every peephole pattern against every line. If a pattern starts with a literal character, the first comparison in `matchsequence()` decides it, so that case is now short-circuited before the call. Patterns starting with `%`, `!`, `-` or a space are excluded from the shortcut and still go through the full match, so the output is unchanged.

## Compatibility

This is a debug-format break and I would rather call it out than bury it.

A `.amxx` compiled with an older `amxxpc` carries magic `0xf1ef` and 16-bit counts, so the new loader rejects it with `AMX_ERR_FORMAT`, and the reverse is also true. It only matters for plugins actually loaded in debug mode (`amx_debug 2`, or the plugin marked debug in `plugins.ini`); normal loading never touches the debug table. Still, it means existing binaries have to be recompiled to keep debug output.

If you would rather not break that, the loader can keep reading `0xf1ef` with the old 16-bit header as a fallback and only use the wide format for `0xf1f0`. Say the word and I will add it to this branch.

## Notes

The commits are kept separate and each one stands on its own, so anything here can be dropped without touching the rest. I did not run the full build matrix locally, so CI is the first real check on the non-Windows compilers.

### DIFF
--- a/amxmodx/amxdbg.cpp
+++ b/amxmodx/amxdbg.cpp
@@ -61,34 +61,55 @@ void memread(void *dest, char **src, size_t size)
 	*src += size;
 }
 
-const char *ClipFileName(const char *inp)
+static int ValidateCount(int count, size_t element_size, const unsigned char *ptr, const unsigned char *end)
 {
-	static char buffer[256];
+	if (count < 0)
+		return AMX_ERR_FORMAT;
+
+	if ((size_t)count > (size_t)(end - ptr) / element_size)
+		return AMX_ERR_FORMAT;
+
+	return AMX_ERR_NONE;
+}
+
+static unsigned char *FindStringEnd(unsigned char *ptr, const unsigned char *end)
+{
+	while (ptr < end && *ptr != '\0')
+		ptr++;
+
+	return (ptr < end) ? ptr : NULL;
+}
+
+static void ClipFileNameInPlace(char *inp)
+{
 	size_t len = strlen(inp);
-	const char *ptr = inp;
+	char *ptr = inp;
 
 	for (size_t i=0; i<len; i++)
 	{
 		if ((inp[i] == '\\' || inp[i] == '/') && (i != len-1))
 			ptr = inp + i + 1;
 	}
-	strcpy(buffer, ptr);
 
-	return buffer;
+	memmove(inp, ptr, strlen(ptr) + 1);
 }
 
 //Note - I changed this function to read from memory instead.
 // -- BAILOPAN
-int AMXAPI dbg_LoadInfo(AMX_DBG *amxdbg, void *dbg_addr)
+int AMXAPI dbg_LoadInfo(AMX_DBG *amxdbg, void *dbg_addr, size_t dbg_size)
 {
   AMX_DBG_HDR dbghdr;
   unsigned char *ptr;
+  unsigned char *end;
   int index, dim;
   AMX_DBG_SYMDIM *symdim;
 
   assert(amxdbg != NULL);
 
   char *addr = (char *)(dbg_addr);
+
+  if (dbg_addr == NULL || dbg_size < sizeof(AMX_DBG_HDR))
+    return AMX_ERR_FORMAT;
 
   memset(&dbghdr, 0, sizeof(AMX_DBG_HDR));
   memread(&dbghdr, &addr, sizeof(AMX_DBG_HDR));
@@ -107,6 +128,11 @@ int AMXAPI dbg_LoadInfo(AMX_DBG *amxdbg, void *dbg_addr)
   #endif
 
   if (dbghdr.magic != AMX_DBG_MAGIC)
+    return AMX_ERR_FORMAT;
+  if (dbghdr.size < (int32_t)sizeof(AMX_DBG_HDR) || (size_t)dbghdr.size > dbg_size)
+    return AMX_ERR_FORMAT;
+  if (dbghdr.files < 0 || dbghdr.lines < 0 || dbghdr.symbols < 0 || dbghdr.tags < 0
+      || dbghdr.automatons < 0 || dbghdr.states < 0)
     return AMX_ERR_FORMAT;
 
   /* allocate all memory */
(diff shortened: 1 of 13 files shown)


dvander:
I appreciate the attention to this project, but... please, no AI-generated mega-pull-request spam. Especially since this wasn't discussed on Discord or over e-mail beforehand. This kind of PR is extremely difficult to review. Ideally, one PR per individual fix. At the very least, it needs some kind of logical grouping (the compatibility changes, for example, should be discussed in their own PR).

Also, the chart. I have no idea what kind of scale of impact is being described here. If the problem being solved is that a synthetic script takes 6 seconds to build... this is a very old project, with no active development. There's not really any reason to do synthetic make-work. But commits like "Fix invalid debug offset error path" appear to be real bug fixes so it'd be good to split things like that out to give them priority.

We've done performance optimizations on the compiler in the past, but I'd much rather see measurements on a real-world script that is public. That would help you gauge whether it's worthwhile.

metita:
Not sure where is the AI-part which this has been an long-work project for several months now, I just had the idea to share it here, but I do like the approach on making small PRs with each issue, but the real optimization do exists and is being heavily user in our community with large mods and 32/32 users on a daily basis @dvander

metita:
Futhermost, the optimization part is pretty clear, we do have plugins with more than 50k lines which takes more than half a minute to compile even reaching an entire minute, I do recognize that his is a very old project but seems odd to justify the work that has been done in order to make the optimization works for everyone, it was and it is still being heavily tested till now even (we do have active servers right now with these changes added and running), feel free to test it as it was done in real environments already, except for the benchmarks part that was done outside of a real server (which does not make real sense).

Segmentation fault tested, -debug tested and so on...

metita:
Not sure what kind of discussion through discord should exist (is there even a discord? or an email?), pretty sure Github IS the place to discuss this kind of things as you have my proposed PR, we can talk it here, we did our job when making changes to the compiler, pretty sure it is enough as it is.

So yeah, but if there is a negativity or claiming that a "summary and well done benchmark" is being targeted as an AI job I would preffer to delete the PR and using it on my closed environment.

Feels like that doing a favor to the community (a very welcome one) feels like an attack rather than a "oh, cool, we will check it out"

dvander:
GitHub is the place to discuss but for non trivial changes it's best to get some feedback beforehand. That way you can have some confidence that it's on the right track. Individually these patches are probably fine, but they need to be presented as separate PRs to be reviewable. Bug fixes + performance changes + new file format at the same time make the review and feedback loop much more difficult for both of us.

Again, this project has been unmaintained for ten years. I do appreciate the interest, but there's no one to maintain these changes if they break stuff. So if you're absolutely dead-set on changing the file format, and don't want to discuss it separate from the other changes... then a downstream fork seems like the right approach.

